PR #31, opened 2014-12-08
Added support for new Java 8 LocalDate and LocalDateTime

alienisty:
Both JSONWriter and the JSONPopulator now treat LocalDate and LocalDateTime as "primitive" dates.
Note that tests for Date conversions for the JSONPopulator were missing so I've added them together with the ones for LocalDate and LocalDateTime.


lukaszlenart:
Please re-open this PR against `master` branch

PR #35, opened 2015-02-15
Improve LocalizedTextUtil defend NPE

quaff:
defend NPE if not ValueStack presented, for example non-struts request or backend thread.

### DIFF
--- a/xwork-core/src/main/java/com/opensymphony/xwork2/util/LocalizedTextUtil.java
+++ b/xwork-core/src/main/java/com/opensymphony/xwork2/util/LocalizedTextUtil.java
@@ -692,9 +692,12 @@ public class LocalizedTextUtil {
         if (bundle == null) {
             return null;
         }
+        if (valueStack != null) 
             reloadBundles(valueStack.getContext());
         try {
-            String message = TextParseUtil.translateVariables(bundle.getString(key), valueStack);
+            String message = bundle.getString(key);
+            if (valueStack != null) 
+                message = TextParseUtil.translateVariables(message, valueStack);
             MessageFormat mf = buildMessageFormat(message, locale);
             return formatWithNullDetection(mf, args);
         } catch (MissingResourceException e) {


wolpi:
could you add a unit test, please?

lukaszlenart:
I'm not sure if this is a good solution, can you provide more details?

quaff:
@lukaszlenart , It doesn't add new thing just null-check added, It's safe if existing unit tests passed.

lukaszlenart:
What I meant is that if `ValueStack` is missing in context something wrong happened, so solution to null-check is error-prone :)

quaff:
ValueStack will be always not null with struts2 request, so It wouldn't change the behavior of struts2, this pull request is intend to use in non-struts2 context.

lukaszlenart:
That's the case, can you explain what you mean by non-struts2 context?

quaff:
simplest case

```java
import java.util.Locale;
public class Main {
	public static void main(String[] args) {
		System.out.println(LocalizedTextUtil.findText(Main.class, "key",Locale.getDefault(), "defaultMessage", null, null));
	}
}
```
Main.properties
```
key=this is a key
```

it will print "this is a key"

lukaszlenart:
So you can use a `StubValueContext` like this one
```java
public class StubValueStack implements ValueStack {
    Map<String, Object> ctx = new HashMap<String, Object>();
    CompoundRoot root = new CompoundRoot();
    
    public Map<String, Object> getContext() {
        return ctx;
    }

    public void setDefaultType(Class defaultType) {
    }

    public void setExprOverrides(Map<Object, Object> overrides) {
    }

    public Map<Object, Object> getExprOverrides() {
        return null;
    }

    public CompoundRoot getRoot() {
        return root;
    }

    public void setValue(String expr, Object value) {
        ctx.put(expr, value);
    }

    public void setParameter(String expr, Object value) {
        throw new UnsupportedOperationException("not implemented");
    }

    public void setValue(String expr, Object value, boolean throwExceptionOnFailure) {
        ctx.put(expr, value);
    }

    public String findString(String expr) {
        return (String) ctx.get(expr);
    }

    public String findString(String expr, boolean throwExceptionOnFailure) {
        return findString(expr, false);
    }

    public Object findValue(String expr) {
        return findValue(expr, false);
    }

    public Object findValue(String expr, boolean throwExceptionOnFailure) {
        return ctx.get(expr);
    }

    public Object findValue(String expr, Class asType) {
        return findValue(expr, asType, false);
    }

    public Object findValue(String expr, Class asType, boolean throwExceptionOnFailure) {
        return ctx.get(expr);
    }

    public Object peek() {
        return root.peek();
    }

    public Object pop() {
        return root.pop();
    }

    public void push(Object o) {
        root.push(o);
    }

    public void set(String key, Object o) {
        ctx.put(key, o);
    }

    public int size() {
        return root.size();
    }
}
```

quaff:
It's just a static util class, It should be light and simple, your suggestion is opposite.

lukaszlenart:
But you cannot pass `null` and expect that it will work properly, if you want such behaviour add method with signature that doesn't expect `ValueStack`

quaff:
I hope so, please add a method drop last two arguments, thanks.

lukaszlenart:
Please re-open this PR against `master` branch

PR #39, opened 2015-05-22
WW-4503 Build Struts with JDK 7

swiftelan:


pom.xml
```diff
@@ -352,7 +353,7 @@
             <dependency>
                 <groupId>org.apache.struts</groupId>
                 <artifactId>struts-annotations</artifactId>
-                <version>1.0.5</version>
+                <version>1.0.6-SNAPSHOT</version>
```
lukaszlenart: Do we need this? I thought after dropping APT we can drop this as well

lukaszlenart:
It's better to use two dedicated PRs where each addresses just one issue.

pom.xml
lukaszlenart: Now I get it, we must first update struts-annotations project, release it and then we can use it here

lukaszlenart:
This PRs mixed fixes for two different things :(

swiftelan:
I understand. I merged changes from the GitHub mirror to my fork. I didn't know that it would update the existing pull request, that was not my intention.

lukaszlenart:
Yeah... each change must be in separated branch and every commit to that branch will be automatically exposed as a part of PR. That's how it works :)

lukaszlenart:
Are you going to open a new PR to introduce JDK7? If not I can do it directly based on your changes - it's an easy fix :)

swiftelan:
Moving to Java 7 requires the new annotation parser or the removal of the maven-apt-plugin. The annotation parser [job](https://builds.apache.org/view/S-Z/view/Struts/job/struts-annotations/) is disabled. Using a snapshot version of the annotation processor will fail the build. It might be easier for you to coordinate changes to both repositories.

lukaszlenart:
Ok, I will handle this based on your work, thanks! It will take some time as releasing new version of struts-annotations requires a vote

PR #43, opened 2015-06-20
WW-4516: Allow JSON plug-in to skip bean cache when devMode is enabled.

Naros:
Adds support for disabling bean cache on JSON plugin when devMode is enabled.

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
@@ -96,7 +96,7 @@ public class JSONResult implements Result {
     private String contentType;
     private String wrapPrefix;
     private String wrapSuffix;
-    private boolean debugMode = false;
+    private boolean devMode = false;
     
     @Inject(StrutsConstants.STRUTS_I18N_ENCODING)
     public void setDefaultEncoding(String val) {
@@ -105,7 +105,7 @@ public class JSONResult implements Result {
     
     @Inject(StrutsConstants.STRUTS_DEVMODE) 
     public void setDebugMode(String val) {
-    	this.debugMode = BooleanUtils.toBoolean(val);
+    	this.devMode = BooleanUtils.toBoolean(val);
     }
     
     /**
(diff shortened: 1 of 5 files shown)


plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -92,12 +96,18 @@
     private String contentType;
     private String wrapPrefix;
     private String wrapSuffix;
-
+    private boolean debugMode = false;
```
lukaszlenart: Please rename this to `devMode`

plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
```diff
@@ -53,7 +68,22 @@
      * @throws JSONException
      */
     public static String serialize(Object object) throws JSONException {
```
lukaszlenart: Is this method used somewhere else? If not, it's the perfect time to drop it

Naros: Not that I can see from just scanning my github fork.  I'll double check when I get home tonight and if not; I'll remove it.

Naros: It is currently used in the test cases
JSONResultTest - testJSONUtilNPEOnNullMethod()
JSONUtilTest = testSerializeDeserialize()
Thoughts?

lukaszlenart: Throw away and use the new version

plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
Naros: Changed in latest commits per your request.

plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -92,12 +96,18 @@
...
+    private boolean devMode = false;
+    
     @Inject(StrutsConstants.STRUTS_I18N_ENCODING)
     public void setDefaultEncoding(String val) {
         this.defaultEncoding = val;
     }
-
+    
+    @Inject(StrutsConstants.STRUTS_DEVMODE) 
+    public void setDebugMode(String val) {
```
lukaszlenart: Can you rename setter as well? Thanks!

PR #44, opened 2015-06-30
WW-4520 - Add prefix for CSS classes.

aleksandr-m:


lukaszlenart:
Hm... I wonder how hard it will be to make this configurable so we can keep backward compatibility and allow enable a custom prefix on demand.

cnenning:
> I wonder how hard it will be to make this configurable 

Making this configurable would be cool. But when looking at changed files, and all those different file types, i fear it would be hard.

yasserzamani:
@apache/struts-committers , I'm going to close this one as already there are users mentioning such things as an information disclosure security issue. objections?

lukaszlenart:
@aleksandr-m do you think it will be hard to update this PR? Or maybe create a new one?

aleksandr-m:
@lukaszlenart You mean to allow to change prefix? Do you have an idea how to do that?

lukaszlenart:
> @lukaszlenart You mean to allow to change prefix? Do you have an idea how to do that?

Yes, I know we are supporting a different set of extensions and I do not have a solution at hand. Other option is to develop a new html5 theme from scratch.

PR #48, opened 2015-09-09
Uniform comparison

quaff:
Add more defense, like compare Enum to String or else.


lukaszlenart:
Is there any related JIRA ticket?

quaff:
You are so quick, I create WW-4544 just now.

PR #49, opened 2015-09-10
WW-4539 Support list of tokens to prevent CSRF attack in async requests

yasserzamani:
Nowadays, modern web applications use JQuery, and so, they can send multiple async requests in parallel to the web server. Struts has a great token mechanism which one of it's use cases is preventing from CSRF attack, but it's not usable or hard to use when the user application has multiple valid submission from one jsp using JQuery.

This pull request contains all changes needed to enable Struts to handle this, transparently (i.e. these changes do not break current user application which uses previous token mechanism). Also, this is protected from DOS attack by supplying a maximum for count of concurrent valid requests.

I hope you enjoy it.
Thanks!

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/Token.java
+++ b/core/src/main/java/org/apache/struts2/components/Token.java
@@ -21,6 +21,7 @@
 
 package org.apache.struts2.components;
 
+import java.util.ArrayList;
 import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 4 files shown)


aleksandr-m:
Do we really need a list of tokens? Currently if you use different names for tokens they all will be [stored into session](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/util/TokenHelper.java#L91). Does using token tag with a name solve you problem?

yasserzamani:
@aleksandr-m is genius. Using token tag with different names in same JSP page handles this need perfectly.

PR #50, opened 2015-09-14
Improve LocalizedTextUtil.getMessage() defend NPE

quaff:
Defend NPE if not ValueStack presented, for example non-struts request like static util method or backend thread.
reopen #35

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/LocalizedTextUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/LocalizedTextUtil.java
@@ -736,9 +736,12 @@ public class LocalizedTextUtil {
         if (bundle == null) {
             return null;
         }
+        if (valueStack != null) 
             reloadBundles(valueStack.getContext());
         try {
-            String message = TextParseUtil.translateVariables(bundle.getString(key), valueStack);
+        	String message = bundle.getString(key);
+        	if (valueStack != null) 
+        		message = TextParseUtil.translateVariables(bundle.getString(key), valueStack);
             MessageFormat mf = buildMessageFormat(message, locale);
             return formatWithNullDetection(mf, args);
         } catch (MissingResourceException e) {


lukaszlenart:
You must update your `fork` - it's outdated and it cannot be merged :\

quaff:
It's up-to-date on branch master. you can check commit history.
https://github.com/quaff/struts/commits/master
https://github.com/apache/struts/commits/master

lukaszlenart:
hm... but Jenkins has a problem with building this PR :\

quaff:
http://stackoverflow.com/questions/16721629/jenkins-returned-status-code-128-with-github
https://issues.jenkins-ci.org/browse/JENKINS-10385
I hope it helps.

lukaszlenart:
Thanks for the links!

It has nothing to do with a key, my PR #47 passed few days ago. I think the second link is the answer but recreating a job means this PR won't be included :\

quaff:
What should I do?

wolpi:
Usually it is not supported to use struts2 utility classes outside of struts context. I would rather not merge this as it increases maintanance burden.

That it cannot be built on jenkins seems to be a configuration issue, it builds on my box.

quaff:
@wolpi , It's a very very simple and safe improvement, It doesn't increases maintenance burden, It make struts more robust.

wolpi:
> It make struts more robust.

When LocalizedTextUtil is used as intended and supported it is already robust :wink: 


>  It doesn't increases maintenance burden

Well, this small patch itself does not. But the idea that struts code should work even when no ValueStack is available. There are surley many more places in the code base where such null checks are necessary to really support Util classes outside of struts context. And the maintanance burden comes over time. Even when your use case would work right now, with every change any struts developer does, or external contributers are doing, they must remember to make their changes working even outside of struts context.

Or put it another way: LocalizedTextUtil is not part of public API. It is a struts internal utility.

What acutally is your use case? Can't you use java resource bundle API?

lukaszlenart:
> It doesn't increases maintenance burden

It does as my plan is to drop all those static beans (static classes with a lot of dependencies passed in methods) and convert them in to fully injectable beans. Thus will allow to simplify its logic and allow for custom implementation (ie. DB based resources)

PR #52, opened 2015-10-05
Adding support for RFC3339 dates with only the date part

andrea-ligios:
RFC 3399 does not exist. RFC 3339 is the one used here: https://tools.ietf.org/html/rfc3339


andrea-ligios:
@wolpi Thanky you! You could vote the [JIRA](https://issues.apache.org/jira/browse/WW-4551) if interested :) 

The tests are in this file: 

https://github.com/andrea-ligios/struts/blob/d95fc63992f3892519d970b07052776e0401003a/core/src/test/java/com/opensymphony/xwork2/conversion/impl/XWorkConverterTest.java

The test could be added in the `testDateConversion()` method, the code should be:

    Date dateRfc3339 = (Date) converter.convertValue(context, null, null, null, "2001-01-10", Date.class);
    assertEquals(date, dateRfc3339);

Could you add it to that file ? I'm just new to the whole thing (Git, GitHub, etc...). Thanks

PR #54, opened 2015-10-09
Make it easier to extend to use own ActionInvocation

perfnorm:
I needed to extend DefaultActionProxyFactory to return my own ActionInvocation, but I had to override all of createActionProxy() to do so.  Seems like a safe change to break out the ActionInvocation creation to make it easier to just override that piece.


wolpi:
Please create a JIRA issue. That makes it easier to track changes and contributions.

https://issues.apache.org/jira/browse/WW

perfnorm:
Done, thanks @wolpi (also added for my other pull request)

PR #56, opened 2015-11-09
Email validation blocks upper case letters

LukaszRacon:
Fix makes email validation regex case insensitive.

### DIFF
--- a/core/src/main/resources/template/xhtml/form-close-validate.ftl
+++ b/core/src/main/resources/template/xhtml/form-close-validate.ftl
@@ -114,7 +114,7 @@ END SNIPPET: supported-validators
                 <#if validator.shortCircuit>continueValidation = false;</#if>
             }
             <#elseif validator.validatorType = "email">
-            if (continueValidation && fieldValue != null && fieldValue.length > 0 && fieldValue.match("${validator.regex?js_string}")==null) {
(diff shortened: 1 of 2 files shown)


core/src/main/resources/template/xhtml/form-close-validate.ftl
```diff
@@ -114,7 +114,7 @@ END SNIPPET: supported-validators
                 <#if validator.shortCircuit>continueValidation = false;</#if>
             }
             <#elseif validator.validatorType = "email">
-            if (continueValidation && fieldValue != null && fieldValue.length > 0 && fieldValue.match("${validator.regex?js_string}")==null) {
+            if (continueValidation && fieldValue != null && fieldValue.length > 0 && fieldValue.match(new RegExp("${validator.regex?js_string}", "i"))==null) {
```
lukaszlenart: What about using simplified form and match the same case few lines below?
```
"/${validator.regex?js_string}/i"
```

LukaszRacon:
I updated PR with a shorter literal regexp notation.

`"/regexString/i"` will not work since it would be converted by str.match function to RegExp object
`new RegExp("/regexString/i")` while we need RegExp with a case-insensitive flag: `new RegExp("regexString", "i")`

So in this case we could use:
```
fieldValue.match(new RegExp("${validator.regex?js_string}", "i"))==null
```
or shorter version:
```
fieldValue.match(/${validator.regex}/i)==null
```
I also fixed url validation that used incorrect RegExp.

PR #58, opened 2015-11-25
WW-4572 Http parameters

lukaszlenart:
Implements WW-4572

This PR converts `parameters` to be represented by a dedicated class - `HttpParameters` - instead of using a simple `Map`.


cnenning:
As far as I see it, there is no new functionality (like security checks) right now. Currently the code is just refactored to work with new interface/class `Parameter` and `HttpParameters`, am I right?

The method `Parameter.isExpression()` seems not to be used or implemented (ohter than `return false`)? The new class `Evaluated` is only used by `AliasInterceptor`?

This looks promising ...

Beside future things, I have two questions about how things are done right now:
* It seems to me that currently converters are not used. Now we got `Parameter.Request.getValue()`. Should converters be used there?
* My first impression is that that method could be slow, as it could be called quite often. What about caching the result of `toStringArray()` in a member?

ke4qqq:
Closing and reopening to make sure Travis is working.

PR #59, opened 2015-11-25
Upgrade tiles plugin

lukaszlenart:
Fixes WW-4568

This PR upgrades the Tiles Plugin and adjusts its API to be compatible with Tiles 2.2.2 - this allows introduce support for OGNL, EL, Wildcards and so on.


cnenning:
As I said on the mailing-list before, I'm not deep into tiles. But from just reading the code it seems reasonable.

I cannot test it with an own app as I switched to tiles3 in the meantime.

When accessing tiles demo in showcase app I got attached exception. The class `javax.el.ELContext` exists in several JARs (like jsp-api.jar or el-api.jar) so it might be a classpath issue. The Exception disapears when I remove the `${title}` expression in `header.jsp`.

I like how dependencies are declared. I.E. that all the tiles jars like `tiles-jsp`, `tiles-freemarker` are explicitly listed. The tiles3-plugin just declares `tiles-extras` which drags in a lot of stuff I don't want in my app.
[tiles-stacktrace.txt](https://github.com/apache/struts/files/45777/tiles-stacktrace.txt)

lukaszlenart:
I have the same issue, no idea what's wrong but will try to figure it out. Regarding Tiles 3, it looks very similar to Tiles 2 and I'm going to work on them with knowledge gathered during upgrading this plugin.

cnenning:
I hope that code can be shared between both tiles plugins. What do you think of creating a strust-tiles-commons.jar or somthing the like?

lukaszlenart:
Will see :)

lukaszlenart:
@cnenning issue with `javax.el` resolved

cnenning:
yay, it runs ! :smiley:

aleksandr-m:
Tested with own app and it is running fine. Looking forward to see the same approach with `StrutsTilesContainerFactory` in tiles3 plugin.

BTW what is accessible through `OGNL:` expression?

lukaszlenart:
> BTW what is accessible through `OGNL:` expression?

I didn't test that but it should be the whole stack

lukaszlenart:
I have moved this PR to be opened against `support-2-3` and it targets Struts 2.3.x. Please see #60 and I'm going to close this PR.

PR #60, opened 2015-11-28
Struts 2.3.x: Tiles plugin upgrade

lukaszlenart:
Fixes [WW-4568](https://issues.apache.org/jira/browse/WW-4568)

This PR upgrades the Tiles Plugin and adjusts its API to be compatible with Tiles 2.2.2 - this allows introduce support for OGNL, EL, Wildcards and so on.


cnenning:
For struts 2.3 I still got an old app with tiles2 which does some custom stuff and which has issues with these changes.

* `tiles-servlet-wildcard` depends on `spring-web`. That causes spring to be pulled in to all apps that use `struts-tiles-pluign`. Due to this I would prefer wildcard support could be optional
* in `TilesContainerFactory.createContainerFactory()` it is necessary for me to register `tilesContainer` in `servletContext`. Can be done like this, would you mind adding this to `StrutsTilesContainerFactory`?
```
@Override
protected BasicTilesContainer instantiateContainer(TilesApplicationContext applicationContext) {
	CachingTilesContainer tilesContainer = new CachingTilesContainer();
	ServletContext servletContext = (ServletContext) applicationContext.getContext();
	ServletUtil.setContainer(servletContext, tilesContainer);
	return tilesContainer;
}
```
* in `StrutsTilesContainerFactory.getSourceURLs()` these two resource patterns are used: `/WEB-INF/**/tiles*.xml` and `classpath*:META-INF/**/tiles*.xml`. The 2nd one causes an exception in WebSphere with IBM JDK. Would be nice if we could make this configurable. The exception is `Caused by: java.net.MalformedURLException: SRVE0238E: Resource paths should have a leading slash`

Currently I have created my own `TilesListener`, `TilesInitializer` and `TilesContainerFactory`. With the above changes I could avoid that and just stick to `StrutsTilesListener`.

Two minor things I noticed in showcase app:
* It logs this message: `WARN (freemarker.jsp:75) - Custom EL functions won't be loaded because no ObjectWarpper was specified .` So it may be that `tiles-el` along with `tiles-freemarker` are not working?
* css layout seems to be broken in firefox (surely not related to this PR, looks good in chrome and ie11, is fixed in master branch)

lukaszlenart:
@cnenning thanks for a really good feedback. I implemented a dedicated ApplicationContext to support wildcard matching without depending on Spring - by default all `tiles*.xml` will be loaded.

After cleaning up dependencies I didn't notice the issue with EL, could you check that on your side?

Thanks!

cnenning:
The new version looks awesome!

Just a small addon to constructor of `StrutsWildcardServletTilesApplicationContext`, that re-enables loading of `tiles*.xml` from classpath and works in IBM stuff, too:
```
        Enumeration<URL> cpResources = getClass().getClassLoader().getResources("/");
        while (cpResources.hasMoreElements()) {
            URL cpResource = cpResources.nextElement();
            urls.add(cpResource);
        }
```


Regarding the other issues:
* css looks alright again in FF, seems like it was a FF bug
* About EL stuff: I'm pretty helpless and pretty confused. That log message is created in `freemarker.ext.jsp.TaglibFactory`. It looks for a `BeansWrapper` which struts creates in `FreemarkerManager.createObjectWrapper()`. When tiles-layout is created with a JSP, and just tiles-attributes are done with FTLs everything seems OK. But when tiles-layout is done with FTL that message is logged and EL support seems broken.

lukaszlenart:
I still cannot reproduce the issue with missing support for EL, here is my example I have been using to test this PR

https://github.com/lukaszlenart/struts2-tiles-demo

cnenning:
> https://github.com/lukaszlenart/struts2-tiles-demo

I created a PR for that demo app which demonstrates the issue.

See https://github.com/lukaszlenart/struts2-tiles-demo/pull/2

lukaszlenart:
This is what I get (I have been using Jetty)

cnenning:
> This is what I get (I have been using Jetty)

Hmm, the freemarker template is not executed but it's code is shown. This is another issue. No idea why that happens.

cnenning:
I would say we merge this and may fix the "EL not working in freemarker-insert with freemarker-template" issue in master.

lukaszlenart:
Let's do it, at least I will have less problem with debugging :D

PR #61, opened 2015-12-31
fix for WW-4269, ww-4576, ww-4577

victorsosa:
blank double results in NumberFormatException being thrown on sending
JSON in request body

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -240,6 +240,7 @@
         <dependency>
             <groupId>ognl</groupId>
             <artifactId>ognl</artifactId>
+            <version>${ognl.version}</version>
         </dependency>
 
         <!-- Velocity -->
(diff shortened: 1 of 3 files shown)


victorsosa:
I included the ww-4576 and ww-4577

core/pom.xml
```diff
@@ -240,6 +240,7 @@
         <dependency>
             <groupId>ognl</groupId>
             <artifactId>ognl</artifactId>
+            <version>${ognl.version}</version>
```
lukaszlenart: It isn't needed, `version` is inherited from parent pom.

victorsosa: Ok, I will remove it

victorsosa: it is ready

victorsosa:
Glad to help, I will be committing more code

lukaszlenart:
Thanks!

PR #62, opened 2016-01-02
Patch ww-4564

victorsosa:
set content type see ww-4564

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/PostbackResult.java
+++ b/core/src/main/java/org/apache/struts2/result/PostbackResult.java
@@ -76,10 +76,8 @@ import java.util.Map;
  */
 public class PostbackResult extends StrutsResultSupport {
 
-    /**
-     * 
-     */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/struts2/result/PostbackResult.java
```diff
@@ -76,6 +76,10 @@
  */
 public class PostbackResult extends StrutsResultSupport {
 
+    /**
+     * 
+     */
+    private static final long serialVersionUID = -2283504349296877429L;
```
lukaszlenart: Could you drop that empty comment section? And please add an empty line to separate static field from the other fields. Thanks :)

victorsosa: Done

lukaszlenart: :+1:

PR #64, opened 2016-01-03
Patch WW-3462 connection open after report generation

victorsosa:
Now the connection is closed after been used.


victorsosa:
:+1:

PR #70, opened 2016-01-05
WW-4582 Permanent patch for security issue CVE-2014-0094 adds 'class' to exclude

victorsosa:
adds 'class' to exclude params in ParametersInterceptor (avoid ClassLoader manipulation)

### DIFF
--- a/core/src/test/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsCheckerTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsCheckerTest.java
@@ -10,10 +10,7 @@ public class DefaultExcludedPatternsCheckerTest extends XWorkTestCase {
 
     public void testHardcodedPatterns() throws Exception {
         // given
-        List<String> params = new ArrayList<String>() {
-
-            private static final long serialVersionUID = 5687184571054993717L;
-            
+        List<String> params = new ArrayList<String>() {           
             {
                 add("%{#application['test']}");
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsCheckerTest.java
```diff
@@ -11,6 +11,9 @@
     public void testHardcodedPatterns() throws Exception {
         // given
         List<String> params = new ArrayList<String>() {
+
+            private static final long serialVersionUID = 5687184571054993717L;
```
lukaszlenart: Could you disable auto-generation of `serialVersionUID`?

victorsosa: OK, I will remove that line

victorsosa:
This close also the CVE-2014-0112, CVE-2014-0113 and CVE-2014-0116

lukaszlenart:
This isn't really needed as access to `Class` is blocked for any Ognl expression, see [Internal Security Mechanism](http://struts.apache.org/docs/security.html#Security-Internalsecuritymechanism)

lukaszlenart:
Also based on a report from a user I have added this 74e26830d2849a84729b33497f729e0f033dc147 - this is almost the same as yours but it bases on a real-life example :)

victorsosa:
ok, PR closed

PR #73, opened 2016-01-12
WW-4584: Upgrade tiles plugin

lukaszlenart:
[WW-4584](https://issues.apache.org/jira/browse/WW-4584) Upgrades Tiles plugin to use Tiles 3 version, it also drops Tiles 3 plugin - there will be only one plugin to support Tiles 3


cnenning:
:+1: 
Looks great!

There is one thing: when I try to run showcase app I get exceptions when trying to access tiles examples: `org.apache.tiles.definition.NoSuchDefinitionException: Cannot find definition named 'showcase.index'`

The definition is defined in `showcase/src/main/webapp/WEB-INF/tiles.xml`. When I try to copy that file to different locations (like `webapp/` or `src/main/resources/`) the error remains.

IMO that can be solved in `master`.

lukaszlenart:
I just ported the solution from `support-2-3` branch and adjusted API, didn't test it yet but will do it later today, tomorrow. I hope it's be an easy fix ;-)

aleksandr-m:
Same thing here. Getting `NoSuchDefinitionException` exception in my app when trying to use tiles result.

Note: there is still reference to tiles3-plugin in bom pom.xml.

lukaszlenart:
Thanks, I found the reason, now working on solution

lukaszlenart:
I have resolved the problem with missing definitions but I'm a bit confused how Tiles resolves resources - 66d29d4 - basically all the definitions are loaded on startup but then are loaded again.

aleksandr-m:
:+1: 
It works.
Thank you for your great work.

cnenning:
> I have resolved the problem with missing definitions but I'm a bit confused how Tiles resolves
> resources - 66d29d4 - basically all the definitions are loaded on startup but then are loaded again.

In my experiments the tiles.xml was only loaded one time, at first request.

lukaszlenart:
> In my experiments the tiles.xml was only loaded one time, at first request.

Yes, it was clarified during my discussion on Tiles Users mailing list

lukaszlenart:
I think the solution is final, I have added support for Struts' Locale and properly implemented support for it. I also updated `struts-examples`. Enjoy!

cnenning:
:+1:

cnenning:
> Simplifies logging

Still not final :wink: ?

Let's merge it!

lukaszlenart:
I have got some comments on my questions sent to Tiles Users mailing list and the current implementation isn't optimal ;-) But I have no better idea so let's merge it :D

PR #74, opened 2016-01-18
propose fix for ww-4456

victorsosa:
HI Lukasz,

This is a proposal to fix WW-4456 - Got NPE when File Upload Limitation exceeds in portal env.

We should catch in any case a FileUploadException, then try to cast down if we want any info about the exception. Valid subtype of FileUploadException:
InvalidContentTypeException
IOFileUploadException
SizeLimitExceededException
FileSizeLimitExceededException

I couldn't get the complete test on a running test env, because I had issues getting the logging to work.

But it seems to works.


victorsosa:
Closed; Just got contaminated with other changes

lukaszlenart:
The simplest approach it to not reverse merge `master` into your PRs till they are green.

PR #75, opened 2016-01-18
WW-4590 - Allow to use multiple names in result

aleksandr-m:
Allow to use multiple values in `result` tag `name` attribute and in `Result` annotation `name`.

So this typical configuration:

```
<action name="save">
    <result>success.jsp</result>
    <result name="error">input-form.jsp</result>
    <result name="input">input-form.jsp</result>
</action>
```

Can be shorten to that:

```
<action name="save">
    <result>success.jsp</result>
    <result name="error, input">input-form.jsp</result>
</action>
```

And this annotations:

```
@Action(results = {
    @Result(name="error", location="input-form.jsp"),
    @Result(name="input", location="input-form.jsp"),
    @Result(name="success", location="success.jsp")
})
```

To that:

```
@Action(results = {
    @Result(name="error, input", location="input-form.jsp"),
    @Result(name="success", location="success.jsp")
})
```

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
@@ -777,12 +777,9 @@ public class XmlConfigurationProvider implements ConfigurationProvider {
                 }
                 params.putAll(resultParams);
 
-                Set<String> resultNamesSet;
-                if (",".equals(resultName.trim())) {
-                    resultNamesSet = new HashSet<>(1);
+                Set<String> resultNamesSet = TextParseUtil.commaDelimitedStringToSet(resultName);
+                if (resultNamesSet.isEmpty()) {
                     resultNamesSet.add(resultName);
-                } else {
-                    resultNamesSet = TextParseUtil.commaDelimitedStringToSet(resultName);
                 }
 
                 for (String name : resultNamesSet) {
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
@@ -777,11 +777,21 @@ public class XmlConfigurationProvider implements ConfigurationProvider {
                 }
                 params.putAll(resultParams);
 
-                ResultConfig resultConfig = new ResultConfig.Builder(resultName, resultClass)
-                        .addParams(params)
-                        .location(DomHelper.getLocationObject(element))
-                        .build();
-                results.put(resultConfig.getName(), resultConfig);
+                Set<String> resultNamesSet;
+                if (",".equals(resultName.trim())) {
+                    resultNamesSet = new HashSet<>(1);
+                    resultNamesSet.add(resultName);
+                } else {
+                    resultNamesSet = TextParseUtil.commaDelimitedStringToSet(resultName);
+                }
+
+                for (String name : resultNamesSet) {
+                    ResultConfig resultConfig = new ResultConfig.Builder(name, resultClass)
+                            .addParams(params)
+                            .location(DomHelper.getLocationObject(element))
+                            .build();
+                    results.put(resultConfig.getName(), resultConfig);
+                }
             }
         }
 
--- a/core/src/test/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProviderResultsTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProviderResultsTest.java
@@ -15,6 +15,7 @@
  */
 package com.opensymphony.xwork2.config.providers;
 
+import com.opensymphony.xwork2.Action;
 import com.opensymphony.xwork2.ActionChainResult;
 import com.opensymphony.xwork2.SimpleAction;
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
```diff
@@ -777,11 +777,21 @@ protected Class verifyResultType(String className, Location loc) {
...
 
-                ResultConfig resultConfig = new ResultConfig.Builder(resultName, resultClass)
-                        .addParams(params)
-                        .location(DomHelper.getLocationObject(element))
-                        .build();
-                results.put(resultConfig.getName(), resultConfig);
+                Set<String> resultNamesSet;
+                if (",".equals(resultName.trim())) {
+                    resultNamesSet = new HashSet<>(1);
+                    resultNamesSet.add(resultName);
```
lukaszlenart: the final resuly set will contain only `,`?

lukaszlenart:
Great idea!

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
aleksandr-m: Currently it is allowed to have a result name as a `,`. And `TextParseUtil.commaDelimitedStringToSet` will produce empty set in that case. Note that it is trimmed so the result name can be something like "&nbsp;&nbsp;&nbsp;," also.

lukaszlenart: `if (",".equals(resultName.trim()))` assumes that the result name contains only `,` and then it will produce a set with `,` (value of `resultName`). `TextParseUtil.commaDelimitedStringToSet` is in the `else` section so it won't be used in this case.

lukaszlenart: Why not just do it this way instead of this `if..else`?
`resultNamesSet = TextParseUtil.commaDelimitedStringToSet(resultName);`

swiftelan:
Annotations allow multivalued attributes in the form of arrays. Using this makes it obvious that you can supply multiple result names for a single result configuration.

@Action(results = {
    @Result(name={"error", "input"}, location="input-form.jsp"),
    @Result(name="success", location="success.jsp")
})

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
aleksandr-m: You mean like this?

    Set<String> resultNamesSet = TextParseUtil.commaDelimitedStringToSet(resultName);
    if (resultNamesSet.isEmpty()) {
        resultNamesSet.add(resultName);
    }

Yeah, looks a bit cleaner to me. I'll change it.

aleksandr-m:
@swiftelan I was thinking about that too, BUT:

* It will break backward compatibility
* I kind of like that annotation attributes looks the same as in xml configuration

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
lukaszlenart: No, I think why `,` is a valid result name :) But that's ok, I mean your proposal :)

swiftelan:
@aleksandr-m

An annotation attribute that is defined as an array does not have to have the curly braces to define a single value for that attribute.

Basing a Java API for configuration on XML syntax is a poor implementation. Why not support the entire XML configuration syntax in the annotation? Java has APIs to express a collection of items. It is much easier to utilize an array than a comma separated value attribute. The array allows the developer to clearly express their intent. Utilizing an array does not require parsing when the configuration value is accessed.

aleksandr-m:
@swiftelan So you are proposing to change only type? The name of the attribute will stay the same i.e `name`, right? There is still some chance that something going to break, note that `ResultInfo` depends on `Result`'s `name`. But it is better than changing the name.

@lukaszlenart WDYT about changing `Result`'s `name` to `String[] name() default com.opensymphony.xwork2.Action.SUCCESS;` and probably removing `public ResultInfo(Result, PackageConfig, String, Class<?>, Map<String, ResultTypeConfig>)` constructor?

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
aleksandr-m: It looks a bit odd, agree. :) But it is supported right now and it is better than mapping `,` to `success` or not creating result at all.

lukaszlenart: Right, thanks for explanation!

lukaszlenart:
@aleksandr-m I'm ok with that and I agree with @swiftelan - let's follow standards

lukaszlenart:
I would like to merge this PR and start release process of BETA3 :)

aleksandr-m:
Done. We should add notice in the release notes about breaking change in `ResultInfo` / `@Result` for those who use `Result.name()` or extend `ResultInfo`.

PR #76, opened 2016-01-18
WW-4412  Context started without being closed

victorsosa:
WW-4412

Context started without being closed


victorsosa:
Closed this pull request is right; I will send another one

PR #79, opened 2016-01-19
WW-4412  Context started without being closed

victorsosa:
Patch for WW-4412

Context started without being closed


plugins/tiles/src/main/java/org/apache/struts2/views/tiles/TilesResult.java
```diff
@@ -110,7 +110,9 @@ public void doExecute(String location, ActionInvocation invocation) throws Excep
         HttpServletResponse httpResponse = ServletActionContext.getResponse();
 
         Request request = new ServletRequest(applicationContext, httpRequest, httpResponse);
-
+        
+        container.startContext(request);
         container.render(location, request);
+        container.endContext(request);
```
lukaszlenart: This isn't needed, `container.render` starts and ends context.

victorsosa: OK, i will close the PR then :+1:

lukaszlenart:
Thanks!

PR #81, opened 2016-01-19
WW-4589 do not overwrite explicit method name

rpii:
Lukasz,

here is a proposed fix for WW-4589.  This makes sure a url specified method name is not overwritten by an implied method name.  It passes the unit tests that I have added


lukaszlenart:
Merged, thanks!

PR #83, opened 2016-01-20
WW-4593 Easymock update to 3.4

victorsosa:
This is the upgrade of the Easymock library for the unit tests

Every unit test is changed to keep the same result but using the new API.


lukaszlenart:
grey bar :(

victorsosa:
Green :+1:

PR #84, opened 2016-01-20
WW-3940 Upgrade to Velocity Tools 2.0

victorsosa:
JWebUnit upgrade to 3.3 too


plugins/sitemesh/src/main/java/org/apache/struts2/sitemesh/VelocityDecoratorServlet.java
```diff
@@ -78,17 +88,18 @@ public void init(ServletConfig config) throws ServletException {
         velocityManager.init(config.getServletContext());
 
         // do whatever we have to do to init Velocity
-        setVelocityEngine(velocityManager.getVelocityEngine());
-        toolboxManager = velocityManager.getToolboxManager();
+        getVelocityView().setVelocityEngine(velocityManager.getVelocityEngine());
+        //getVelocityView().getToolboxFactory(). toolboxManager = velocityManager.getToolboxManager();
```
victorsosa: WARN: I didn't find a way to translate this line to the new API, but the ALL test are passing

PR #86, opened 2016-02-10
Allow html5 attributes

gregh3269:
See WW-4607

### DIFF
--- a/plugins/javatemplates/src/main/java/org/apache/struts2/views/java/simple/TextFieldHandler.java
+++ b/plugins/javatemplates/src/main/java/org/apache/struts2/views/java/simple/TextFieldHandler.java
@@ -32,7 +32,7 @@ public class TextFieldHandler extends AbstractTagHandler implements TagGenerator
         Map<String, Object> params = context.getParameters();
         Attributes attr = new Attributes();
         Object type = params.get("type");
-        attr.add("type", type == null ? "text" : (String) type)
+        attr.add("type", type == null ? "text" : type.toString())
                 .addDefaultToEmpty("name", params.get("name"))
                 .addIfExists("size", params.get("size"))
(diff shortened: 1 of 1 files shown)


plugins/javatemplates/src/main/java/org/apache/struts2/views/java/simple/TextFieldHandler.java
```diff
@@ -31,8 +31,8 @@
     public void generate() throws IOException {
         Map<String, Object> params = context.getParameters();
         Attributes attr = new Attributes();
-
-        attr.add("type", "text")
+        Object type = params.get("type");
+        attr.add("type", type == null ? "text" : (String) type)
```
lukaszlenart: Maybe it would be better to use a safe way, ie. `String.valueOf(type)`?

gregh3269: I did use originally .valueOf and then checking the other handlers they all used the cast, so matched what they use, for code style.

lukaszlenart: Maybe you could refactor the rest as well and use `String.valueOf` or leave it as is, up to you :)

gregh3269: Object type = params.get("type");
attr.add("type", type == null ? "text" : type.toString())

valueOf checks for nulls and calls .toString can skip the check?

lukaszlenart: You must check for `null` because `String.valueOf(null)` will return a String "null" :\ I think the current solution is ok :)

PR #87, opened 2016-02-15
WW-4558 contentType override ignored for JSONInterceptor

victorsosa:
From now on:

The "accept" request header parameter must be "application/json" or "application/json-rpc"

Also the default encoding is "UTF-8"


plugins/json/src/main/java/org/apache/struts2/json/JSONInterceptor.java
```diff
@@ -70,17 +70,22 @@
...
         HttpServletResponse response = ServletActionContext.getResponse();
-        String contentType = request.getHeader("content-type");
-        if (contentType != null) {
+        
+        //parameter wasn't set by the interceptor
+        if (accept == null) {
+            accept = request.getHeader("accept");
+        }
+        
+        if (accept != null) {
```
victorsosa: This can be remove, it is not needed anymore, we are using accept parameter

ke4qqq:
Closing to test travis integration - reopen shortly.

PR #96, opened 2016-05-22
WW-4634 for support-2-3

victorsosa:
Cherry-pick PR for support-2-3


lukaszlenart:
As far I understand the issue was in 2.5 only, 2.3 works as expected.

victorsosa:
But it is using the parameter align instead of a css class. the align parameter is deprecated in html 5.

lukaszlenart:
That's why people should migrate to 2.5 ;-) I think we can ask reported if he wants the same in 2.3

victorsosa:
OK

victorsosa:
PR remove, it wasn't requested

PR #97, opened 2016-05-25
WW-4636 File upload error message always in default language

victorsosa:
File upload error message always in default English language.

Content negotiation is a mechanism defined in the HTTP specification is
used in the code; now using the 'Accept-Language: us' in the request the language is choose.

The bug was that it was always using the English language for the
message.


aleksandr-m:
First of all, there is [some chance](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java#L71) that `defaultLocale` isn't set to English.

Can we get locale from the action context instead of request?

victorsosa:
@aleksandr-m  It is set to English in here: https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java#L62

get locale from the action context instead of request?... I will check that

aleksandr-m:
@victorsosa, @lukaszlenart Please close this one. WW-4636 is fixed in https://github.com/apache/struts/pull/113.

victorsosa:
done

PR #99, opened 2016-06-09
add allowed methods to handleUnknownAction

justinjmiller:
Prior to strict-method-invocation, the convention plugin would allow for executing results (JSP) without an associated Action via the ConventionUnknownHandler.handleUnknownAction method.  This continues to work with SMI disabled but when enabled the allowedMethods list in the ActionConfig does not get populated which causes a 404.  There was already a member Set hard coded to the most common methods to which I added "index" used by the REST plugin.  

Ideally, line 224 would use the globalAllowedMethods set from the PackageConfig instead of a hard coded list but there is currently no associated getter method and I figured it would be better to discuss adding that functionality via the PR.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/entities/PackageConfig.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/entities/PackageConfig.java
@@ -346,6 +346,13 @@ public class PackageConfig extends Located implements Comparable, Serializable, 
         return globalExceptionMappingConfigs;
     }
 
+    /**
+     * gets the GlobalAllowedMethods local to this package
+     *
+     * @return a Set of method names allowed to be executed if strict method invocation is enabled
+     */
+    public Set<String> getGlobalAllowedMethods() { return globalAllowedMethods; }
+
     public boolean isStrictMethodInvocation() {
         return strictMethodInvocation;
     }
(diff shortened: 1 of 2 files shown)


cnenning:
> Ideally, line 224 would use the globalAllowedMethods set from the PackageConfig instead of a hard coded list but there is currently no associated getter method and I figured it would be better to discuss adding that functionality via the PR.

IMHO a getter `getGlobalAllowedMethods()` can just be added to `PackageConfig`.

@lukaszlenart Do you see reasons not to add the getter?

justinjmiller:
Agreed.  I added it.

cnenning:
Please create a jira ticket, too. We usually document all changes in jira to generate change logs.

justinjmiller:
done - https://issues.apache.org/jira/browse/WW-4643

core/src/main/java/com/opensymphony/xwork2/config/entities/PackageConfig.java
```diff
@@ -346,6 +346,13 @@ public boolean isNeedsRefresh() {
         return globalExceptionMappingConfigs;
     }
 
+    /**
+     * gets the GlobalAllowedMethods local to this package
+     *
+     * @return a Set of method names allowed to be executed if strict method invocation is enabled
+     */
+    public Set<String> getGlobalAllowedMethods() { return globalAllowedMethods; }
```
lukaszlenart: Can this return unmodifiable collections?

lukaszlenart: Ok, I will do it after merge

PR #101, opened 2016-06-21
sync up execution ids with top-level Apache pom, otherwise the sources are deployed twice during release

pbruski:


lukaszlenart:
Please create an issue in JIRA and link here - we use this to prepare version notes.

lukaszlenart:
Is this the same PR as #100? I see the same changes

pbruski:
I've accidentally mixed up two PRs in #100 .
The issue is: https://issues.apache.org/jira/browse/WW-4653

lukaszlenart:
Szl@g! I have merged the wrong PR :\

PR #102, opened 2016-06-21
action parameters should be included when building the URL to action

pbruski:
At least that's what Webwork used to do


lukaszlenart:
Please register an issue in JIRA and link here

pbruski:
Registered: https://issues.apache.org/jira/browse/WW-4654

PR #103, opened 2016-06-24
WW-4650 Json deserialization does not work in 2.5.1

victorsosa:
Fix for WW-4650 broked by WW-4558 and test cases changes too.


lukaszlenart:
This looks good!

PR #104, opened 2016-06-27
WW-4620 ParametersInterceptor should check collection index to against DOS

victorsosa:
ParametersInterceptor should check collection index to avoid DOS

Check the parameters map to have only 255 objects to avoid DOS.

https://dzone.com/articles/spring-initbinder-for-handling-large-list-of-java


victorsosa:
A second thought  max of 512 could be a better number.

victorsosa:
this fix doesn't make sense

PR #105, opened 2016-06-28
[WW-4620] Improve XWorkListPropertyAccessor to against DOS attack

quaff:


core/src/main/java/com/opensymphony/xwork2/ognl/accessor/XWorkListPropertyAccessor.java
```diff
@@ -36,6 +36,7 @@
  * this class will create the necessary blank JavaBeans.
  *
  * @author Gabriel Zimmerman
+ * @author Yanming Zhou <zhouyanming@gmail.com>
```
aleksandr-m: Please remove author tag. In general you [should refrain from doing so](http://struts.apache.org/releases.html#clarifications). Why? [Because](http://directory.apache.org/apacheds/coding-standards.html#classinterface-headers).

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/XWorkListPropertyAccessor.java
```diff
@@ -45,6 +45,12 @@
     private ObjectFactory objectFactory;
     private ObjectTypeDeterminer objectTypeDeterminer;
     private OgnlUtil ognlUtil;
+    private int autoGrowCollectionLimit = 255;
+    
+    @Inject(value="java.util.Collection.autoGrowCollectionLimit", required = false)
```
lukaszlenart: Is it a predefined constant? I mean, does it exist somewhere?

quaff: It doesn't exists anywhere else.

lukaszlenart: Hmm... I would add `struts.` prefix, though. The name suggests that it is something Java specific.

lukaszlenart:
ping :)

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/XWorkListPropertyAccessor.java
quaff: It's OK.

lukaszlenart: > It's OK.

Do you mean you will change this?

quaff: I changed it with prefix "xwork"

PR #107, opened 2016-07-26
Change to the documentation

andrea-ligios:
To make it clear that the behavior is not limited to validation errors, but also extendend to conversion errors and (most important) to action errors, that will be stored and retrieved by the Message Store Interceptor during a redirection, and that will generate an undesired result.


lukaszlenart:
👍

PR #108, opened 2016-08-04
ConversionErrorInterceptor to extend MethodFilterIntercept

gregh3269:
See WW-4676

Cheers

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ConversionErrorInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ConversionErrorInterceptor.java
@@ -75,7 +75,7 @@ import java.util.Map;
  *
  * @author Jason Carreira
  */
-public class ConversionErrorInterceptor extends AbstractInterceptor {
+public class ConversionErrorInterceptor extends MethodFilterInterceptor {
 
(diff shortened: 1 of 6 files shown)


lukaszlenart:
Some tests failed https://builds.apache.org/job/Struts-JDK7-pull-request/209/org.apache.struts$struts2-core/testReport/

gregh3269:
OK, looks like it needs a doIntercept(..) rather than in Intercept().

lukaszlenart:
Still two tests are failing ... I think it would be better to setup `excludeMethods`/`includeMethods` in test and call`intercept()`

gregh3269:
Revert "WW-4628: proper decoding of parameters in query-string"

Sorry, this was a bad by not really understanding git and git making
automatic commits, it should not have modified any other code.  Especially not the revert.

How would this be fixed now?

cnenning:
> How would this be fixed now?

I fixed it already just by running `git cherry-pick`. (there is a jira comment about it)

I feared that monstrous issue ([WW-4628](https://issues.apache.org/jira/browse/WW-4628)) would get another round of discussion and was relieved that it was just an accident :wink:

gregh3269:
Great thanks!

lukaszlenart:
@gregh3269 the best options is to not perform those reverse merges (like b6be359) if there is no conflicts.

PR #111, opened 2016-09-16
replaced String constants for marking status with enum

sdutry:


lukaszlenart:
Could you register an issue in [JIRA](https://issues.apache.org/jira/browse/WW)? We use it to track changes and prepare Version Notes

sdutry:
registerd jira issue [WW-4689](https://issues.apache.org/jira/browse/WW-4689)

PR #114, opened 2016-11-21
Virtual file representation

lukaszlenart:
WW-4717 adds virtual file representation to allow implement `MultiPartRequest` on Google AppEngine


cnenning:
Looks good to me 👍

PR #117, opened 2017-02-05
[WW-4694] annotation processing improved

yasserzamani:
With these changes, [AnnotationUtils](https://github.com/apache/struts/blob/master/core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java) can navigate around proxies, superclasses and interfaces.
known issue: for keep it simple, currently it does not support [Bridge Methods](https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html).

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/DefaultWorkflowInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/DefaultWorkflowInterceptor.java
@@ -207,7 +207,7 @@ public class DefaultWorkflowInterceptor extends MethodFilterInterceptor {
      */
     protected String processInputConfig(final Object action, final String method, final String currentResultName) throws Exception {
         String resultName = currentResultName;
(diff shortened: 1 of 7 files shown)


lukaszlenart:
I have one general issue: indentation in some files are too big, it should be `4 spaces`

yasserzamani:
Thank you. I got it and will fix it by update my eclipse Preferences.

And what about the new codes and comments? some of them are copied from `Spring` framework which itself has `Apache License 2.0` license like Struts2. Do you recommend to rewrite them to something different with same functionality or no, those do not cause any legal issues?

And finally, these changes are needed to fix [WW-4472]. Could you please add this issue id to the title to automatically linking and resolving that issue too?

Thanks in advance!

lukaszlenart:
Cool, this look great and I don't mind copping code from other libraries if they match license :)

Regarding issue linkage I think you can update the description and put an issue ID there, e.g. `Fixes WW-4472` should be enough

PR #118, opened 2017-02-12
[WW-4105] OgnlUtil improved in order to only setting properties defined

yasserzamani:
With these changes, [OgnlUtil](https://github.com/apache/struts/blob/master/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java) can set only properties defined in the given "editable" class (or interface) which can be used e.g. in chaining interceptor to skip proxy properties of proxied chained actions, that avoid breaking upper level technologies like Spring Proxying.

ChainingInterceptor equipped with an optional param (editableClass) which user can set to a custom class (or interface) fully qualified name to restrict property setting like below:
```xml
<interceptor-ref name="chainStack">
	<param name="chain.editableClass">me.myname.myactions.MyClassOrInterface</param>
</interceptor-ref>
```
**It is useful when chained actions are unknown proxies created by an unknown object factory and we do not want to copy upper level proxy information which usually breaks their functionality** 👌

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
@@ -161,7 +161,20 @@ public class ChainingInterceptor extends AbstractInterceptor {
         Map<String, Object> ctxMap = invocation.getInvocationContext().getContextMap();
         for (Object object : list) {
             if (shouldCopy(object)) {
-                reflectionProvider.copy(object, invocation.getAction(), ctxMap, prepareExcludes(), includes);
+                String actionConfigClassName = invocation.getProxy().getConfig().getClassName();
+                Object action = invocation.getAction();
+                if (action.getClass().getName().equals(actionConfigClassName))
+                    reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, null);
+                else {	//only setting properties defined in the given config class e.g. to skip proxy properties (WW-4105)
+                    Class <?> editable;
+                    try {
+                        editable = Class.forName(actionConfigClassName);
+                    } catch (ClassNotFoundException e) {
+                        LOG.warn("An unexpected state. This may be due to a bug. Please report this: ClassNotFoundException" + actionConfigClassName, e);
+                        return;
+                    }
+                    reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
+                }            		
             }
         }
     }
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
@@ -125,6 +125,8 @@ public class ChainingInterceptor extends AbstractInterceptor {
 
     protected Collection<String> includes;
     protected ReflectionProvider reflectionProvider;
+    
+    protected String editableClass;
 
     @Inject
     public void setReflectionProvider(ReflectionProvider prov) {
@@ -161,20 +163,18 @@ public class ChainingInterceptor extends AbstractInterceptor {
         Map<String, Object> ctxMap = invocation.getInvocationContext().getContextMap();
         for (Object object : list) {
             if (shouldCopy(object)) {
-                String actionConfigClassName = invocation.getProxy().getConfig().getClassName();
                 Object action = invocation.getAction();
-                if (action.getClass().getName().equals(actionConfigClassName))
-                    reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, null);
-                else {	//only setting properties defined in the given config class e.g. to skip proxy properties (WW-4105)
-                    Class <?> editable;
+                Class <?> editable = null;
+                if (null != editableClass && !action.getClass().getName().equals(editableClass)) {
+                    //only setting properties defined in the given editable class e.g. to skip proxy properties (WW-4105)
                     try {
-                        editable = Class.forName(actionConfigClassName);
+                        editable = Class.forName(editableClass);
                     } catch (ClassNotFoundException e) {
-                        LOG.warn("An unexpected state. This may be due to a bug. Please report this: ClassNotFoundException" + actionConfigClassName, e);
-                        return;
+                        //warn and continue without any property restriction
+                        LOG.warn("Restrict property setting failed: " + editableClass + " class not found", e);
                     }
-                    reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
                 }            		
+                reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
             }
         }
     }
@@ -270,4 +270,22 @@ public class ChainingInterceptor extends AbstractInterceptor {
         this.includes = includes;
     }
 
+    /**
+     * Sets the class (or interface) to restrict property setting
+     * i.e. only setting properties defined in the given "editable" class (or interface).
+     *
+     * @param editableClass the class (or interface) fully qualified name
+     */
+    public void setEditableClass(String editableClass) {
+        this.editableClass = editableClass;
+    }
+
+    /**
+     * Gets the class (or interface) which restricts property setting
+     *
+     * @return the class (or interface) fully qualified name
+     */
+    public String getEditableClass() {
+        return editableClass;
+    }
 }
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlReflectionProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlReflectionProvider.java
@@ -70,7 +70,12 @@ public class OgnlReflectionProvider implements ReflectionProvider {
 
     public void copy(Object from, Object to, Map<String, Object> context,
             Collection<String> exclusions, Collection<String> inclusions) {
-        ognlUtil.copy(from, to, context, exclusions, inclusions);
+        copy(from, to, context, exclusions, inclusions, null);
+    }
+
+    public void copy(Object from, Object to, Map<String, Object> context,
+            Collection<String> exclusions, Collection<String> inclusions, Class<?> editable) {
+        ognlUtil.copy(from, to, context, exclusions, inclusions, editable);
     }
 
     public Object getRealTarget(String property, Map<String, Object> context, Object root)
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -415,7 +415,7 @@ public class OgnlUtil {
             }
         });
     }
-    
+
     private void checkEnableEvalExpression(Object tree, Map<String, Object> context) throws OgnlException {
         if (!enableEvalExpression && isEvalExpression(tree, context)) {
             throw new OgnlException("Eval expressions/chained expressions have been disabled!");
@@ -441,9 +441,37 @@ public class OgnlUtil {
      *                   note if exclusions AND inclusions are supplied and not null nothing will get copied.
      */
     public void copy(final Object from, final Object to, final Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions) {
+        copy(from, to, context, exclusions, inclusions, null);
+    }
+
+    /**
+     * Copies the properties in the object "from" and sets them in the object "to"
+     * only setting properties defined in the given "editable" class (or interface)
+     * using specified type converter, or {@link com.opensymphony.xwork2.conversion.impl.XWorkConverter} if none
+     * is specified.
+     *
+     * @param from       the source object
+     * @param to         the target object
+     * @param context    the action context we're running under
+     * @param exclusions collection of method names to excluded from copying ( can be null)
+     * @param inclusions collection of method names to included copying  (can be null)
+     *                   note if exclusions AND inclusions are supplied and not null nothing will get copied.
+     * @param editable the class (or interface) to restrict property setting to
+     */
+    public void copy(final Object from, final Object to, final Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions, Class<?> editable) {
         if (from == null || to == null) {
             LOG.warn("Attempting to copy from or to a null source. This is illegal and is bein skipped. This may be due to an error in an OGNL expression, action chaining, or some other event.");
             return;
+        }
+
+        Class<?> toClass = to.getClass();
+        if (editable != null) {
+            if (!editable.isInstance(to)) {
+                LOG.warn("Target class [" + to.getClass().getName() +
+                        "] not assignable to Editable class [" + editable.getName() + "]");
+                return;
+            }
+            toClass = editable;
         }
 
         TypeConverter converter = getTypeConverterFromContext(context);
@@ -457,7 +485,7 @@ public class OgnlUtil {
 
         try {
             fromPds = getPropertyDescriptors(from);
-            toPds = getPropertyDescriptors(to);
+            toPds = getPropertyDescriptors(toClass);
         } catch (IntrospectionException e) {
             LOG.error("An error occurred", e);
             return;
@@ -629,7 +657,7 @@ public class OgnlUtil {
         } else {
             throw new IllegalArgumentException("Cannot find type converter in context map");
         }
-        */
+         */
         return defaultConverter;
     }
 
--- a/core/src/main/java/com/opensymphony/xwork2/util/reflection/ReflectionProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/reflection/ReflectionProvider.java
@@ -72,6 +72,22 @@ public interface ReflectionProvider {
      *                   note if exclusions AND inclusions are supplied and not null nothing will get copied.
      */
     void copy(Object from, Object to, Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions);
+
+    /**
+     * Copies the properties in the object "from" and sets them in the object "to"
+     * only setting properties defined in the given "editable" class (or interface)
+     * using specified type converter, or {@link com.opensymphony.xwork2.conversion.impl.XWorkConverter} if none
+     * is specified.
+     *
+     * @param from       the source object
+     * @param to         the target object
+     * @param context    the action context we're running under
+     * @param exclusions collection of method names to excluded from copying ( can be null)
+     * @param inclusions collection of method names to included copying  (can be null)
+     *                   note if exclusions AND inclusions are supplied and not null nothing will get copied.
+	 * @param editable the class (or interface) to restrict property setting to
+     */
+    void copy(Object from, Object to, Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions, Class<?> editable);
     
     /**
      * Looks for the real target with the specified property given a root Object which may be a
--- a/core/src/test/java/com/opensymphony/xwork2/TestSubBean.java
+++ b/core/src/test/java/com/opensymphony/xwork2/TestSubBean.java
@@ -1,0 +1,14 @@
+package com.opensymphony.xwork2;
+
+public class TestSubBean extends TestBean {
+
+    private String issueId;
+
+    public String getIssueId() {
+        return issueId;
+    }
+
+    public void setIssueId(String issueId) {
+        this.issueId = issueId;
+    }
+}
--- a/core/src/test/java/com/opensymphony/xwork2/interceptor/ChainingInterceptorTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/interceptor/ChainingInterceptorTest.java
@@ -17,7 +17,6 @@ package com.opensymphony.xwork2.interceptor;
 
 import com.mockobjects.dynamic.Mock;
 import com.opensymphony.xwork2.*;
-import com.opensymphony.xwork2.config.entities.ActionConfig;
 import com.opensymphony.xwork2.util.ValueStack;
 
 import java.util.*;
@@ -34,15 +33,11 @@ public class ChainingInterceptorTest extends XWorkTestCase {
     ChainingInterceptor interceptor;
     Mock mockInvocation;
     ValueStack stack;
-    ActionConfig config;
-    Mock proxy;
 
 
     public void testActionErrorsCanBeAddedAfterChain() throws Exception {
         SimpleAction action1 = new SimpleAction();
         SimpleAction action2 = new SimpleAction();
-        config = new ActionConfig.Builder("", "action", action1.getClass().getName()).build();
-        proxy.matchAndReturn("getConfig", config);
         action1.addActionError("foo");
         mockInvocation.matchAndReturn("getAction", action2);
         stack.push(action1);
@@ -62,8 +57,6 @@ public class ChainingInterceptorTest extends XWorkTestCase {
     public void testActionErrorsNotCopiedAfterChain() throws Exception {
         SimpleAction action1 = new SimpleAction();
         SimpleAction action2 = new SimpleAction();
-        config = new ActionConfig.Builder("", "action", action1.getClass().getName()).build();
-        proxy.matchAndReturn("getConfig", config);
         action1.addActionError("foo");
         mockInvocation.matchAndReturn("getAction", action2);
         stack.push(action1);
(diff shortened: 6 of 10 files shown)


core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
```diff
@@ -160,7 +160,20 @@ private void copyStack(ActionInvocation invocation, CompoundRoot root) {
...
+                String actionConfigClassName = invocation.getProxy().getConfig().getClassName();
+                Object action = invocation.getAction();
+                if (action.getClass().getName().equals(actionConfigClassName))
+                    reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, null);
+                else {	//only setting properties defined in the given config class e.g. to skip proxy properties (WW-4105)
+                    Class <?> editable;
+                    try {
+                        editable =  Class.forName(actionConfigClassName);
+                    } catch (ClassNotFoundException e) {
+                        LOG.warn("An unexpected state. This may be due to a bug. Please report this: ClassNotFoundException" + actionConfigClassName);
```
lukaszlenart: What about logging exception here as well?

yasserzamani: It's better, thank you! done and committed.

yasserzamani: I also fixed Indentations to use 4 spaces instead of tab, both in my current PRs and my eclipse. Sorry, I'll be more careful about formatting in my next works

core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
```diff
@@ -160,7 +160,20 @@ private void copyStack(ActionInvocation invocation, CompoundRoot root) {
         Map<String, Object> ctxMap = invocation.getInvocationContext().getContextMap();
         for (Object object : list) {
             if (shouldCopy(object)) {
-                reflectionProvider.copy(object, invocation.getAction(), ctxMap, prepareExcludes(), includes);
+                String actionConfigClassName = invocation.getProxy().getConfig().getClassName();
```
yasserzamani: Please wait and do not merge until I check if getConfig and getClassName are always ready and are as expected e.g. when using convention or spring plugin!

yasserzamani: Please wait and do not merge! until I check if `getConfig` and `getClassName` are always ready and are as expected e.g. when using convention or spring plugin!

lukaszlenart:
@yasserzamani you should be able to mark this PR with a label `Ready for merge`

core/src/main/java/com/opensymphony/xwork2/interceptor/ChainingInterceptor.java
yasserzamani: As I guessed, this line does not work when action is a spring bean (`getClassName` returns the spring bean id, not fully qualified name of class).

Fortunately it works with actions created by convention plugin.

I will be working on it by checking all references which set or get class name. I also will add some unit tests to check if similar issues occurs in future.

yasserzamani: @lukaszlenart , I study a lot but, while user can equip any custom `ObjectFactory` to create actions (here user has many options like spring aop, javassist, jdk, cglib or any other library), it seems there is no way to determine which properties to set and which should be skipped **except asking from user itself**.

While there are `includes` and `excludes` available for user, but for save user from a lot of typing and also more readability, **what do you think about featuring a new optional parameter named `editableClass` like below?**
```xml
<result type="chain">
    <param name="actionName">myOriginalAction</param>
    <param name="editableClass">me.myname.myactions.MyOriginalAction</param>
</result>
```
or
```xml
<interceptor-ref name="chainStack">
    <param name="chain.editableClass">me.myname.myactions.MyOriginalAction</param>
</interceptor-ref>
```
After this featuring, rest of codes of this PR will work and fixes issue in a clean gracefully way.

lukaszlenart: What `editableClass` means here? Shouldn't it be a `originClass` or `targetClass`?

yasserzamani: > editable - the class (or interface) to restrict property setting to i.e. only setting properties defined in the given "editable" class (or interface))
I learnt this name from Spring docs. User also can set a class or interface which is not source class nor target class but an encapsulation of desired properties inside it.

lukaszlenart: Hm... ok, let's do it :)

yasserzamani: thank you :heart_eyes:

yasserzamani:
😍 PR re-based with last master and ChainingInterceptor equipped with an optional param (editableClass) which user can set to a custom class (or interface) fully qualified name to restrict property setting like below:
```xml
<interceptor-ref name="chainStack">
	<param name="chain.editableClass">me.myname.myactions.MyClassOrInterface</param>
</interceptor-ref>
```
It is useful when chained actions are unknown proxies created by an unknown object factory and we do not want to copy upper level proxy information which usually breaks their functionality 👌

yasserzamani:
@lukaszlenart , I am not able to mark this PR with a label `Ready for merge`. I think because I should not have write access to this repository.

lukaszlenart:
ech... will ask Infra

lukaszlenart:
only someone with write access can do it, so let's assume that adding :+1: by the author means `Ready for merge` - I will update docs

yasserzamani:
I improved this PR with an interesting Unit Test which prevents similar issues occurrence in future i.e. it checks if `chaining Spring AOPed actions` always works. This Unit Test reproduce the problem and fails in current code which verify the issue [WW-4105] existence.

yasserzamani:
@lukaszlenart , this could be postponed to resolution of [WW-4751](https://issues.apache.org/jira/browse/WW-4751).

lukaszlenart:
Sure, feel free to implement that :)

yasserzamani:
@lukaszlenart , I love to implement that but as changes may occur in several places from Core to Plugins e.g. JSONResult, I decided to get your agreement and confirmation if you also agree that Struts2 needs this improvement. Likely everywhere Struts2 works with Action's fields, we will need a change. If you agree, I can start carefuly improve there  👌 thank you for your confidence.

yasserzamani:
@lukaszlenart , another thing where I need your recommendation is, after this improvement, all users have to convert from
```xml
<action name="myAction" class="myBeanName">
```
to
```xml
<action name="myAction" class="mypackage.MyAction" bean="myBeanName">
```
in their existing projects which worries me. 😢 What do you do in such changes? thank you.

lukaszlenart:
Unacceptable ... what about annotations or conventions? Also it would be good to have the attribute `bean` optional.

yasserzamani:
@lukaszlenart , Yes, surely attribute `bean` is, will be and should be optional. Also there are no problem with actions which used attribute `class` as a class not a bean name. But unfortunately, Struts2 uses attribute `class` also for a bean name rather than a separate attribute which causes not backward compatible of this improvement :(

About Annotations I know they works without any change. For conventions and wildcards and etc I should see.

Anyway, in case of when action are beans inside user's object factory, we do not have any way except force user to change config and specify real class , Or add another optional attribute and add more conditional and not general codes for this improvement.

What is your recommendation in such changes? thanks.

lukaszlenart:
so if there is no `bean` attribute, it should fallback to an old behaviour, though...

yasserzamani:
@lukaszlenart , thank you, so I will add only one attribute named `bean`. if specified, we use that improvement, elsewhere current behavior. i.e. we will keep old behavior but with some specific if-else statements rather than more clean and general codes. If you agree with this and my previous comment about several places changes, the please 👍 for confirmation then I start :)

lukaszlenart:
Sure, go ahead :) I know it would be better to have just one implementation but we can mark the old code as deprecated and remove it at some point.

aleksandr-m:
@yasserzamani *in case of when action are beans inside user's object factory, we do not have any way except force user to change config and specify real class*

Can you elaborate? It really looks weird to specify both bean and the class name.

yasserzamani:
@aleksandr-m , as an example, consider when action proxified with another technology like Spring and the user assumes that Struts2 only operates in it's borders, but currently, Struts2 does not know his borders about the action object e.g. it copies proxy information like advices when it chains actions, it generates json data from them in JSONResult, it copies ActionSupport's data when it chains which forced us to declare includes/excludes, In a security point of view, someone may successfully change proxy information of the action by an HTTP request with named parameters!

All of these can be avoided if Struts2 know *user desired config time class of the action*. Struts2 knows this except when user gives actions up to any object factory like Spring or any other custom object factory. in such situation, Stuts2 looses information about config time class name of the action. yes, Struts2 has the object, but there is no any clean way to check if it's a proxy and unwrap it to target class!

So, by attribute `bean`, the user specifies object, and by attribute `class`, specifies Struts2's borders i.e. Struts2 will exclude all subclasses above `class` and all superclasses under and including ActionSupport. I defined these as an improvement at [WW-4751](https://issues.apache.org/jira/browse/WW-4751).

aleksandr-m:
Let's break it down:

**chain**
First of all using `chain` is discouraged. Proxying the action itself is not the best practice too.

If the only viable solution to know target class is to declare a real class of the bean then it should be done in the chain interceptor not in the action configuration.

**json result**
What is the problem with generating proxy data into json? What if this is what is really needed?

**excluding `ActionSupport`**
If `ActionSupport` is excluded then its methods cannot be used in the JSP (e.g. `getText`). In case of `chain` action errors / messages won't be moved to the next action. Etc.

**overall**
*someone may successfully change proxy information of the action by an HTTP request with named parameters*
Are you sure? Can you provide some example?

*there is no any clean way to check if it's a proxy and unwrap it to target class*
How knowing the name of the real class helps in that case? What are you going to do with it?
If it is spring proxy then there are helper methods to get target class from the instance (e.g. `AopUtils`). If there is no *clean* way to do this in the Struts core utility class then it can be delegated to current object factory.

yasserzamani:
@lukaszlenart , @aleksandr-m , thank you for your comments. Today I did not have enough time to prepare more dangerous security issue example but until I prepare one in comming days, please consider the following example:

Simply, I created an action with json result and with Spring's Transactional annotation (please consider that I know transactional actions are not good practice but it's an example, the Struts2 user may use any unknown 3rd party annotation **assuming Struts2 only operates in his borders rather than overlapping other technologies like Spring**, the user may not get any exception or error and so, may not check the generated json, but Struts2 emits also his objects information to hackers by json result).

OK, this is the result of http://localhost:7780/springAction1
```json
{
  "advisors":[
    {
      "advice":{
        "transactionAttributeSource":{

        },
        "transactionManager":null
      },
      "adviceBeanName":"org.springframework.transaction.interceptor.TransactionInterceptor#0",
      "order":2147483647,
      "perInstance":true,
      "pointcut":{
        "classFilter":{

        },
        "methodMatcher":null,
        "runtime":false
      }
    }
  ],
  "callbacks":[    {    },    {    },    {    },    {    },    {    },    {    },    {    }
  ],
  "exposeProxy":false,
  "frozen":false,
  "preFiltered":true,
  "proxiedInterfaces":[

  ],
  "proxyTargetClass":true,
  "targetClass":"class me.zamani.yasser.ww_convention.springactions.springAction1",
  "targetSource":{
    "static":true,
    "target":{

    },
    "targetClass":"class me.zamani.yasser.ww_convention.springactions.springAction1"
  }
}
```

And this is the result of http://localhost:7780/springAction1?advisors[0].adviceBeanName=&advisors[0].order=-1&exposeProxy=true&preFiltered=false
```json
{
  "advisors":[
    {
      "advice":{
        "transactionAttributeSource":{

        },
        "transactionManager":null
      },
      "adviceBeanName":"",
      "order":-1,
      "perInstance":true,
      "pointcut":{
        "classFilter":{

        },
        "methodMatcher":null,
        "runtime":false
      }
    }
  ],
  "callbacks":[    {    },    {    },    {    },    {    },    {    },    {    },    {    }
  ],
  "exposeProxy":true,
  "frozen":false,
  "preFiltered":false,
  "proxiedInterfaces":[

  ],
  "proxyTargetClass":true,
  "targetClass":"class me.zamani.yasser.ww_convention.springactions.springAction1",
  "targetSource":{
    "static":true,
    "target":{

    },
    "targetClass":"class me.zamani.yasser.ww_convention.springactions.springAction1"
  }
}
```

As you see, I changed something :) and then if I re-get even without any query string, these changes are persistence :\

Furthermore, json result on Spring AOPed actions simply fails with:
```
org.apache.struts2.json.JSONException: org.apache.struts2.json.JSONException: org.apache.struts2.json.JSONException: java.lang.IllegalAccessException: Class org.apache.struts2.json.JSONWriter can not access a member of class org.springframework.aop.TruePointcut with modifiers "public"
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:269)
	at org.apache.struts2.json.JSONWriter.processCustom(JSONWriter.java:197)
	at org.apache.struts2.json.JSONWriter.process(JSONWriter.java:182)
	at org.apache.struts2.json.JSONWriter.value(JSONWriter.java:143)
	at org.apache.struts2.json.JSONWriter.write(JSONWriter.java:110)
	at org.apache.struts2.json.JSONUtil.serialize(JSONUtil.java:194)
	at org.apache.struts2.json.JSONResult.createJSONString(JSONResult.java:222)
	at org.apache.struts2.json.JSONResult.execute(JSONResult.java:196)
	at com.opensymphony.xwork2.DefaultActionInvocation.executeResult(DefaultActionInvocation.java:373)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:277)
	at org.apache.struts2.interceptor.debugging.DebuggingInterceptor.intercept(DebuggingInterceptor.java:253)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.DefaultWorkflowInterceptor.doIntercept(DefaultWorkflowInterceptor.java:177)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.validator.ValidationInterceptor.doIntercept(ValidationInterceptor.java:260)
	at org.apache.struts2.interceptor.validation.AnnotationValidationInterceptor.doIntercept(AnnotationValidationInterceptor.java:73)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ConversionErrorInterceptor.doIntercept(ConversionErrorInterceptor.java:139)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ParametersInterceptor.doIntercept(ParametersInterceptor.java:133)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ParametersInterceptor.doIntercept(ParametersInterceptor.java:133)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.StaticParametersInterceptor.intercept(StaticParametersInterceptor.java:192)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.MultiselectInterceptor.intercept(MultiselectInterceptor.java:69)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.DateTextFieldInterceptor.intercept(DateTextFieldInterceptor.java:115)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.CheckboxInterceptor.intercept(CheckboxInterceptor.java:88)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.FileUploadInterceptor.intercept(FileUploadInterceptor.java:246)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ModelDrivenInterceptor.intercept(ModelDrivenInterceptor.java:99)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ScopedModelDrivenInterceptor.intercept(ScopedModelDrivenInterceptor.java:139)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ChainingInterceptor.intercept(ChainingInterceptor.java:156)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.PrepareInterceptor.doIntercept(PrepareInterceptor.java:174)
	at com.opensymphony.xwork2.interceptor.MethodFilterInterceptor.intercept(MethodFilterInterceptor.java:98)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.I18nInterceptor.intercept(I18nInterceptor.java:122)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.interceptor.ServletConfigInterceptor.intercept(ServletConfigInterceptor.java:171)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.AliasInterceptor.intercept(AliasInterceptor.java:195)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at com.opensymphony.xwork2.interceptor.ExceptionMappingInterceptor.intercept(ExceptionMappingInterceptor.java:193)
	at com.opensymphony.xwork2.DefaultActionInvocation.invoke(DefaultActionInvocation.java:247)
	at org.apache.struts2.factory.StrutsActionProxy.execute(StrutsActionProxy.java:54)
	at org.apache.struts2.dispatcher.Dispatcher.serviceAction(Dispatcher.java:564)
	at org.apache.struts2.dispatcher.ExecuteOperations.executeAction(ExecuteOperations.java:81)
	at org.apache.struts2.dispatcher.filter.StrutsPrepareAndExecuteFilter.doFilter(StrutsPrepareAndExecuteFilter.java:143)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:243)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:210)
	at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:222)
	at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:123)
	at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:502)
	at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:171)
	at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:100)
	at org.apache.catalina.valves.AccessLogValve.invoke(AccessLogValve.java:953)
	at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:118)
	at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:408)
	at org.apache.coyote.http11.AbstractHttp11Processor.process(AbstractHttp11Processor.java:1041)
	at org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:603)
	at org.apache.tomcat.util.net.AprEndpoint$SocketProcessor.doRun(AprEndpoint.java:2430)
	at org.apache.tomcat.util.net.AprEndpoint$SocketProcessor.run(AprEndpoint.java:2419)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.apache.struts2.json.JSONException: org.apache.struts2.json.JSONException: java.lang.IllegalAccessException: Class org.apache.struts2.json.JSONWriter can not access a member of class org.springframework.aop.TruePointcut with modifiers "public"
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:269)
	at org.apache.struts2.json.JSONWriter.processCustom(JSONWriter.java:197)
	at org.apache.struts2.json.JSONWriter.process(JSONWriter.java:182)
	at org.apache.struts2.json.JSONWriter.value(JSONWriter.java:143)
	at org.apache.struts2.json.JSONWriter.array(JSONWriter.java:550)
	at org.apache.struts2.json.JSONWriter.process(JSONWriter.java:170)
	at org.apache.struts2.json.JSONWriter.value(JSONWriter.java:143)
	at org.apache.struts2.json.JSONWriter.add(JSONWriter.java:428)
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:254)
	... 75 more
Caused by: org.apache.struts2.json.JSONException: java.lang.IllegalAccessException: Class org.apache.struts2.json.JSONWriter can not access a member of class org.springframework.aop.TruePointcut with modifiers "public"
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:269)
	at org.apache.struts2.json.JSONWriter.processCustom(JSONWriter.java:197)
	at org.apache.struts2.json.JSONWriter.process(JSONWriter.java:182)
	at org.apache.struts2.json.JSONWriter.value(JSONWriter.java:143)
	at org.apache.struts2.json.JSONWriter.add(JSONWriter.java:428)
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:254)
	... 83 more
Caused by: java.lang.IllegalAccessException: Class org.apache.struts2.json.JSONWriter can not access a member of class org.springframework.aop.TruePointcut with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:109)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:261)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:253)
	at java.lang.reflect.Method.invoke(Method.java:599)
	at org.apache.struts2.json.JSONWriter.bean(JSONWriter.java:249)
	... 88 more
```

yasserzamani:
@aleksandr-m , Thank you for your time and comments; please let me know what do you think about below; I would like to be sure about usefulness before starting implementation.

> Are you sure? Can you provide some example?

Yes, I created an attack example with latest Struts2 but I think I'm not allowed to post details here so I emailed to security@struts.apache.org because the example is really can be harmful and can be applied in an almost common usage by Struts2 users. **The vulnerability is because of operating Struts inside other technologies borders!**

> How knowing the name of the real class helps in that case? What are you going to do with it?

Knowing that helps Struts2 to not operate inside other technologies borders which may arise some vulnerability as mentioned above.

> If it is spring proxy then there are helper methods to get target class from the instance (e.g. AopUtils). If there is no clean way to do this in the Struts core utility class then it can be delegated to current object factory.

Struts2 dependency of Spring is optional e.g. AopUtils is not available in core. Furthermore, Struts2 user has several options for proxy creator from cglib, jdk to any unknown third party.

> Proxying the action itself is not the best practice too.

Please see [This is useful, for example, if you wish to apply more complex AOP or Spring-enabled technologies, such as Acegi](https://struts.apache.org/docs/spring-plugin.html).

> What is the problem with generating proxy data into json? What if this is what is really needed?

User may not get any exception then may not check the json result but actual result may help hackers. If this is what is really needed, then we can provide an option for user.

> If ActionSupport is excluded then its methods cannot be used in the JSP (e.g. getText). In case of chain action errors / messages won't be moved to the next action. Etc.

By word excluding, I meant in sensitive places rather than complete exclusion. In case of chain or any not sensitive place, we should think about solution :)

yasserzamani:
I close this as more general, better, cleaner and robust resolution is implementing [WW-4751](https://issues.apache.org/jira/browse/WW-4751) improvement which I am working on :)

PR #125, opened 2017-03-24
Immutable context

lukaszlenart:
This PR uses the latest OGNL version which drops access to `#context` via a magical key and makes `OgnlContext` a bit immutable.

[OGNL 3.2 version notes](https://github.com/jkuhnert/ognl#release-notes---version-32)

Implements [WW-4927](https://issues.apache.org/jira/browse/WW-4927)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
@@ -18,8 +18,10 @@
  */
 package com.opensymphony.xwork2;
 
+import com.opensymphony.xwork2.conversion.TypeConverter;
 import com.opensymphony.xwork2.inject.Container;
 import com.opensymphony.xwork2.util.ValueStack;
+import ognl.OgnlContext;
 import org.apache.struts2.dispatcher.HttpParameters;
 
 import java.io.Serializable;
@@ -184,6 +186,10 @@ public class ActionContext implements Serializable {
      * @return the context map.
      */
     public Map<String, Object> getContextMap() {
+        Map<String, Object> context = getContext().context;
+        if (context instanceof OgnlContext) {
+            ((OgnlContext) context).put(TypeConverter.TYPE_CONVERTER_CONTEXT_KEY, ((OgnlContext) context).getTypeConverter());
+        }
         return context;
     }
 
--- a/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
@@ -84,11 +84,6 @@ public class ActionContext implements Serializable {
     public static final String LOCALE = "com.opensymphony.xwork2.ActionContext.locale";
 
     /**
-     * Constant for the action's type converter.
-     */
-    public static final String TYPE_CONVERTER = "com.opensymphony.xwork2.ActionContext.typeConverter";
-
-    /**
      * Constant for the action's {@link com.opensymphony.xwork2.ActionInvocation invocation} context.
      */
     public static final String ACTION_INVOCATION = "com.opensymphony.xwork2.ActionContext.actionInvocation";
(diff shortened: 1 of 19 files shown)


cnenning:
Sounds like a very good idea! A short check showed that my apps are not affected 😆

yasserzamani:
Below is my new design which may be helpful as an example for whom is affected. Those are about access to `%{#context['com.opensymphony.xwork2.dispatcher.HttpServletRequest'].requestURI}`.

I wrote following Filter which operates after struts-prepare but before struts-execute. It supplies an utility into request scope:
```
package utils;

import java.io.IOException;

import javax.servlet.Filter;
import javax.servlet.FilterChain;
import javax.servlet.FilterConfig;
import javax.servlet.ServletException;
import javax.servlet.ServletRequest;
import javax.servlet.ServletResponse;
import javax.servlet.http.HttpServletRequest;

import org.apache.struts2.StrutsStatics;
import com.opensymphony.xwork2.ActionContext;
import com.opensymphony.xwork2.util.ValueStack;

/**
 * @author zamani
 *
 */
public class MYStrutsPrepareFilter implements Filter {

	private MYUtils MYUtils;

	@Override
	public void init(FilterConfig filterConfig) throws ServletException {
		MYUtils = new MYUtils();
	}

	@Override
	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
			throws IOException, ServletException {

		ValueStack stack = ActionContext.getContext().getValueStack();
		stack.setValue("#request['MYUtils']", MYUtils);
		
		chain.doFilter(req, res);
	}

	@Override
	public void destroy() {
		MYUtils = null;
	}

	
	public class MYUtils {
		public String getRequestURI() {
			HttpServletRequest httpsr = ((HttpServletRequest) ActionContext.getContext()
					.get(StrutsStatics.HTTP_REQUEST));
			return httpsr.getRequestURI();
		}
	}
}
```

Then `"#request['MYUtils'].requestURI"` can be used in jsp, instead.

aleksandr-m:
Spoted same `#context['com.opensymphony.xwork2.dispatcher.HttpServletRequest']` expression [here](https://stackoverflow.com/q/44291034/1700321). @yasserzamani What do you use it for?

yasserzamani:
@aleksandr-m , not me but my workmate has used it as an ID for each jsp. I do not know his work details but in general, he then uses this ID to decide where he should place the result after an AJAX call. For example a parent jsp that have div1, div2 and div3 calls an AJAX then if result was from '/APP/jsp/customer/error.jsp' the it puts the result in div3, if result was from '/APP/jsp/customer/payment.jsp' the it puts the result in div1, if result was from '/APP/jsp/customer/free.jsp' the it puts the result in div2.

asfgit:
Can one of the admins verify this patch?

lukaszlenart:
@apache/struts-committers this is ready for another look.

core/src/main/java/com/opensymphony/xwork2/ActionContext.java
```diff
@@ -185,6 +187,10 @@ public void setContextMap(Map<String, Object> contextMap) {
      * @return the context map.
      */
     public Map<String, Object> getContextMap() {
+        Map<String, Object> context = getContext().context;
+        if (context instanceof OgnlContext) {
+            ((OgnlContext) context).put(TypeConverter.TYPE_CONVERTER_CONTEXT_KEY, ((OgnlContext) context).getTypeConverter());
```
aleksandr-m: What is the deal with the type converter?

lukaszlenart: Good question, probably it's a workaround :/ Need to re-think this.

lukaszlenart: Whooops... it can be thrown away, goot catch 👍

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -47,14 +49,35 @@
...
 
+    @Override
+    public Object setup(Map context, Object target, Member member, String propertyName) {
+        Object result = null;
+
+        if (isAccessible(context, target, member, propertyName)) {
+            AccessibleObject accessible = (AccessibleObject) member;
+
+            if (!accessible.isAccessible()) {
+                result = Boolean.TRUE;
```
yasserzamani: Will `result` be used in `restore` method as `state` parameter? if so I think it should be `Boolean.FALSE`.

lukaszlenart: Hm... this was copy/pasted from existing code but I think you are right.

yasserzamani:
LGTM :+1:  Thanks!

PR #127, opened 2017-03-29
Drop deprecated logging layer

sdutry:
WW-4763


lukaszlenart:
Did you notice that this targets Struts 2.6? I would rather postpone this

sdutry:
> Did you notice that this targets Struts 2.6

Oops, didn't notice that.

Best to just close this pull request then for now i guess.

PR #130, opened 2017-04-19
WW-3952: Credit card validator

lukaszlenart:
Implements a credit card validator
[WW-3952](https://issues.apache.org/jira/browse/WW-3952)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/validator/AnnotationValidationConfigurationBuilder.java
+++ b/core/src/main/java/com/opensymphony/xwork2/validator/AnnotationValidationConfigurationBuilder.java
@@ -122,6 +122,14 @@ public class AnnotationValidationConfigurationBuilder {
                         result.add(temp);
                     }
                 }
+                // Process CrediCardValidator
+                else if (a instanceof CreditCardValidator) {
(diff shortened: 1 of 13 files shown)


lukaszlenart:
This is ready for review

cnenning:
I hope the regex is stable 😉

lukaszlenart:
> I hope the regex is stable 

What do you mean by that? I have used this http://www.regular-expressions.info/creditcard.html

cnenning:
How often do credit card companies change their number-patterns in a way that the regex needs to be updated? 'guess it is rare enough.

lukaszlenart:
Good question... I have used those random card numbers in tests and looks ok (but we will see) https://www.paypalobjects.com/en_US/vhelp/paypalmanager_help/credit_card_numbers.htm

cnenning:
If patterns change users have another reason to upgrade to latest struts 😆

gregh3269:
Last time I did some credit card validation, I needed to use a java script version as the payment process obscured the field names by not having the "name" attribute (so it was not obvious when these were sent over https) and I was unable to do server side validation.

PR #131, opened 2017-04-20
WW-4210: Type conversion class

lukaszlenart:
This PR extends existing `TypeConversion` annotation to allow define converters as a class instead of a raw string.

Implements [WW-4210](https://issues.apache.org/jira/browse/WW-4210)


cnenning:
👍

PR #135, opened 2017-04-30
WW-4105 Unwraps Spring proxy in actions chain

yasserzamani:
Resolves [WW-4105](https://issues.apache.org/jira/browse/WW-4105).

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
+++ b/core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.opensymphony.xwork2.spring;
+
+import com.opensymphony.xwork2.util.ClassLoaderUtil;
+import org.apache.commons.lang3.reflect.MethodUtils;
+
+import java.lang.reflect.Proxy;
+
+/**
+ * <code>SpringUtils</code>
+ * <p>
+ * Various utility methods dealing with spring framework
+ * </p>
+ *
+ */
+public class SpringUtils {
+    /**
+     * Get the ultimate <em>target</em> object of the supplied {@code candidate}
+     * object, unwrapping not only a top-level proxy but also any number of
+     * nested proxies.
+     * <p>If the supplied {@code candidate} is a Spring proxy, the ultimate target of all
+     * nested proxies will be returned; otherwise, the {@code candidate}
+     * will be returned <em>as is</em>.
+     * @param candidate the instance to check (potentially a Spring AOP proxy;
+     * never {@code null})
+     * @return the target object or the {@code candidate} (never {@code null})
+     * @throws IllegalStateException if an error occurs while unwrapping a proxy
+     */
+    public static <T> T getUltimateTargetObject(Object candidate) {
+        try {
+            if (isAopProxy(candidate) &&
+                    implementsInterface(candidate.getClass(), "org.springframework.aop.framework.Advised")) {
+                Object targetSource = MethodUtils.invokeMethod(candidate, "getTargetSource");
+                Object target = MethodUtils.invokeMethod(targetSource, "getTarget");
+                return getUltimateTargetObject(target);
+            }
+        }
+        catch (Throwable ex) {
+            throw new IllegalStateException("Failed to unwrap proxied object", ex);
+        }
+        return (T) candidate;
+    }
+
+    /**
+     * Check whether the given object is a Spring proxy.
+     * @param object the object to check
+     */
+    public static boolean isAopProxy(Object object) {
+        Class<?> clazz = object.getClass();
+        return (implementsInterface(clazz, "org.springframework.aop.SpringProxy") &&
+                (Proxy.isProxyClass(clazz) || isCglibProxyClass(clazz)));
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
```diff
@@ -0,0 +1,89 @@
...
+import java.lang.reflect.Proxy;
+
+/**
+ * <code>SpringUtils</code>
+ * <p>
+ * Various utility methods dealing with spring framework
+ * </p>
+ *
+ */
+public class SpringUtils {
```
aleksandr-m: Maybe name it `ProxyUtil` or something like that. This class can be used to deal with known proxies and not only spring proxies.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param candidate the instance to check (potentially a Spring AOP proxy;
+     * never {@code null})
+     * @return the target object or the {@code candidate} (never {@code null})
+     * @throws IllegalStateException if an error occurs while unwrapping a proxy
+     */
+    public static <T> T getUltimateTargetObject(Object candidate) {
+        try {
+            if (isAopProxy(candidate) &&
+                    implementsInterface(candidate.getClass(), "org.springframework.aop.framework.Advised")) {
+                Object targetSource = MethodUtils.invokeMethod(candidate, "getTargetSource");
```
aleksandr-m: What is `targetSource` here?

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
```diff
@@ -0,0 +1,89 @@
...
+        return (T) candidate;
+    }
+
+    /**
+     * Check whether the given object is a Spring proxy.
+     * @param object the object to check
+     */
+    public static boolean isAopProxy(Object object) {
+        Class<?> clazz = object.getClass();
+        return (implementsInterface(clazz, "org.springframework.aop.SpringProxy") &&
```
aleksandr-m: Extract strings to constants.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
aleksandr-m: Another maybe. This can be turned to injectable bean so it can be easily overridable if needed. Wdyt?

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: It is what Spring himself does to unwrap it's proxies. Please see [AopTestUtils.java](https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/util/AopTestUtils.java)

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: Why someone may need to override such specific methods?

As all methods could be static, I thought normally it should be a static utils class rather than a bean object.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: Thank you! I will do in next commit.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: Thank you! I will do in next commit. However, I also try my effort to remove it to commons-lang if guys there accepted.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
aleksandr-m: Hmm... Never used `AopTestUtils`. But `Advised` extends `TargetClassAware`. Can `getTargetClass` just be invoked instead of two method calls? See [`AopUtils`](https://github.com/spring-projects/spring-framework/blob/master/spring-aop/src/main/java/org/springframework/aop/support/AopUtils.java).

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
aleksandr-m: To support other possible proxies. Or in case something changes in spring-aop.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: `getTargetClass` just returns a `Class` but getting thae target `Object` (i.e. actual action object) let us pass it directly to `ReflectionProvider` without any needed modification of `ReflectionProvider` codes.

core/src/main/java/com/opensymphony/xwork2/spring/SpringUtils.java
yasserzamani: I think we should update Struts itself in both cases when reported by a user.

lukaszlenart: `SpringUtils` doesn't sound good but I have no better name for now. Also it's the only case we have now so we can always refactor this code and introduce an interface with few implementation latter on.

For future reference: it would be good to have a `ProxyAware` interface in the `core` with a default implementation (that does nothing) and then in the Spring Plugin provide a real implementation.

yasserzamani: Thank you! Unfortunately I did not understand well. Do you mean that the user has to be careful to implement `ProxyAware` or inherit Spring Plugin's default implementation in his proxied actions? If so, there are two disadvantages, first he has to be careful about which actions may be proxied, and second he has to add Spring Plugin dependency or implement it manually.

lukaszlenart: no, no ... it was just for a future reference when we will want turn this into a bean

yasserzamani:
@lukaszlenart , I updated with @aleksandr-m reviews. I hope it is ready for merge if you pals agree.

lukaszlenart:
I think with can go ahead and merge this, @aleksandr-m any objections?

aleksandr-m:
@lukaszlenart I think it is ok.

PR #136, opened 2017-05-01
WW-4793 only add JBossFileManager when supported

sdutry:


core/src/main/java/org/apache/struts2/util/fs/JBossFileManager.java
```diff
@@ -210,4 +187,33 @@ private void addIfAbsent(List<URL> urls, URL fileUrl) {
         }
     }
 
+    public static class JBossFileManagerSupportUtil {
```
aleksandr-m: Why inner class? Why not just static methods in `JBossFileManager`?

sdutry: @aleksandr-m 
Good point, let me change that.

sdutry: @aleksandr-m 
Changed to static methods.
(don't know why i didn't just do that from the start)

sdutry:
Given the input on the issue, this pull request should not be merged.

PR #139, opened 2017-05-29
WW-4800 Executes aspects when chaining AOPed actions

yasserzamani:
These fix some bugs in my previous PR, #135 which passes the unwrapped object of the proxied action to the `reflectionProvider.copy`, then we miss aspects executions which are implemented inside the proxied object rather than the unwrapped one.

Also fix some bugs in unit tests of my previous PR and improves them.


lukaszlenart:
Thanks, merged :)

PR #141, opened 2017-06-12
WW-4799 Adding struts.date.format as a conversion format for DateConv…

dcabasson:
…erter

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
+++ b/core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
@@ -1,6 +1,9 @@
 package com.opensymphony.xwork2.conversion.impl;
 
+import com.opensymphony.xwork2.ActionContext;
+import com.opensymphony.xwork2.TextProvider;
 import com.opensymphony.xwork2.XWorkException;
+import com.opensymphony.xwork2.util.ValueStack;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
```diff
@@ -48,7 +51,28 @@ public Object convertValue(Map<String, Object> context, Object target, Member me
...
+                    String globalFormat = tp.getText(org.apache.struts2.components.Date.DATETAG_PROPERTY);
+
+                    // if tp.getText can not find the property then the
+                    // returned string is the same as input =
+                    // DATETAG_PROPERTY
+                    if (globalFormat != null
+                        && !org.apache.struts2.components.Date.DATETAG_PROPERTY.equals(globalFormat)) {
+                        globalDateFormat = new SimpleDateFormat(globalFormat, locale);
+                    }
+                }
```
lukaszlenart: Could you move this whole section into `getDateFormats` to avoid passing the `globalDateFormat` as a parameter? Also it would be good to extract this logic into a protected method to allow override, so the finally it should look like this:

```java
     DateFormat[] dfs = getDateFormats(globalDateFormat, locale);

...

private DateFormat[] getDateFormats(Locale locale) {
    ...
    DateFormat globalDateFormat = getGlobalDateFormat();
    ...
}

protected getGlobalDateFormat() {
     // your code here
}
```

lukaszlenart:
Please rebuild this

lukaszlenart:
I will fix the outstanding issues in another PR.

PR #146, opened 2017-06-21
add constant to control proxy member access

aleksandr-m:
See discussion in #142.


yasserzamani:
Thank you. Would we need same for support-2-3?

lukaszlenart:
I think it would be good to have it

yasserzamani:
Thanks! I am working on it.

PR #151, opened 2017-07-25
WW-4818 change default Multipart validation regex to comply with RFC1341

sdutry:


core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -88,7 +88,7 @@
      */
     public static final String REQUEST_POST_METHOD = "POST";
 
-    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart\\/form-data(; boundary=[\\-a-zA-Z0-9]{1,70})?";
+    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart/form-data(; boundary=[0-9a-zA-Z'()+_,\\-./:=?]{1,70})?";
```
lukaszlenart: You have to escape `.` as it now means `Any single character`

lukaszlenart: Also `-` has to be escaped as it is treated as a group separator

sdutry: not within the square brackets according to my knowledge

lukaszlenart: o! didn't know that :)

sdutry: you are correct about the `-` needing to be escaped , but that's the only one i escaped. (they're in the order they're mentioned in RFC1341 (so i moved the `-` backwards in the pattern)

lukaszlenart: There are two test cases that you can extend or another one - see `DispatcherTest`

sdutry: Ok, i'll add some tests to confirm the pattern is correct.
(Or if needed to correct it)

sdutry:
@lukaszlenart 
I added 2 simple tests.
- one containing all the special allowed characters
- another one containing a single not-allowed character

Please feel free to tell me what other test-cases you want added.
(for example, any specific characters you want tested?)

lukaszlenart:
Looks good 👍  LGTM :)

sdutry:
@lukaszlenart 
Am i allowed to merge this or is there more work/checks that needs to happen first?

lukaszlenart:
@sdutry yes, you are :) 
http://struts.apache.org/submitting-patches.html#how-to-merge-pull-requests

PR #153, opened 2017-07-31
WW-4827 Not fully initialized ObjectFactory tries to create beans

aleksandr-m:
Inject `Container` in constructor of the `ObjectFactory`.

Can someone test `cdi`, `osgi` and `plexus` plugins.

Better ideas how to fix the issue are welcome. :)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
@@ -49,7 +49,7 @@ public class ObjectFactory implements Serializable {
     private static final Logger LOG = LogManager.getLogger(ObjectFactory.class);
 
     private transient ClassLoader ccl;
-    private Container container;
+    private final Container container;
 
     private ActionFactory actionFactory;
     private ResultFactory resultFactory;
@@ -58,14 +58,14 @@ public class ObjectFactory implements Serializable {
     private ConverterFactory converterFactory;
     private UnknownHandlerFactory unknownHandlerFactory;
 
+    @Inject
+    public ObjectFactory(Container container) {
+        this.container = container;
+    }
+
     @Inject(value="objectFactory.classloader", required=false)
     public void setClassLoader(ClassLoader cl) {
         this.ccl = cl;
-    }
-    
-    @Inject
-    public void setContainer(Container container) {
-        this.container = container;
     }
 
     @Inject
--- a/core/src/main/java/com/opensymphony/xwork2/spring/SpringObjectFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/spring/SpringObjectFactory.java
@@ -16,6 +16,7 @@
 package com.opensymphony.xwork2.spring;
 
 import com.opensymphony.xwork2.ObjectFactory;
+import com.opensymphony.xwork2.inject.Container;
 import com.opensymphony.xwork2.inject.Inject;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
@@ -70,6 +71,11 @@ public class SpringObjectFactory extends ObjectFactory implements ApplicationCon
     @Inject(value = "enableAopSupport", required = false)
     public void setEnableAopSupport(String enableAopSupport) {
         this.enableAopSupport = BooleanUtils.toBoolean(enableAopSupport);
+    }
+
+    @Inject
+    public SpringObjectFactory(Container container) {
+        super(container);
     }
 
     /**
--- a/core/src/main/java/com/opensymphony/xwork2/spring/SpringProxyableObjectFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/spring/SpringProxyableObjectFactory.java
@@ -23,6 +23,9 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
 
+import com.opensymphony.xwork2.inject.Container;
+import com.opensymphony.xwork2.inject.Inject;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -37,6 +40,11 @@ public class SpringProxyableObjectFactory extends SpringObjectFactory {
     private static final Logger LOG = LogManager.getLogger(SpringProxyableObjectFactory.class);
 
     private List<String> skipBeanNames = new ArrayList<>();
+
+    @Inject
+    public SpringProxyableObjectFactory(Container container) {
+        super(container);
+    }
 
     @Override
     public Object buildBean(String beanName, Map<String, Object> extraContext) throws Exception {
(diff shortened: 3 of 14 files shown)


lukaszlenart:
No better idea and I think we can assume when you need to inject the `Container` you should always do it in a constructor.

yasserzamani:
To keep `backward-compatible` if is very important, I think we may have following:


```java
protected Object injectInternalBeans(Object obj) {
    if (obj != null) {
        if(container != null) container.inject(obj);
        else myPrivateInjectQueue.add(obj);
    }
    return obj;
}
   		
@Inject		
public void setContainer(Container container) {		
    this.container = container;
    if(null != this.container && myPrivateInjectQueue.getCount()>0)
    {
        foreach(Object obj in myPrivateInjectQueue) this.container.inject(obj);
        myPrivateInjectQueue.clear();
    }		
}
```

However I never tried and tested it but I can if you would like.

lukaszlenart:
I'm not sure if this is a good idea ... basically we need a `@PostConstruct` mechanism in Struts DI - this can be easily achieved by switching to Guice 3 and using CDI annotations.

lukaszlenart:
Hm... give me few minutes, maybe I will be able add support for it ;-)

lukaszlenart:
I have opened #155 - what do you think? @aleksandr-m can you test it locally as I do not have the exact setup.

aleksandr-m:
Nope, #155 doesn't seem have any effect.

Add following files to project, run under *nix, execute some action which lead to the JSP,  `localizedTextProvider` should not be `null` in system outs. 

xwork-conversion.properties;

    java.util.Date=com.DateConverter

JSP:
  
    <s:bean var="d" name="java.util.Date" />
    <s:property value="#d" />

DateConverter:
```
public class DateConverter extends StrutsTypeConverter {

    private LocalizedTextProvider localizedTextProvider;

    @Inject
    public void setLocalizedTextProvider(LocalizedTextProvider localizedTextProvider) {
        this.localizedTextProvider = localizedTextProvider;
    }

    @Override
    public Object convertFromString(Map context, String[] values, Class toClass) {
        System.out.println("DateConverter.convertFromString() localizedTextProvider=" +
                                      localizedTextProvider);
        return null;
    }

    @Override
    public String convertToString(Map context, Object obj) {
        System.out.println("DateConverter.convertToString() localizedTextProvider=" +
                                      localizedTextProvider);
        return "";
    }
}
```

lukaszlenart:
hm... but this example works for both branches - `master` and this:
```
DateConverter.convertToString() localizedTextProvider=com.opensymphony.xwork2.util.StrutsLocalizedTextProvider@342e1b8b
```

lukaszlenart:
(on OSX)

lukaszlenart:
One more question: JDK version?

aleksandr-m:
Ubuntu, oracle jdk8.

If following code prints `setContainer` before other setters (especially before `setInterceptorFactory`) it will work. On Ubuntu using oracle jdk8 `setContainer` is printed way below other setters and injection fails.

    Method[] methods = ObjectFactory.class.getDeclaredMethods();
    for (Method m : methods) {
        System.out.println(m);
    }

lukaszlenart:
See these screenshots

#### master


#### this branch


as you see, on the `master` branch the object isn't fully initialised but when `init()` method is called when using this branch, the object got fully initialised.

Maybe you have discovered another place where there is some work done in setter which should be moved into the `init()` method with the `PostInit` interface.

aleksandr-m:
It isn't about `XWorkConverter` per se. In #155 the injection of `ConversionPropertiesProcessor` still happens and `TypeConverterCreator` is injected next and `ObjectFactory#buildBean` is called and `container` is `null` in `ObjectFactory`.

BTW `ObjectFactory.class.getDeclaredMethods()` on Ubuntu, openJDK 7 prints `setContainer` as the last method.

lukaszlenart:
... but where the `ObjectFactory#buildBean` is called? It should only be called when building non-internal beans and when all the internal beans were already instantiated. Can you post a call trace?

lukaszlenart:
I have changed code a bit, can you try to test it on your side?

aleksandr-m:
With the new code in #155 the custom type converter isn't initialized at all. :(

Call trace of previous code in #155 
```

**Comment by lukaszlenart** [comment]
> With the new code in #155 the custom type converter isn't initialized at all. :(

w00t!

**Comment by lukaszlenart** [comment]
Would it be possible to share an application that replicates this behaviour? Sorry for bothering you but I want to understand what's going on :)

**Comment by lukaszlenart** [comment]
I don't see any other options now to merge this. @aleksandr-m could you prepare that example app? I will be able to investigate what's wrong with my solution ;-)

**Comment by aleksandr-m** [comment]
Sure. No problem.

Test project - https://github.com/aleksandr-m/struts2-objectfactory-container/

This commit adds custom date converter - https://github.com/aleksandr-m/struts2-objectfactory-container/commit/49406c1d8e816acd1f227d55db6729cd92a5de99.

This commit adds local copy of `ContainerImpl` in order to reproduce the issue in various environments - https://github.com/aleksandr-m/struts2-objectfactory-container/commit/e3fb324c56c8517491ea336f8682601ae33a5f0b

**Comment by lukaszlenart** [comment]
Thanks a lot :) I see now what's going on but have no idea why :\

**Comment by yasserzamani** [comment]
@aleksandr-m , I modified `ObjectFactory` as below

```java
...
    private List<Object> myPrivateInjectQueue;
...
    @Inject
    public void setContainer(Container container) {
        this.container = container;
        if(null != this.container && null!=myPrivateInjectQueue && myPrivateInjectQueue.size()>0)
        {
            for(Object obj : myPrivateInjectQueue) this.container.inject(obj);
            myPrivateInjectQueue.clear();
            myPrivateInjectQueue=null;
        }
    }
...
    protected Object injectInternalBeans(Object obj) {
        if (obj != null) {
            if(container != null) container.inject(obj);
            else {
                if(null==myPrivateInjectQueue)myPrivateInjectQueue=new ArrayList<Object>();
                myPrivateInjectQueue.add(obj);}
        }
        return obj;
    }

```
in your sample I got following ok result :)

```
moving setContainer method to the last element of the array
DateConverter.DateConverter()
moving setContainer method to the last element of the array
DateConverter.DateConverter()
DateConverter.convertToString() localizedTextProvider=com.opensymphony.xwork2.util.StrutsLocalizedTextProvider@3f6f10fa

```

aleksandr-m:
@yasserzamani  Looks like a hack :) I would rather leave as it is, then to introduce that. If container is needed then it **should** be injected in the constructor. Take a look at spring object factory.

yasserzamani:
:) Yes I know, just a bit worry about backward compatibility between minor versions.

I also am studying your sample and @lukaszlenart 's pr to understand what is going on as I think same issues may exists because it seems an object's injections is not an atomic operations! shouldn't be?

yasserzamani:
> Thanks a lot :) I see now what's going on but have no idea why :\

@lukaszlenart , as I debug, the trace is `object factory creation` then `set1 (inject it's 1st property)` then `set2 (inject it's 2nd property)` ... then ...  then `setK (inject it's Knd property)` then `your PostInit.init()` which reaches this object factory with null container ...  then `setN (inject it's Nnd property)`  finally `setContainer (inject it's container property)`.

As you see, if you would like to make it works, the `ObjectFactory` itself has to implement `PostInit` and do followings in it's `init()`:

```java
processRequired("struts-default-conversion.properties");
process("xwork-conversion.properties");
```

lukaszlenart:
The case is that on second container initialisation we do not create singletons upfront ... I have tried to change that but got some problems, not sure why. I must re-think how we initialise Struts internals ...

PR #156, opened 2017-08-01
WW-4834 Improve RegEx used to validate URLs

sdutry:
changed the regex according to https://github.com/apache/struts/commit/8df5a897f61f3ef45c36fdd9275e66669ae4516c#commitcomment-23414493

Can anyone tell me if the optional `#` on line 54 is needed?


lukaszlenart:
Ach... you did merge it :\ But this isn't a valid RegEx :(

PR #158, opened 2017-08-02
WW-4835: Configurable handlers

lukaszlenart:
This PR changes API a bit to allow pass current `ActionInvocation` to the REST content type handlers to allow implement different behaviour based on a current action.

[WW-4835](https://issues.apache.org/jira/browse/WW-4835)

### DIFF
--- a/plugins/rest/src/main/java/org/apache/struts2/rest/ContentTypeHandlerManager.java
+++ b/plugins/rest/src/main/java/org/apache/struts2/rest/ContentTypeHandlerManager.java
@@ -21,6 +21,7 @@
 
 package org.apache.struts2.rest;
 
+import com.opensymphony.xwork2.ActionInvocation;
 import org.apache.struts2.rest.handler.ContentTypeHandler;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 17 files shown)


cnenning:
Using `AbstractContentTypeHandler` to implement deprecated methods to log on WARN seems like a good idea.

I don't see where new parameter `ActionInvocation` is actually used. Is that yet to come?

lukaszlenart:
> I don't see where new parameter ActionInvocation is actually used. Is that yet to come?

Not yet :)

cnenning:
Do you want to further develop this PR or merge it first?

lukaszlenart:
Merge it first as I think those changes are less intrusive as the whole idea in my head ;-)

cnenning:
and what is your idea?

lukaszlenart:
With the `ActionInvocation` in place I can access an action which means the action can provide additional informations how to serialise/deserialize object/string via dedicated interfaces, e.g.

```java
public interface DateFormatHandler {
  Date toDate(String dateStr);
  String fromDate(Date date);
}
```

and when action will implement such interface, a handler can pass formatting a date to the action instead of doing this by itself. And many more :)

cnenning:
sounds great 👍

PR #159, opened 2017-08-04
WW-4834 Improve RegEx used to validate URLs

sdutry:


sdutry:
see #157 and #156

lukaszlenart:
Is it ready to merge? Does it contain the latest changes from #157?

sdutry:
@lukaszlenart 
i copied the regex after the changes from #157 .
So it should be the same. Feel free to check.

PR #162, opened 2017-08-22
return request HTTP method to fix problem for portlets with multipart validation

pheaber:
multipart validation added to Dispatcher in WW-4768 and JakartaMultiPartRequest in WW-4767 can't be passed within a portlet since PortletServletRequest.getMethod() only returns null


lukaszlenart:
Merged, thanks!

PR #164, opened 2017-08-30
WW-4846 Does not traverse/include proxy info in JSONResult

yasserzamani:


yasserzamani:
In second commit I added a control in such planning to also resolve [WW-4034: Allow to use custom JSONwriter](https://issues.apache.org/jira/browse/WW-4034)

PR #169, opened 2017-10-02
WW-4865: Add "requiredValue" parameter to Checkbox Struts Tag

burtonrhodes:
Added requiredValue to Checkbox Struts Tag as a parameter to give developer flexibility to use/not use "submit unchecked values" design pattern.  When requiredValue set to "false", unchecked checkbox elements will not be submitted with the form (i.e. no hidden input html elements will be created for the checkbox).  Default value is true for backwards compatibility.

[WW-4865](https://issues.apache.org/jira/browse/WW-4865)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/Checkbox.java
+++ b/core/src/main/java/org/apache/struts2/components/Checkbox.java
@@ -60,7 +60,7 @@ public class Checkbox extends UIBean {
     final public static String TEMPLATE = "checkbox";
 
     protected String fieldValue;
-    protected String requiredValue;
+    protected String submitUnchecked;
 
     public Checkbox(ValueStack stack, HttpServletRequest request, HttpServletResponse response) {
         super(stack, request, response);
@@ -76,10 +76,10 @@ public class Checkbox extends UIBean {
         } else {
             addParameter("fieldValue", "true");
         }
-        if (requiredValue != null) {
-            addParameter("requiredValue", findString(requiredValue));
+        if (submitUnchecked != null) {
+            addParameter("submitUnchecked", findString(submitUnchecked));
         } else {
-            addParameter("requiredValue", "true");
+            addParameter("submitUnchecked", "true");
         }
     }
 
@@ -93,7 +93,7 @@ public class Checkbox extends UIBean {
     }
 
     @StrutsTagAttribute(description="If set to false, unchecked elements will not be submitted with the form.", defaultValue="true")
-    public void setRequiredValue(String requiredValue) {
-        this.requiredValue = requiredValue;
+    public void setSubmitUnchecked(String submitUnchecked) {
+        this.submitUnchecked = submitUnchecked;
     }
 }
--- a/core/src/main/java/org/apache/struts2/components/Checkbox.java
+++ b/core/src/main/java/org/apache/struts2/components/Checkbox.java
@@ -60,6 +60,7 @@ public class Checkbox extends UIBean {
     final public static String TEMPLATE = "checkbox";
 
     protected String fieldValue;
+    protected String requiredValue;
 
     public Checkbox(ValueStack stack, HttpServletRequest request, HttpServletResponse response) {
         super(stack, request, response);
@@ -75,6 +76,11 @@ public class Checkbox extends UIBean {
         } else {
             addParameter("fieldValue", "true");
         }
+        if (requiredValue != null) {
+            addParameter("requiredValue", findString(requiredValue));
+        } else {
+            addParameter("requiredValue", "true");
+        }
     }
 
     protected Class getValueClassType() {
@@ -86,4 +92,8 @@ public class Checkbox extends UIBean {
         this.fieldValue = fieldValue;
     }
 
+    @StrutsTagAttribute(description="If set to false, unchecked elements will not be submitted with the form.", defaultValue="true")
+    public void setRequiredValue(String requiredValue) {
+        this.requiredValue = requiredValue;
+    }
 }
--- a/core/src/main/java/org/apache/struts2/components/Checkbox.java
+++ b/core/src/main/java/org/apache/struts2/components/Checkbox.java
@@ -79,7 +79,7 @@ public class Checkbox extends UIBean {
         if (submitUnchecked != null) {
             addParameter("submitUnchecked", findString(submitUnchecked));
         } else {
-            addParameter("submitUnchecked", "true");
+            addParameter("submitUnchecked", "false");
         }
     }
 
@@ -92,7 +92,7 @@ public class Checkbox extends UIBean {
         this.fieldValue = fieldValue;
     }
 
-    @StrutsTagAttribute(description="If set to false, unchecked elements will not be submitted with the form.", defaultValue="true")
+    @StrutsTagAttribute(description="If set to true, unchecked elements will be submitted with the form.", defaultValue="false")
     public void setSubmitUnchecked(String submitUnchecked) {
         this.submitUnchecked = submitUnchecked;
     }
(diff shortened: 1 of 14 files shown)


aleksandr-m:
Without other info the `requiredValue` attribute sounds like it is related to `required` attribute, which it isn't. Can you rename it to something less ambiguous.

BTW `fieldValue` should fix your original issue.

burtonrhodes:
Sure.... what about `submitUncheckedValues`?

Also, setting `fieldValue `won't fix the issue because even if `fieldValue `is set to an integer in the tag, the `CheckboxInterceptor `automatically resets the value to false when there are no checked values.  

CheckboxInterceptor.java:78
```
if (value.isMultiple()) {
LOG.debug("Bypassing automatic checkbox detection due to multiple checkboxes of the same name: {}", name);
continue;
}

// is this checkbox checked/submitted?
if (!parameters.contains(checkboxName)) {
// if not, let's be sure to default the value to false
extraParams.put(checkboxName, new Parameter.Request(checkboxName, uncheckedValue));
}
```

aleksandr-m:
Yes, you're right. The `fieldValue` won't help. The `uncheckedValue` parameter can be used in the `CheckboxInterceptor`, but it is not very convenient.

`submitUncheckedValue` or even `submitUnchecked` sound good.

burtonrhodes:
I like your suggestion better `submitUnchecked `.  I have pushed the changes.

yasserzamani:
However, this still is not a real solution for that weakness which I described at [1]. I like to resolve this issue inside `CheckboxInterceptor` instead.

[1] [Re: Extending StrutsTypeConverter :: How to detect specific type of "Class toClass"?](https://www.mail-archive.com/user@struts.apache.org/msg107462.html)

> COPY OF ABOVE LINK'S CONTENT:

Hello Burton,

Thank you; I examined your example and found out this behavior is 
because of something in Struts named "automatic checkbox detection", 
CheckboxInterceptor.

When your `choices` count is bigger than 1, you'll see your desired 
behavior because CheckboxInterceptor does nothing with following message 
where value.isMultiple() is true:

```java
if(value.isMultiple()) {
     LOG.debug("Bypassing automatic checkbox detection due to multiple 
checkboxes of the same name: {}", name);
} else if(!parameters.contains(checkboxName)) {
     extraParams.put(checkboxName, new Request(checkboxName, 
this.uncheckedValue));
}
```

But what happens when your `choices` count is one and you don't click on 
that one (your current example): Actually W3C says unchecked checkbox 
wont be posted, so, the above else will be executed (because 
!parameters.contains(checkboxName) is true) and CheckboxInterceptor sets 
a new parameter, "itemsFail->false". Then OGNL cannot convert `false` to 
an array of Integer into your action, so, the result will be changed to 
`input`.

burtonrhodes:
I don’t disagree. Unfortunately, i will have to do more learning on how the interceptor interacts with the framework before suggesting a change. Unless you see an easy fix inside interceptor?

yasserzamani:
Thanks! I just think we should not force user to fix Strut's weakness via providing another extra option. This is Strut's failure that cannot distinguish well if one checkbox is one or one of a list. It's not user's business to help :)

I re-thought and saw it's not `CheckboxInterceptor`'s business also. It does not have any way to know this. I think we should fix this in our converters, ognl or etc (I should see). i.e. when Struts sees one pair of `(param1, false)` in http parameters, if found `setParam1(boolean)` then invokes it, but if not found, then don't be disappointed Struts :) search for `setParam1(list or array of boolean)` and invoke it with just one element.

yasserzamani:
I see this issue could being avoided via overriding `uncheckedValue` of `checkbox` interceptor as below also:

```xml
<action name="ListConverterTest" class="com.afs.ListConverterTestAction">
    <interceptor-ref name="defaultStack">
        <param name="checkbox.uncheckedValue">-1</param>
    </interceptor-ref>
```

I know you like a `null` value instead of -1 but as far as I see for now, Struts ignores empty params like below, but I want to see if I can add this to Struts to accept also null and empty values like below (@apache/struts-committers ❓ ):

```xml
<action name="ListConverterTest" class="com.afs.ListConverterTestAction">
    <interceptor-ref name="defaultStack">
        <param name="checkbox.uncheckedValue"></param>
    </interceptor-ref>
```

cnenning:
> but I want to see if I can add this to Struts to accept also null and empty values like below

Yes, would be good if that would be possible. As far as I remember there was an issue that browsers don't send parameters if they don't have a value. I would be happy to review a PR 😉.

yasserzamani:
>  As far as I remember there was an issue that browsers don't send parameters if they don't have a value

No I did not mean accepting null or missing values from client browser. I meant accepting null and empty values (strings) from Struts developer in xmls e.g. `<param name="checkbox.uncheckedValue"></param>` should set `uncheckedValue` of `checkbox` interceptor to null :) Currently Struts simply ignores these.

cnenning:
>  I meant accepting null and empty values (strings) from Struts developer in xmls

Yes, and I thought the reason to that is somewhat related to browsers not sending values, in case of checkboxes. But I might be wrong here.

yasserzamani:
Aha you meant this issue. Yes you're right :) This issue is because of that; `checkbox` interceptor tries to fix this but cannot distinguish well if one checkbox is one checkbox or one of a list of checkboxes. I like to enable user to setting `uncheckedValue` of checkbox interceptor to null when a specific action needs such behavior.

yasserzamani:
IMO, I think it and an additional parameter is not required, as it seems, the corresponding issue already can be resolved in a clean way as I described at [this comment](https://github.com/apache/struts/pull/169#issuecomment-352999309).

aleksandr-m:
@yasserzamani Can you tell why do you think additional parameter will not fix this issue?

> I know you like a null value instead of -1 but as far as I see for now, Struts ignores empty params like below, but I want to see if I can add this to Struts to accept also null and empty values like below 

How do you set `null` value then?

Also in some cases it can be more convenient to set it per tag then per action.

burtonrhodes:
The interceptor solution is interesting, but IMHO feel that it is a departure from the idea that the Struts Tags should "just work".  Only a programmer intimately familiar with Struts (and "checkbox.uncheckedValue" parameter) would know to modify the interceptor in this manner.  And most likely this knowledge would only be ascertained after a situation like I found myself in - a lengthy debugging session to figure out why Struts wasn't converting the checkbox values correctly.  

My vote would be to 1) add a parameter (as these are well documented) or 2) fix the inner workings of Struts some other way so it "just works".  #2 is much preferred, but I'm afraid that is beyond my understanding of the framework to fix.

aleksandr-m:
It is 2.6. Maybe we just need to flip the switch and disable submitting unchecked value by default. Wdyt?

yasserzamani:
@aleksandr-m , an additional parameter will fix this issue but I thought, with this parameter, we actually ask user to fix a very specific Struts internal issue; i.e. I feel we can even name this parameter, `fixWW4865`. Because AFAIR, the issue occurs in a very specific situation: checkboxlist has only one element and it's value has mapped to a non boolean. In any other cases, checkboxinterceptor doesn't fail to set unchecked value.

However, @burtonrhodes is right that the interceptor solution is somehow same. In another side of coin, disabling this feature may have a lot of breaks or silent defects on user's apps. If we're OK with such breaks or silent defects, then I personally prefer removing this feature and behaving same as W3C has said, else we can merge this.

aleksandr-m:
> the issue occurs in a very specific situation: checkboxlist has only one element and it's value has mapped to a non boolean.

Most of checkboxlist-s in my apps are **not** mapped to booleans. ;)

> behaving same as W3C has said

Yep, would be great. As for the backward compatibility we can leave `submitUnchecked` parameter which defaults to `false` and can be set to `true` to restore old behavior.

lukaszlenart:
We can break things in 2.6 but I prefer to have an option to switch it back as this allows easier migration in some cases.

burtonrhodes:
As you can see... I went ahead and updated the pull request with the default value changed to "false" (and added tests for both values).  Let me know what you think.

aleksandr-m:
@burtonrhodes Thanks. Very good first PR. :+1: Hope to see more contribution from you.

burtonrhodes:
@aleksandr-m many thanks!  hope to contribute more as well.

PR #178, opened 2017-10-25
WW-4884: Allow define just a TextProvider

lukaszlenart:
Implements [WW-4884](https://issues.apache.org/jira/browse/WW-4884)


lukaszlenart:
@apache/struts-committers can you take a look?

cnenning:
Looks good to me. Straightforward and simple. 👍

PR #179, opened 2017-10-28
WW-4874 Introduces Async plugin (adds support for async methods)

yasserzamani:
> Tested with high concurrent requests using JMeter (a simple async action which returns a json result after 40 seconds).
# Simple usage
## Action
```java
public class AsyncAction {
    public Callable<String> execute() throws InterruptedException {
        return new Callable<String>() {
            @Override
            public String call() throws Exception {
                waitForEvent();
                return "success";
            }
        });
    }

    private void waitForEvent() throws InterruptedException {
        Thread.sleep(40000);
    }
}
```
## struts.xml
```xml
<action name="async" class="me.zamani.yasser.ww_convention.actions.AsyncAction">
	<result name="success" type="json" />
</action>
```
## web.xml
```xml
    <servlet>
        <servlet-name>appServlet</servlet-name>
        <servlet-class>org.apache.struts2.dispatcher.servlet.StrutsServlet</servlet-class>
        <load-on-startup>1</load-on-startup>
        <async-supported>true</async-supported>
    </servlet>

    <servlet-mapping>
        <servlet-name>appServlet</servlet-name>
        <url-pattern>/</url-pattern>
    </servlet-mapping>
```
# Customize timeout
## Action
```java
public class AsyncAction {
    public Callable<String> execute() throws InterruptedException {
        return new org.apache.struts2.servlet3.async.AsyncAction(60000/*timeout*/,
 new Callable<String>() {
            @Override
            public String call() throws Exception {
                waitForEvent();
                return "success";
            }
        });
    }

    private void waitForEvent() throws InterruptedException {
        Thread.sleep(40000);
    }
}
```
# Serializing multiple async tasks ❗️ 😮 
## Action
```java
public class AsyncAction {
    public Callable<String> execute() throws InterruptedException {
        return new Callable<String>() {
            @Override
            public String call() throws Exception {
                waitForEvent1();
        return new Callable<String>() {
            @Override
            public String call() throws Exception {
                waitForEvent2();
                return "success";
            }
        });
            }
        });
    }

    private void waitForEvent1() throws InterruptedException {
        Thread.sleep(40000);
    }

    private void waitForEvent2() throws InterruptedException {
        Thread.sleep(40000);
    }
}
```

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
@@ -561,8 +561,16 @@ public class Dispatcher {
             String name = mapping.getName();
             String method = mapping.getMethod();
 
-            ActionProxy proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(
-                    namespace, name, method, extraContext, true, false);
+            ActionProxy proxy;
+
+            //check if we are probably in an async resuming
+            ActionInvocation inv = ActionContext.getContext().getActionInvocation();
+            if (inv == null || inv.isExecuted()) {
+                proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(namespace, name, method,
+                        extraContext, true, false);
+            } else {
+                proxy = inv.getProxy();
+            }
 
             request.setAttribute(ServletActionContext.STRUTS_VALUESTACK_KEY, proxy.getInvocation().getStack());
 
--- a/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
@@ -564,12 +564,12 @@ public class Dispatcher {
             ActionProxy proxy;
 
             //check if we are probably in an async resuming
-            ActionInvocation inv = ActionContext.getContext().getActionInvocation();
-            if (inv == null || inv.isExecuted()) {
+            ActionInvocation invocation = ActionContext.getContext().getActionInvocation();
+            if (invocation == null || invocation.isExecuted()) {
                 proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(namespace, name, method,
                         extraContext, true, false);
             } else {
-                proxy = inv.getProxy();
+                proxy = invocation.getProxy();
             }
 
             request.setAttribute(ServletActionContext.STRUTS_VALUESTACK_KEY, proxy.getInvocation().getStack());
--- a/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
+++ b/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
@@ -106,7 +106,7 @@ public class DefaultAsyncManager implements AsyncManager, AsyncListener {
 
         String log = "Async result [" + asyncActionResult + "] of " + asyncContext;
         if (asyncCompleted) {
-            LOG.error(log + " - could not complete result executing due to timeout or network error");
+            LOG.debug(log + " - could not complete result executing due to timeout or network error");
         } else {
             LOG.debug(log + " - dispatching request to execute result in container");
             asyncContext.dispatch();
--- a/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
+++ b/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
@@ -132,7 +132,7 @@ public class DefaultAsyncManager implements AsyncManager, AsyncListener {
     @Override
     public void onTimeout(AsyncEvent asyncEvent) throws IOException {
         LOG.debug("Processing timeout for " + asyncEvent.getAsyncContext());
-        setAsyncActionResultAndDispatch("timeout");
+        setAsyncActionResultAndDispatch(AsyncAction.TIMEOUT);
     }
 
     @Override
--- a/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
+++ b/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
@@ -35,7 +35,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 /**
  * Implements {@link AsyncManager} to add support for invoke async actions via Servlet 3's API.
  *
- * @since 2.5.14
+ * @since 2.6
  */
 public class DefaultAsyncManager implements AsyncManager, AsyncListener {
     private static final Logger LOG = LogManager.getLogger(DefaultAsyncManager.class);
--- a/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
+++ b/plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.struts2.servlet3.async;
+package org.apache.struts2.async;
 
 import com.opensymphony.xwork2.AsyncManager;
 import org.apache.logging.log4j.LogManager;
@@ -37,8 +37,8 @@ import java.util.concurrent.atomic.AtomicInteger;
  *
  * @since 2.5.14
  */
-public class Servlet3AsyncManager implements AsyncManager, AsyncListener {
-    private static final Logger LOG = LogManager.getLogger(Servlet3AsyncManager.class);
+public class DefaultAsyncManager implements AsyncManager, AsyncListener {
+    private static final Logger LOG = LogManager.getLogger(DefaultAsyncManager.class);
     private static final AtomicInteger threadCount = new AtomicInteger(0);
 
     private AsyncContext asyncContext;
--- a/plugins/servlet3/src/main/java/org/apache/struts2/servlet3/async/Servlet3AsyncManager.java
+++ b/plugins/servlet3/src/main/java/org/apache/struts2/servlet3/async/Servlet3AsyncManager.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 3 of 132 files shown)


Tal-Daniel:
Is that in relation to the REST plugin? I don't understand why we need this one. Doesn't the server support concurrent requests already?

yasserzamani:
@Tal-Daniel ,

> Is that in relation to the REST plugin?

No, but could be used side-by-side with REST plugin as does not override `StrutsActionProxyFactory` or `DefaultActionInvocation`.

> I don't understand why we need this one. Doesn't the server support concurrent requests already?

Firstly this was my question also :) However, server supports concurrent requests but does not process any one of them asynchronously (processes a request in one specific thread). The main idea with Servlet3's async API is to process action in another not-container thread when action takes too long to being completed. This keeps container's (e.g. tomcat) thread free to be able to accept and process new incoming requests. i.e. users do not get `Connection Refused`.

cnenning:
> I don't understand why we need this one. Doesn't the server support concurrent requests already?

There have been discussions on dev mailing list in last few month. It is intended as thin wrapper around servlet api 3 async servlets.


On a first quick look this seems alright :+1:

aleksandr-m:
@yasserzamani Can you explain how this is different to `ExecuteAndWaitInterceptor`? What benefit does it bring, if any?

yasserzamani:
With a quick look on `ExecuteAndWaitInterceptor` I found these, @aleksandr-m : 

* In same session, looks like it cannot execute same action more than once.
* User has to repeat and repeat the request with same data until catch the result. Struts does everything from scratch and calls all interceptors and preResultListeners for each one of them ❗️ while with this PR, user like always only sends one request, waits then got only one result. Struts wont do everything from scratch but really resumes the current execution.
* In above mentioned repeated requests, Struts loses data and creates new ones from scratch and interceptors re-generates them from scratch ❗️ I thought about if I have `ActionContext.put("saveMe", "saved");sleep(10000)` in my action and then want to use it in my result. I'm not sure what happens when I executes my action with `ExecAndWaitInterceptor` as it creates new context in each request; while with this PR, all objects are that objects which are created and populated firstly before action execution.
* and at bottom line, I think this PR does things with fewer line of codes, more cleaner and robust that before, powered by Servlet 3's standard APIs.

Tal-Daniel:
Thanks @yasserzamani. Understood. Sounds good. Thanks for adding.

yasserzamani:
Very glad to hear your thanks and you like it :) For my part, I should thank Łukasz that mentioned this point.

lukaszlenart:
I wonder if this plugin should be named: Struts 2 Async plugin to not suggest usage of a given technology. Maybe in the future we can have other handlers implemented in the plugin.

lukaszlenart:
Also using `org.apache.struts2.servlet3` looks a bit strange, what about using `org.apache.struts2.async`?

rgielen:
+1 - not again - please don't use "Servlet 3" in any name. "Async" is the asset, so let's name it like that

yasserzamani:
Thanks for your comments! I named such this because I planned to package all Servlet 3's things here to facilitate it's deprecation and removeation when we upgraded to Servlet 3 in core. If you still like a change, I simply can refactor as you mentioned :)

rgielen:
It's also valid for Servlet 4.0, isn't it? :) We should better stay with features than with version numbers, we had this problem already with adding the 2 to the product brand name (Struts 2) - will a 3.0 release called Struts 3 or Struts 2 3.x ?? Same goes for Servlet versioning. So I'm still +1 for a refactoring

yasserzamani:
Nice! I got it. Thanks a lot for your time! I'll refactor all "Servlet 3" names :)

yasserzamani:
Łukasz, I should thanks you again for your nice suggestion, you were right, no it looks better ❤️ 

I also found some time and added unit tests covering all of my changes :)

cnenning:
If struts 2.6 depends on servlet-api 3 this code could be moved to core. As far as I understand it has been implemented as plugin just to be able to use servlet-api 3 without changing dependencies of struts itself.

yasserzamani:
No, Struts 2.6 will be upgraded to use Servlet 2.5.

plugins/servlet3/src/main/java/org/apache/struts2/servlet3/async/Servlet3AsyncManager.java
```diff
@@ -0,0 +1,149 @@
...
+                    setAsyncActionResultAndDispatch(asyncAction.call());
+                } catch (Throwable e) {
+                    setAsyncActionResultAndDispatch(e);
+                }
+            }
+        };
+        if (executor != null) {
+            executor.execute(task);
+        } else {
+            final Thread thread = new Thread(new Runnable() {
```
cnenning: If apps enter this else branch things are likely slow. So we should provide defaults to avoid this. I cannot find where `executor` is created. Could we initialize it in `AsyncAction`?

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -561,8 +561,16 @@ public void serviceAction(HttpServletRequest request, HttpServletResponse respon
...
             String method = mapping.getMethod();
 
-            ActionProxy proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(
-                    namespace, name, method, extraContext, true, false);
+            ActionProxy proxy;
+
+            //check if we are probably in an async resuming
+            ActionInvocation inv = ActionContext.getContext().getActionInvocation();
+            if (inv == null || inv.isExecuted()) {
+                proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(namespace, name, method,
```
cnenning: Name `inv` is a little short for my taste

yasserzamani: Thank you so much for your time and attention. I'll fix both on my pleasure :)

plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
```diff
@@ -0,0 +1,149 @@
...
+                    setAsyncActionResultAndDispatch(asyncAction.call());
+                } catch (Throwable e) {
+                    setAsyncActionResultAndDispatch(e);
+                }
+            }
+        };
+        if (executor != null) {
+            executor.execute(task);
+        } else {
+            final Thread thread = new Thread(new Runnable() {
```
cnenning: Maybe we should just throw an `Exception` here to force App developers to take care of their `Executor`. WDYT?

cnenning:
I wonder if we could get issues as Interceptors are run twice.

First time their "pre action" code is run before action executes and everything should behave as expected. But then their "post result" code is run before action (or result) are executed. At this point in time there is no result code so hopefully all interceptors behave well.

After action executed and request is re-dispatched to run result in container-thread interceptors run again. This time their "pre action" code runs but as action is not executed anymore it is quite useless. "post result" code should do what is expected this time.

Am I getting things wrong?

plugins/async/src/main/java/org/apache/struts2/async/DefaultAsyncManager.java
yasserzamani: No, providing an executor is optional. This if statement is not a fall back on user mistake but means an executor is optional :)

cnenning: And I wonder if it would be better to have it not optional

yasserzamani: I saw such in spring framework during my elementary studies and I think they have made a good decision to add this option. I think this is an additional option not a default needed one.

cnenning: Then we come back to my first comment: it is important to provide a sane default to avoid this else branch

yasserzamani: No, I meant default executor is a thread created and started by Struts. If user did not like, then can provide an executor as wish :)

cnenning: > I meant default executor is a thread created and started by Struts

I want to avoid this. The whole async topic is about performance. Createing a single thread is expensive and slow. That's why thread pools and executors exist. Having slow default behavior in a performance critical situation is a bad thing.

yasserzamani: > If apps enter this else branch things are likely slow. So we should provide defaults to avoid this. I cannot find where executor is created. Could we initialize it in AsyncAction?

I'm sorry, I , now, accidentally saw above in my mails. I don't know why it's not present here!

Anyway, now, I understand your worry. Please don't worry as I checked other frameworks like Spring and saw they also have such behavior as default and works. Please see [WebAsyncManager.java:177](https://github.com/spring-projects/spring-framework/blob/6c3a64578c10dd4e6fed933864f1721cf35203ae/spring-web/src/main/java/org/springframework/web/context/request/async/WebAsyncManager.java#L77) and then [SimpleAsyncTaskExecutor.java:236](https://github.com/spring-projects/spring-framework/blob/b6cae219d824f4990d77cac61477316ec678a09b/spring-core/src/main/java/org/springframework/core/task/SimpleAsyncTaskExecutor.java#L236).

For now, the importance is to provide scaleability by just simply running action outside of the container. For performance, if I add thread pool, then I will have to enable user to configure it (number of threads, scheduling mechanism, etc). somehow a lot of work :( I think, as an introduction, we can keep it simple as is. When needed, user can create and configure an executor finely as wish. If really they complain, then we have to add these. let what happens :)

yasserzamani:
Thanks a lot for this care.

I think interceptors wont run twice. When request is re-dispatched, at dispatcher:

```java
            //check if we are probably in an async resuming
            ActionInvocation invocation = ActionContext.getContext().getActionInvocation();
            if (invocation == null || invocation.isExecuted()) {
                proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(namespace, name, method,
                        extraContext, true, false);
            } else {
                proxy = invocation.getProxy();
            }
```

else branch will be executed then `proxy.execute();` will `return invocation.invoke();` where at DefaultActionInvocation.invoke:

```java
            if (asyncManager == null || !asyncManager.hasAsyncActionResult()) {
```

will be evaluated to `false` because async manager has an async result, so skips interceptors runs and goes to only executing pre result listeners and result :)

yasserzamani:
@cnenning , I added a showcase, a minimal simple chat room which uses `server push` to retrieve new messages instead of polling server. Thanks for your notice 👍

lukaszlenart:
I think we can merge this, an astonish work done by Yasser :)

PR #181, opened 2017-11-08
WW-4888 add escaping possibilities to text-tag

sdutry:
Added the option of escaping the result of an `<s:text>` tag.

Defaults set to not escaping in order not to break current behavior.

possible improvement:
* now holds duplicate prepare method from the `Property` class

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/Text.java
+++ b/core/src/main/java/org/apache/struts2/components/Text.java
@@ -19,6 +19,7 @@
 package org.apache.struts2.components;
 
 import com.opensymphony.xwork2.util.ValueStack;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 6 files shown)


yasserzamani:
Maybe we can ask user to use `PropertyTag` instead for such cases.

Or if we would like to add this also to `TextTag` , I would refactor this into another super tag and then inherit it in both `PropertyTag` and `TextTag` :)

lukaszlenart:
`PropertyTag` shouldn't be used to display messages and I have no issue with having these properties duplicated

yasserzamani:
So looks good; I could not fuss about any thing at first ;)

lukaszlenart:
ok, LGTM 👍

aleksandr-m:
And we can easily change `false` to `true` in `2.6`.

yasserzamani:
As we removed snippets in docs, Parameters section of [/source/tag-developers/text-tag.md](https://github.com/apache/struts-site/tree/master/source/tag-developers/text-tag.md) should being updated also.

sdutry:
> As we removed snippets in docs, Parameters section of /source/tag-developers/text-tag.md should being updated also.

@yasserzamani 
I see you updated the docs.
Personaly i would have waited till the official release of the new struts version.
Changing it automaticaly updates the struts site, which now references a non-officialy released struts version.

Personaly I don't think its a problem for this specific case, it's just something to take into account when updating docs for upcoming versions..

Maybe a possible solution to prevent this in the future could be to have a seperate branch for the documentation updates for the new release that gets merged when it gets officialy released?

What do you think @lukaszlenart ?

yasserzamani:
> Parameters section of /source/tag-developers/text-tag.md should being updated also.

Done! To test my first direct commit as a Committer 😉

yasserzamani:
Oh I'm sorry :( I thought you did not find a time and I just want to test if I can commit as a Committer and found this is a good case.

I was sure because about one month ago, I already have an accepted merge at [WW-4034 Adds docs for how customizing JSONWriter](https://github.com/apache/struts-site/pull/29).

IMHO, maybe it is not bad users know what comings up 😉  I think new branches may add complexities and release needed works.

lukaszlenart:
@sdutry an another branch is a nice idea, I think this would work.

Anyway, as we do not provide a separated copy of documentation per each release (as it used to be), we have to be sure that we're adding a `since` tag to each new option.

PR #184, opened 2017-11-11
WW-4713 Added deprecation warnings as preparation to drop searchValue…

sdutry:
…Stack attribute from the <s:text/> tag

Added deprecation warnings in order to remove the attribute starting struts version 2.6


lukaszlenart:
Looks good 👍

PR #168, opened 2017-09-26
WW-4849: Initializable interface

lukaszlenart:
Fixes [WW-4849](https://issues.apache.org/jira/browse/WW-4849)

PR #185, opened 2017-11-12
WW-4876 WW-4841 Parses validators after all injections done

yasserzamani:
Also fixes already existed silent errors of `SpringObjectFactoryTest` which have born after #168 :)


lukaszlenart:
Looks good to me 👍

PR #186, opened 2017-11-13
changed Map iterations to use entrySet when both key and value are used

sdutry:
## Changes
Changed iterations over the keySet of Maps when the values were (also) used.
### Reported by
Sonar build 
* see: [Iterate over the "entrySet" instead of the "keySet"](https://builds.apache.org/analysis/issues/search#resolved=false|rules=squid%3AS2864|projectUuids=AVkMC6uolrNkgx9HJrOf)

## Encountered possible problems
I did not change 1 of the warnings, because i think the code could throw a  `ConcurrentModificationException`
* see: [CheckboxInterceptor.java#L63-L80](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/interceptor/CheckboxInterceptor.java#L63-L80)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
@@ -154,15 +154,15 @@ public abstract class AbstractMatcher<E> implements Serializable {
         Map<String, String> map = new LinkedHashMap<>();
         
         //this will set the group index references, like {1}
-        for (String key : orig.keySet()) {
-            map.put(key, convertParam(orig.get(key), vars));
+        for (Map.Entry<String,String> entry : orig.entrySet()) {
+            map.put(entry.getKey(), convertParam(entry.getValue(), vars));
         }
         
         //the values map will contain entries like name->"Lex Luthor" and 1->"Lex Luthor"
         //now add the non-numeric values
-        for (String key: vars.keySet()) {
-            if (!NumberUtils.isNumber(key)) {
-                map.put(key, vars.get(key));
+        for (Map.Entry<String,String> entry: vars.entrySet()) {
+            if (!NumberUtils.isNumber(entry.getKey())) {
+                map.put(entry.getKey(), entry.getValue());
             }
         }
         
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/ActionConfigMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/ActionConfigMatcher.java
@@ -59,8 +59,8 @@ public class ActionConfigMatcher extends AbstractMatcher<ActionConfig> implement
             Map<String, ActionConfig> configs,
             boolean looseMatch) {
         super(patternMatcher);
-        for (String name : configs.keySet()) {
-            addPattern(name, configs.get(name), looseMatch);
+        for (Map.Entry<String, ActionConfig> entry : configs.entrySet()) {
+            addPattern(entry.getKey(), entry.getValue(), looseMatch);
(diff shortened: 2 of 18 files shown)


sdutry:
> I did not change 1 of the warnings, because i think the code could throw a ConcurrentModificationException
> 
> see: CheckboxInterceptor.java#L63-L80

I was wrong here,
The HttpParameters class overrides the getKey() method and returns an unmodifiable copy of the underlying keySet
see: [HttpParameters.java#L139-L142](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java#L139-L142)

Given the entrySet() method doesn't return a copy, this can't be changed as simply as the others.

lukaszlenart:
But it returns https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java#L150-L152 ?

sdutry:
> But it returns https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java#L150-L152 ?

While it does, i don't know if this is sufficient. The keySet() method returns a unmodifiable view of a newly created set where as the entrySet() returns a unmodifiable view of the actual entrySet().

I'll test later today if the tests still work if i change it.

yasserzamani:
> The keySet() method returns a unmodifiable view of a newly created set where as the entrySet() returns a unmodifiable view of the actual entrySet().

I tested and you were right :) when I remove `new TreeSet<` there, tests fails!

As we finally should fix it, I tested something as below and seems work.

```diff
Index: core/src/main/java/org/apache/struts2/interceptor/CheckboxInterceptor.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- core/src/main/java/org/apache/struts2/interceptor/CheckboxInterceptor.java	(revision 47a4a57d2d7fc4d5ab1765f6bd31bf1b0f0d583c)
+++ core/src/main/java/org/apache/struts2/interceptor/CheckboxInterceptor.java	(revision )
@@ -26,6 +26,7 @@
 import org.apache.struts2.dispatcher.HttpParameters;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 /**
@@ -60,12 +61,14 @@
         HttpParameters parameters = ai.getInvocationContext().getParameters();
         Map<String, Parameter> extraParams = new HashMap<>();
 
-        for (String name : parameters.keySet()) {
+        HashSet<String> checkboxes = new HashSet<String>();
+        for (Map.Entry<String, Parameter> param : parameters.entrySet()) {
+            String name = param.getKey();
             if (name.startsWith("__checkbox_")) {
                 String checkboxName = name.substring("__checkbox_".length());
 
-                Parameter value = parameters.get(name);
-                parameters = parameters.remove(name);
+                Parameter value = param.getValue();
+                checkboxes.add(name);
                 if (value.isMultiple()) {
               	    LOG.debug("Bypassing automatic checkbox detection due to multiple checkboxes of the same name: {}", name);
                     continue;
@@ -78,6 +81,7 @@
                 }
             }
         }
+        parameters.remove(checkboxes);
 
 
         ai.getInvocationContext().getParameters().appendAll(extraParams);
```

sdutry:
@yasserzamani 
added the change.

yasserzamani:
Great! nice work! I reviewed and seems it's ready to go :)

lukaszlenart:
We can go ahead if you want, if there be any release blocker in 2.5.14 I will create a branch off the STRUTS_2_5_14 tag to fix that and I will prepare new packages

PR #187, opened 2017-11-17
WW-4892 WW-4893: Use Jackson to handle JSON request

lukaszlenart:
Implements [WW-4892](https://issues.apache.org/jira/browse/WW-4892)
Closes [WW-4893](https://issues.apache.org/jira/browse/WW-4893)


cnenning:
Oh, I see there is already an abstraction to switch json libraries. Using Jackson as default is probably what most people do anyway 😄

lukaszlenart:
This is need to fix some minor issues

PR #188, opened 2017-11-29
WW-4900 WW-4873 Fixes DefaultActionInvocation & BackgroundProcess serialization

yasserzamani:
Fixes WW-4900 and WW-4873 .

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java
+++ b/core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java
@@ -503,9 +503,9 @@ public class DefaultActionInvocation implements ActionInvocation {
     }
 
     /**
-     * Version ready to be serialize
+     * Version ready to be serialize via removing all unserializable but restorable references
      *
-     * @return instance without reference to {@link Container}
+     * @return instance without unserializable references
      */
     public ActionInvocation serialize() {
         DefaultActionInvocation that = this;
@@ -531,7 +531,7 @@ public class DefaultActionInvocation implements ActionInvocation {
     }
 
     /**
-     * Restoring Container
+     * Restoring references
      *
      * @param actionContext current {@link ActionContext}
      * @return instance which can be used to invoke action
--- a/core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java
+++ b/core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java
@@ -35,8 +35,9 @@ import ognl.MethodFailedException;
 import ognl.NoSuchPropertyException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.struts2.ServletActionContext;
 
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
 import java.util.ArrayList;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java
```diff
@@ -508,6 +509,13 @@ protected String saveResult(ActionConfig actionConfig, Object methodResult) {
     public ActionInvocation serialize() {
         DefaultActionInvocation that = this;
         that.container = null;
+
+        if(that.getInvocationContext() != null && that.getInvocationContext().getContextMap() != null) {
+            Map<String, Object> thatContextMap = that.getInvocationContext().getContextMap();
+            thatContextMap.remove(ServletActionContext.HTTP_REQUEST);
+            thatContextMap.remove(ServletActionContext.HTTP_RESPONSE);
```
lukaszlenart: I'm not 100% sure but in such way you will break Token Session support as I far remember it will try to restore previous request on invalid token.

yasserzamani: No, as I see, currently, it restores previous `invocation` and replaces previous `request` and `response` with current ones. It does not use them.

```java
            ActionInvocation savedInvocation = InvocationSessionStore.loadInvocation(sessionTokenName, token);

            if (savedInvocation != null) {
                // set the valuestack to the request scope
                ValueStack stack = savedInvocation.getStack();
                request.setAttribute(ServletActionContext.STRUTS_VALUESTACK_KEY, stack);//current request

                ActionContext savedContext = savedInvocation.getInvocationContext();
                savedContext.getContextMap().put(ServletActionContext.HTTP_REQUEST, request);//replace prev req with current
                savedContext.getContextMap().put(ServletActionContext.HTTP_RESPONSE, response);//replace prev res with current
                Result result = savedInvocation.getResult();

                if ((result != null) && (savedInvocation.getProxy().getExecuteResult())) {
                    result.execute(savedInvocation);
                }

                // turn off execution of this invocations result
                invocation.getProxy().setExecuteResult(false);

                return savedInvocation.getResultCode();//no usage of prev req

```

yasserzamani: `InvocationSessionStore` also does not use previous `request` and use session to restore previous `invocation`.

yasserzamani:
Previous commit fixes these:

* I wrongly used/added `org.apache.struts2` package in `com.opensymphony.xwork2` in first commit. Previous commit deletes this efferent coupling.
* I missed `deserialize` method in first commit. Previous commit covers it.

yasserzamani:
I'm thinking about these about two weeks and finally I don't like this PR and am going to close this specially when I also saw [user additional objections](https://issues.apache.org/jira/browse/WW-4900?focusedCommentId=16287770&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16287770).

In general I don't like disturbing user's session by complex objects like action invocation etc. I am going to work if I can store these somewhere else but likewise the session.

tiztm:
This Future is also important in Struts2 2.3.x, for "InvocationSessionStore.java".
Can it be appled in 2.3.x branch?

PR #189, opened 2017-12-04
WW-4899: upgrade Spring

lukaszlenart:
Upgrades Spring to version 4.3.13

Refers [WW-4899](https://issues.apache.org/jira/browse/WW-4899)

### DIFF
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
-            <version>${struts2.springPlatformVersion}</version>
+            <version>${spring.platformVersion}</version>
             <scope>test</scope>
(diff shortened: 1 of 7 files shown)


yasserzamani:
Thank you! LGTM 👍 What do you feel to deliver this with 2.5.15 too?

lukaszlenart:
This can be a major change, switching from Spring 4.1 to 4.3 can break users' apps. I would wait till 2.6.

aleksandr-m:
It should be rather safe. Spring dependencies are optional in core and for those who use spring in their projects they've probably overridden it already.

lukaszlenart:
Ok, let's merge it, who want to pull the trigger?

lukaszlenart:
Ok @yasserzamani & @aleksandr-m you have convinced me, please merge this puppy :)

PR #190, opened 2017-12-11
WW-4901 Decouples from URL.openConnection implementation of container

yasserzamani:


core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
```diff
@@ -0,0 +1,93 @@
...
+        return jarFile;
+    }
+
+    @Override
+    public void connect() throws IOException {
+        if (connected)
+            return;
+
+        try (final InputStream in = getJarFileURL().openConnection().getInputStream()) {
+            jarFile = AccessController.doPrivileged(
```
cnenning: could you add something like `if (!getJarFileURL().openConnection() instanceof JarURLConnection)` to avoid copiing of file if not necessary?

yasserzamani: Thank you very much for your attention 👍 fixed :)

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
```diff
@@ -0,0 +1,102 @@
...
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.util.jar.JarFile;
+
+/**
+ * WW-4901 Decouples from underlying implementation of {@link URL#openConnection()}
+ * e.g. from IBM WebSphere com.ibm.ws.classloader.Handler$ClassLoaderURLConnection
+ * @since 2.5.15
+ */
+class StrutsJarURLConnection extends JarURLConnection {
```
lukaszlenart: 👍

lukaszlenart:
Looks good to me, LGMT 👍

PR #191, opened 2017-12-13
WW-4900 Makes BackgroundProcess transient

yasserzamani:
Currently we won't support exec and wait from de-serialized session and maybe add this support some day on user demand. Why I think to drop such support? It's not a good practice to try serializing such large or variant unpredictable objects like action and invocation ([CWE-579: J2EE Bad Practices: Non-serializable Object Stored in Session](https://cwe.mitre.org/data/definitions/579.html)).

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
@@ -45,7 +45,7 @@ public class BackgroundProcess implements Serializable {
      * @param invocation The action invocation
      * @param threadPriority The thread priority
      */
-    BackgroundProcess(String threadName, final ActionInvocation invocation, int threadPriority) {
+    public BackgroundProcess(String threadName, final ActionInvocation invocation, int threadPriority) {
         this.invocation = invocation;
         try {
             final Thread t = new Thread(new Runnable() {
@@ -75,7 +75,7 @@ public class BackgroundProcess implements Serializable {
      *
      * @throws Exception any exception thrown will be thrown, in turn, by the ExecuteAndWaitInterceptor
      */
-    private void beforeInvocation() throws Exception {
+    protected void beforeInvocation() throws Exception {
         ActionContext.setContext(invocation.getInvocationContext());
     }
 
@@ -86,7 +86,7 @@ public class BackgroundProcess implements Serializable {
      *
      * @throws Exception any exception thrown will be thrown, in turn, by the ExecuteAndWaitInterceptor
      */
-    private void afterInvocation() throws Exception {
+    protected void afterInvocation() throws Exception {
         ActionContext.setContext(null);
     }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
```diff
@@ -44,9 +45,8 @@
      * @param invocation The action invocation
      * @param threadPriority The thread priority
      */
-    public BackgroundProcess(String threadName, final ActionInvocation invocation, int threadPriority) {
+    BackgroundProcess(String threadName, final ActionInvocation invocation, int threadPriority) {
```
lukaszlenart: In such case you are reducing re-usability of this class in other interceptors, user will have to copy this class if she wants to re-implement the interceptor

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
```diff
@@ -75,7 +75,7 @@ public void run() {
      *
      * @throws Exception any exception thrown will be thrown, in turn, by the ExecuteAndWaitInterceptor
      */
-    protected void beforeInvocation() throws Exception {
+    private void beforeInvocation() throws Exception {
```
lukaszlenart: Reducing scope can affect users and blocks adding other implementations.

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
```diff
@@ -86,7 +86,7 @@ protected void beforeInvocation() throws Exception {
      *
      * @throws Exception any exception thrown will be thrown, in turn, by the ExecuteAndWaitInterceptor
      */
-    protected void afterInvocation() throws Exception {
+    private void afterInvocation() throws Exception {
```
lukaszlenart: Same here, using `protected` allows users re-implementing this class easily

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
yasserzamani: Thanks 👍 I did not know user may want. fixed :)

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
yasserzamani: Thanks 👍 . fixed :)

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
yasserzamani: Thanks 👍 . fixed :)

core/src/main/java/org/apache/struts2/interceptor/BackgroundProcess.java
lukaszlenart: Yeah... this is a bit different than development of a normal app. This a framework that should be easily extendable by users and quite often we need to break good practices (like reducing scope) because users would like to use it. You wasn't the first one who did that ;-)

lukaszlenart:
`testSerializeDeserialize` of `BackgroundProcessTest` is failing ...

yasserzamani:
Was a good catch by junit 👍 I should have wait for background thread. fixed. thanks!

lukaszlenart:
Let's rock 👍

tiztm:
I think so , such huge object in session is too bad
It seems that,TokenSession needs some changes too.

PR #192, opened 2017-12-14
WW-4873 Makes ActionInvocation not serializable and InvocationSessionStore$InvocationContext transient

yasserzamani:
⚠️  For simpler review, please review each commit separately.

❓ Why `serialize` and `deserialize` methods of `ActionInvocation` deleted? `ActionInvocation` is a large complex object which contains request, response, container, value stack and action context. It's hard to maintain it serializable e.g. user may want to store a private not serializable object in action. And It's not a good practice ([CWE-579: J2EE Bad Practices: Non-serializable Object Stored in Session](https://cwe.mitre.org/data/definitions/579.html)).

ℹ️ I know I also deleted `that.container = actionContext.getContainer();` (replacing restored invocation container with current container) with `ActionInvocation.deserialize` method. I saw it's not needed and even it's better for restored default invocation to work with itself stored container to keep consistency. Anyway currently it's really not needed because only `TokenSessionStoreInterceptor` uses it and it only passes the invocation to result and result is not aware about `DefaultActionInvocation` (only knows `ActionInvocation`):

```java
if ((result != null) && (savedInvocation.getProxy().getExecuteResult())) {
    result.execute(savedInvocation);
}
```

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ActionInvocation.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ActionInvocation.java
@@ -21,8 +21,6 @@ package com.opensymphony.xwork2;
 import com.opensymphony.xwork2.interceptor.PreResultListener;
 import com.opensymphony.xwork2.util.ValueStack;
 
-import java.io.Serializable;
-
 /**
  * An {@link ActionInvocation} represents the execution state of an {@link Action}. It holds the Interceptors and the Action instance.
  * By repeated re-entrant execution of the <code>invoke()</code> method, initially by the {@link ActionProxy}, then by the Interceptors, the
@@ -31,7 +29,7 @@ import java.io.Serializable;
  * @author Jason Carreira
  * @see com.opensymphony.xwork2.ActionProxy
  */
-public interface ActionInvocation extends Serializable {
+public interface ActionInvocation {
 
     /**
      * Get the Action associated with this ActionInvocation.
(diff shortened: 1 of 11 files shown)


lukaszlenart:
wtf :\

lukaszlenart:
Looks lik a known issue https://github.com/lemurheavy/coveralls-public/issues/351

yasserzamani:
That is weird. [https://coveralls.io/github/apache/struts](https://coveralls.io/github/apache/struts) always claims it's a `FIRST BUILD ON MASTER`. To test `coveralls.io`, I simulated at [https://coveralls.io/github/yasserzamani/struts](https://coveralls.io/github/yasserzamani/struts) but it works there!

Another thing is it jumps from `#1` to `#606` at [https://coveralls.io/repos/188307/builds?page=2](https://coveralls.io/repos/188307/builds?page=2), I'm not sure if it's a problem but [https://coveralls.io/repos/52976/builds?page=64](https://coveralls.io/repos/52976/builds?page=64) is not like this.

lukaszlenart:
I think I have fixed it, I forgot that I had setup the build on Travis to be triggered only by PRs - now also changes to the `master` branch trigger a build :)

lukaszlenart:
Looks good to me, 👍

yasserzamani:
Works :) 👍  [example](https://github.com/apache/struts/pull/193#issuecomment-352194799)

PR #194, opened 2017-12-20
used char instead of a single character string

sdutry:
Fixes sonar reported issue about single character strings being used inside the struts2 core module.

Sonar reported issue:
[Sonar - char instead of string](https://builds.apache.org/analysis/issues/search#resolved=false|tags=performance|rules=squid%3AS3027|projectUuids=AVkMC6uolrNkgx9HJrOf|moduleUuids=AVkMC7FC5zvJuQXqiyMz)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/entities/AllowedMethods.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/entities/AllowedMethods.java
@@ -56,7 +56,7 @@ public class AllowedMethods {
             if (isPattern && !method.startsWith("regex:") && !strictMethodInvocation) {
                 allowedMethods.add(new PatternAllowedMethod(methodPattern.toString(), method));
             } else if (method.startsWith("regex:")) {
-                String pattern = method.substring(method.indexOf(":") + 1);
+                String pattern = method.substring(method.indexOf(':') + 1);
                 allowedMethods.add(new PatternAllowedMethod(pattern, method));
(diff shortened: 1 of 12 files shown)


lukaszlenart:
👍

rgielen:
From the peanut gallery:

If I'd stumble about this commit later, I would ask myself if the message describes the problem this commit fixes or if it describes what the contributor did before submitting it as a commit. That is: "used char instead of a single character string [before this commit, now fixed]" vs "used char instead of a single character string [and then committed this as an improvement]"

Over the last years I came to the conclusion that most Git commit message conventions make perfect sense, especially rule 5 in https://chris.beams.io/posts/git-commit/ - use imperative mood for your summary, such that it completes the sentence "If applied, this commit will [your subject line here]"

When applied, the commit message would read "use char instead of a single character string" (I have no strong opinion on whether to start with a capital letter). IMO this message would be unambiguous and clear. WDYT?

To be clear, this is just for future consideration, no intention to change that now merged commit  anyhow!

yasserzamani:
👍 I agree with you and will follow all of that article.

I could not find in that any thing about when several commits are related to same issue. Should we use `See also: WW-####` in body of each one? In which one's body we should use `Resolves: WW-####`? Or maybe we should not use them in anyone and instead, always use PR and it's merge commit as an aggregator and parent?

sdutry:
@rgielen 
I'll try and keep it in mind.

PR #195, opened 2017-12-28
WW-4903: fixes PrefixBasedActionProxyFactory

lukaszlenart:
Resolves issue [WW-4903](https://issues.apache.org/jira/browse/WW-4903)

### DIFF
--- a/core/src/main/java/org/apache/struts2/factory/PrefixBasedActionProxyFactory.java
+++ b/core/src/main/java/org/apache/struts2/factory/PrefixBasedActionProxyFactory.java
@@ -59,12 +59,11 @@ import java.util.Set;
  * </pre>
  * <!-- END SNIPPET: description -->
  */
-public class PrefixBasedActionProxyFactory extends StrutsActionProxyFactory implements Initializable {
+public class PrefixBasedActionProxyFactory extends StrutsActionProxyFactory {
 
     private static final Logger LOG = LogManager.getLogger(PrefixBasedActionProxyFactory.class);
 
     private Map<String, ActionProxyFactory> actionProxyFactories = new HashMap<>();
-    private Set<String> prefixes = new HashSet<>();
 
     @Inject
     public void setContainer(Container container) {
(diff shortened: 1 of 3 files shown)


lukaszlenart:
@apache/struts-committers this is ready for review

lukaszlenart:
... or not ... adding one more test :)

lukaszlenart:
now is ready :)

yasserzamani:
To review, during reproducing [WW-4903](https://issues.apache.org/jira/browse/WW-4903), I reached a few difficulties which I decided to ask if you know a quick answer please (to save time):

* `<constant name="struts.actionProxyFactory" value="prefix"/>` is not mentioned at `REST and non-RESTful URL’s Together Configuration` section of [docs](http://struts.apache.org/plugins/rest/#usage). I checked and it does not exist in first revision of our new docs also. It seems it did not exist from first. Do you think it is a forgotten item? If so, why users did not encounter this issue sooner? Or maybe no one has used such configuration (REST together non-REST) so far?

* Do you know why user gets 

```
com.opensymphony.xwork2.inject.ContainerImpl$MissingDependencyException: No mapping found for dependency [type=com.opensymphony.xwork2.ActionProxyFactory, name='struts.actionProxyFactory'] in public void org.apache.struts2.factory.PrefixBasedActionProxyFactory.setActionProxyFactory(com.opensymphony.xwork2.ActionProxyFactory).
```
while he has defined `<constant name="struts.actionProxyFactory" value="prefix"/>` already?!

lukaszlenart:
> re: docs

It's documented [here](http://struts.apache.org/core-developers/action-mapper.html#prefixbasedactionmapper) next to the `PrefixBasedActionMapper` mapper but maybe it's a good idea to mention it also on the REST plugin page - I will add that.

> re: exception

It's because of a circle dependency,  injecting itself into itself ;-) This was broken from the very beginning but nobody has discovered that which means nobody is used it till now :)

lukaszlenart:
Added a note https://github.com/apache/struts-site/pull/90

yasserzamani:
I reached two issues:

`PrefixBasedActionProxyFactoryTest` does not have license so `clean install` fails here :)

I could not make `PrefixBasedActionProxyFactory.init()` being called. It seems the framework does not call `init` because in `ContainerBuilder#95`:

```java
        if (Initializable.class.isAssignableFrom(key.getType())) {
            initializableFactories.add(createCallableFactory(key, scopedFactory));
        }
```

`key` is `com.opensymphony.xwork2.ActionProxyFactory` which does not implement `Initializable` (it's implementation i.e. `PrefixBasedActionProxyFactory` implements that interface).

I could not find a way to know that implementation (here e.g. `PrefixBasedActionProxyFactory`) in line#95 to add it also to if statement i.e. if implementation implements `Initializable`. Do you know any please? Another way is to make `com.opensymphony.xwork2.ActionProxyFactory` initializable but I think it's not a good idea and will make all super classes have to implement `init` method :(

lukaszlenart:
> re: license

Ach ... right but why CI didn't catch that 🤔  ...

> re: init

Oh damn... good catch, totally forgot about the limitations and for now there is no way to figure out the target type, I think we should extend `Key` and add `T` there. Reverted back to the previous mechanism as constants are injected last (noticed that when I was exploring `Container`).

yasserzamani:
>  why CI didn't catch that

It seems because CI uses `clean package` not `clean install`. But I tested your fix and now also fails on `clean package` :)

> I think we should extend Key and add T there.

I see that `key` already has defined as `Key<T>` there. Seems we cannot reach implementation class via key at compiletime. But instead I see the target implementation class is saved inside `factory` but only in runtime. In compiletime, it's just an interface.

lukaszlenart:
Yeah... factory is the key here ;-) I have a fix already and will open a new PR. When we close this PR we can start preparing another 2.5.x release and focus on 2.6 :)

yasserzamani:
LGTM 👍 let's go :)

PR #200, opened 2018-01-09
WW-4907 support JSR 303 Validation Groups in BeanValidation-Plugin

HedjuHor:
it would be nice Feature if BeanValidation-Plugin would support 'Grouping Constraints'

https://docs.jboss.org/hibernate/validator/5.1/reference/en-US/html/chapter-groups.html

Hedju Hor
 
Implements [WW-4907](https://issues.apache.org/jira/browse/WW-4907)

### DIFF
--- a/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
+++ b/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
@@ -16,24 +16,35 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.struts.beanvalidation.models;
+package org.apache.struts.beanvalidation.constraints;
 
-import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Size;
-import javax.validation.groups.Default;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
-public class Address {
+/**
+ * Validation with Group Constrains on execution method (Action)
+ * <p>
+ * Example, Validate with on special group
+ * </p>
+ * <pre>
+ *
+ *  {@literal @}ValidateGroup(CarChecks.class)
+ *  {@literal @}Action...
+ * </pre>
+ * <p>
+ * <p>
+ * Example, Validate with severals special group
+ * </p>
+ * <pre>
+ *  {@literal @}ValidateGroup(Default.class, CarChecks.class, DriverChecks.class)
+ *  {@literal @}Action...
+ * </pre>
+ */
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValidateGroup {
 
-    @NotNull(message = "streetNotNull", groups = {Default.class, Person.StreetChecks.class, Person.NameAndStreetChecks.class})
-    @Size(min = 3, max = 64, message = "streetSize", groups = {Default.class, Person.StreetChecks.class, Person.NameAndStreetChecks.class})
-    private String street;
-
-    public void setStreet(String street) {
-        this.street = street;
-    }
-
-    public String getStreet() {
-        return street;
-    }
-
+    Class<?>[] value() default {};
 }
--- a/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
+++ b/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
@@ -27,16 +27,18 @@ import com.opensymphony.xwork2.interceptor.MethodFilterInterceptor;
 import com.opensymphony.xwork2.validator.DelegatingValidatorContext;
 import com.opensymphony.xwork2.validator.ValidatorContext;
 import org.apache.commons.lang3.BooleanUtils;
-import org.apache.commons.lang3.reflect.MethodUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.reflect.MethodUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.struts.beanvalidation.constraints.ValidateGroup;
 import org.apache.struts.beanvalidation.validation.constant.ValidatorConstants;
 import org.apache.struts2.interceptor.validation.SkipValidation;
 
 import javax.validation.ConstraintViolation;
 import javax.validation.Validator;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Set;
 
 /**
@@ -95,36 +97,42 @@ public class BeanValidationInterceptor extends MethodFilterInterceptor {
         if (LOG.isDebugEnabled()) {
             LOG.debug("Validating [{}/{}] with method [{}]", invocation.getProxy().getNamespace(), invocation.getProxy().getActionName(), methodName);
         }
+        Class<?>[] validationGroup = getValidationGroups(action, methodName);
 
         if (null == MethodUtils.getAnnotation(getActionMethod(action.getClass(), methodName), SkipValidation.class,
                 true, true)) {
             // performing bean validation on action
-            performBeanValidation(action, validator);
+            performBeanValidation(action, validator, validationGroup);
         }
 
         return invocation.invoke();
     }
 
-    protected void performBeanValidation(Object action, Validator validator) {
+    protected Class<?>[] getValidationGroups(Object action, String methodName) throws NoSuchMethodException {
+        ValidateGroup validateGroup = MethodUtils.getAnnotation(getActionMethod(action.getClass(), methodName), ValidateGroup.class, true, true);
+        return validateGroup == null ? new Class[]{} : validateGroup.value();
+    }
 
-        LOG.trace("Initiating bean validation..");
+    protected void performBeanValidation(Object action, Validator validator, Class<?>[] groups) {
+
+        LOG.trace("Initiating bean validation.. with groups [{}]", Arrays.toString(groups));
 
         Set<ConstraintViolation<Object>> constraintViolations;
 
         if (action instanceof ModelDriven) {
             LOG.trace("Performing validation on model..");
             Object model = (Object)((ModelDriven<?>) action).getModel();
-            constraintViolations = validator.validate(model);
+            constraintViolations = validator.validate(model, groups);
         } else {
             LOG.trace("Performing validation on action..");
-            constraintViolations = validator.validate(action);
+            constraintViolations = validator.validate(action, groups);
         }
 
         addBeanValidationErrors(constraintViolations, action);
     }
 
     @SuppressWarnings("nls")
-    private void addBeanValidationErrors(Set<ConstraintViolation<Object>> constraintViolations, Object action) {
+    protected void addBeanValidationErrors(Set<ConstraintViolation<Object>> constraintViolations, Object action) {
         if (constraintViolations != null) {
             ValidatorContext validatorContext = new DelegatingValidatorContext(action, textProviderFactory);
             for (ConstraintViolation<Object> constraintViolation : constraintViolations) {
--- a/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
+++ b/plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
@@ -31,7 +31,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.reflect.MethodUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.struts.beanvalidation.constraints.ValidateGroup;
+import org.apache.struts.beanvalidation.constraints.ValidationGroup;
 import org.apache.struts.beanvalidation.validation.constant.ValidatorConstants;
 import org.apache.struts2.interceptor.validation.SkipValidation;
 
@@ -109,8 +109,8 @@ public class BeanValidationInterceptor extends MethodFilterInterceptor {
     }
 
     protected Class<?>[] getValidationGroups(Object action, String methodName) throws NoSuchMethodException {
-        ValidateGroup validateGroup = MethodUtils.getAnnotation(getActionMethod(action.getClass(), methodName), ValidateGroup.class, true, true);
-        return validateGroup == null ? new Class[]{} : validateGroup.value();
+        ValidationGroup validationGroup = MethodUtils.getAnnotation(getActionMethod(action.getClass(), methodName), ValidationGroup.class, true, true);
+        return validationGroup == null ? new Class[]{} : validationGroup.value();
     }
 
     protected void performBeanValidation(Object action, Validator validator, Class<?>[] groups) {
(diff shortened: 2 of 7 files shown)


lukaszlenart:
Nice but please register an issue [here](https://issues.apache.org/jira/projects/WW/) - this allows control us what's changed and when

plugins/bean-validation/src/test/java/org/apache/struts/beanvalidation/models/Address.java
```diff
@@ -20,11 +20,12 @@
...
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
+import javax.validation.groups.Default;
 
 public class Address {
 
-    @NotNull(message = "streetNotNull")
-    @Size(min = 3, max = 64, message = "streetSize")
+    @NotNull(message = "streetNotNull", groups = {Default.class, Person.StreetChecks.class, Person.NameAndStreetChecks.class})
+    @Size(min = 3, max = 64, message = "streetSize", groups = {Default.class, Person.StreetChecks.class, Person.NameAndStreetChecks.class})
```
lukaszlenart: `Default.class` is required?

plugins/bean-validation/src/test/resources/bean-validation-test.xml
```diff
@@ -20,8 +20,8 @@
  */
 -->
 <!DOCTYPE xwork PUBLIC
-        "-//Apache Struts//XWork 2.0//EN"
-        "http://struts.apache.org/dtds/xwork-2.0.dtd"
+        "-//Apache Struts//XWork 2.5//EN"
+        "http://struts.apache.org/dtds/xwork-2.5.dtd"
```
lukaszlenart: 👍

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
```diff
@@ -0,0 +1,50 @@
...
+ * Example, Validate with severals special group
+ * </p>
+ * <pre>
+ *  {@literal @}ValidateGroup(Default.class, CarChecks.class, DriverChecks.class)
+ *  {@literal @}Action...
+ * </pre>
+ */
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValidateGroup {
```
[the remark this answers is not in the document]
HedjuHor: Yes you are right.
shall I refactoring that?

plugins/bean-validation/src/test/java/org/apache/struts/beanvalidation/models/Address.java
HedjuHor: Yes, 
see Examples http://beanvalidation.org/1.1/spec/#validationapi-validatorapi-groups
... groups = {Default.class, Person.StreetChecks.class, Person.NameAndStreetChecks.class}) means, this Constraints is valid for on of them.

If validator called without a group ...validator.validate(foo);... it validates the group Default (implicitly). 

Testcase 'testModelDrivenActionSize' would fail

plugins/bean-validation/src/test/resources/bean-validation-test.xml
HedjuHor: ...<allowed-methods>... was the reason :-)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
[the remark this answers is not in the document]
lukaszlenart: Yes, please

plugins/bean-validation/src/test/java/org/apache/struts/beanvalidation/models/Address.java
lukaszlenart: Ok, I think I understand, `Default.class` is used when no group was defined but I would not force users to put `Default.class` in the existing annotations - this can break backward comaptibility. Instead it should be defined inside the annotation.

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
```diff
@@ -0,0 +1,50 @@
...
+
+    Class<?>[] value() default {};
```
lukaszlenart: I would put `Default.class` as a default value, in such case specifying an empty annotation won't break anything.

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
```diff
@@ -95,36 +97,42 @@ protected String doIntercept(ActionInvocation invocation) throws Exception {
...
+            performBeanValidation(action, validator, validationGroup);
         }
 
         return invocation.invoke();
     }
 
-    protected void performBeanValidation(Object action, Validator validator) {
+    protected Class<?>[] getValidationGroups(Object action, String methodName) throws NoSuchMethodException {
+        ValidateGroup validateGroup = MethodUtils.getAnnotation(getActionMethod(action.getClass(), methodName), ValidateGroup.class, true, true);
+        return validateGroup == null ? new Class[]{} : validateGroup.value();
```
lukaszlenart: With default value defined (as proposed above with `Default.class`) this check won't be needed

plugins/bean-validation/src/test/java/org/apache/struts/beanvalidation/models/Address.java
yasserzamani: It seems it's out of our control, they are annotations inside `javax.validation`, @lukaszlenart . It seems we have to use `Default.class` when we want to override default `groups` but we also want to have Default in our basket.

lukaszlenart: I was referring to the the `ValidataGroup` annotation, to put `Default.class` as a default value.

HedjuHor: if i add the Default.class to the default value in the Annotation (ValidationGroup), it becomes a depencies to the javax.validation package. Is this a good practise?

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
HedjuHor: if i add the Default.class to the default value in the Annotation (ValidationGroup), it becomes a depencies to the javax.validation package. Is this a good practise?

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
```diff
@@ -95,36 +97,42 @@ protected String doIntercept(ActionInvocation invocation) throws Exception {
         if (LOG.isDebugEnabled()) {
             LOG.debug("Validating [{}/{}] with method [{}]", invocation.getProxy().getNamespace(), invocation.getProxy().getActionName(), methodName);
         }
+        Class<?>[] validationGroup = getValidationGroups(action, methodName);
```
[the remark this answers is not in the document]
HedjuHor: Fixed
got a good hint  on..LOG.isDebugEnabled()...
https://logging.apache.org/log4j/2.0/manual/api.html 'Substituting Parameters'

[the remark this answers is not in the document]
yasserzamani: Oh! I'm sorry :( I did not mean those logging lines. I meant line#100, `Class<?>[] validationGroup = getValidationGroups(action, methodName);` This line is better to being moved inside that if, please. i.e. when user is specified a skip validation annotation, there is no need to check for validationGroup annotation :) logging lines were ok and should being reverted, thanks.

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/constraints/ValidateGroup.java
lukaszlenart: but this plugin already depends on that package so it won't be a problem :)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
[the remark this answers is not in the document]
HedjuHor: facepalm :-)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
HedjuHor: 'null' check? , get NPE whithout them if no @ValidationGroup is set on ActionMethode

yasserzamani: @lukaszlenart meant you will not need to check if null, if you accept to change annotation as below (Currently you get NPE because you didnot accept and are discussing if `javax....` package dependencies is ok)

```java
@Retention(RetentionPolicy.RUNTIME)
public @interface ValidateGroup {

    Class<?>[] value() default {Default.class};
```

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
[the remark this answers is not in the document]
yasserzamani: Please do not delete ` if (LOG.isDebugEnabled()) {`. It seems they are there because of performance when computing the string to log is expensive :) thanks!

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
HedjuHor: MethodUtils.getAnnotation(...) returns 'null' if the ActionMethode has no Annotation.

yasserzamani: 👍 you're right :) thanks! So , again , here also , to keep backward compatibility , what do you feel about following? (I'm worry because current version does not pass any group then `javax....` assumes `Default.class` then we have to keep this behavior for backward compatibility)
```java
return validateGroup == null ? new Class[]{Default.class} : validateGroup.value();
```

HedjuHor: 👍

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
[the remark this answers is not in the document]
yasserzamani: > got a good hint on..LOG.isDebugEnabled()...
> https://logging.apache.org/log4j/2.0/manual/api.html 'Substituting Parameters'

👍 you're right. Thanks! Please ignore my previous comment. I found some time to read above your mentioned link and you were ok to delete them 😉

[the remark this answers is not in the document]
rgielen: The ```if (LOG.isDebugEnabled()) {``` practice is not longer needed in all cases when using placeholders like found in the log statement. Preliminary String concatenation is avoided by this new pattern.
Nevertheless, all given parameters will be evaluated before the actual method call, especially ```invocation.getProxy().getNamespace(), invocation.getProxy().getActionName()``` - if these two calls are costly, the if-guard is still justified!

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
```diff
@@ -92,39 +95,43 @@ protected String doIntercept(ActionInvocation invocation) throws Exception {
         ActionProxy actionProxy = invocation.getProxy();
         String methodName = actionProxy.getMethod();
 
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("Validating [{}/{}] with method [{}]", invocation.getProxy().getNamespace(), invocation.getProxy().getActionName(), methodName);
-        }
+        LOG.debug("Validating [{}/{}] with method [{}]", invocation.getProxy().getNamespace(), invocation.getProxy().getActionName(), methodName);
```
yasserzamani: It seems we must revert deleting `if (LOG.isDebugEnabled()) {`. Please see [René Gielen comment](https://github.com/apache/struts/pull/200#discussion_r161456354)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
```diff
@@ -148,9 +155,7 @@ private void addBeanValidationErrors(Set<ConstraintViolation<Object>> constraint
                     if (action instanceof ModelDriven && fieldName.startsWith(ValidatorConstants.MODELDRIVEN_PREFIX)) {
                         fieldName = fieldName.replace("model.", ValidatorConstants.EMPTY_SPACE);
                     }
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug("Adding field error [{}] with message [{}]", fieldName, validationError.getMessage());
-                    }
+                    LOG.debug("Adding field error [{}] with message [{}]", fieldName, validationError.getMessage());
```
yasserzamani: It seems we must revert deleting `if (LOG.isDebugEnabled()) {`. Please see [René Gielen comment](https://github.com/apache/struts/pull/200#discussion_r161456354)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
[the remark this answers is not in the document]
yasserzamani: 👍 sorry... I failed to catch this because of some workload that day :( Thanks!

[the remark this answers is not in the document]
rgielen: No worries :)

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
lukaszlenart: In this case it's safe to leave as is, there is no additional cost to fetch these values - it's a chain of getters.

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
lukaszlenart: Same here, no additional cost to get `message`

plugins/bean-validation/src/main/java/org/apache/struts/beanvalidation/validation/interceptor/BeanValidationInterceptor.java
yasserzamani: Thank you very much! I approved, however I'm not familiar with bean validation but it covers all codes and not fails. If you're familiar and ok then please merge :)

PR #204, opened 2018-01-19
WW-4910 Align OptGroup with Select

quaff:
1. allow set freemarker sequence or hash as list
2. if list is a List, use 'top' as listKey and listValue

Implements [WW-4910](https://issues.apache.org/jira/browse/WW-4910)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/ListUIBean.java
+++ b/core/src/main/java/org/apache/struts2/components/ListUIBean.java
@@ -70,7 +70,7 @@ public abstract class ListUIBean extends UIBean {
 
         if (list instanceof String) {
             value = findValue((String) list);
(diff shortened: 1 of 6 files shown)


lukaszlenart:
It would be good to add a test to cover this case.

quaff:
Updated with test case.

yasserzamani:
@quaff , thanks a lot. I also discovered a few issues during reviewing which I fixed in three pushes into your branch :)

@apache/struts-committers , LGTM , However, I also pushed some changes, so, Once reviewed, I think you can squash and merge it.

lukaszlenart:
2.5.16 is out so we can start working on 2.6 :)

lukaszlenart:
👍

PR #205, opened 2018-01-25
Initial test dedicated for the DefaultUnknownHandlerManager

zalsaeed:
This is a very trivial test dedicated for the DefaultUnknownHandlerManager. 

I'm doing some research that includes Apache Struts as an example. Most of the classes I'm interested in have sufficient tests. However, even though DefaultUnknownHandlerManager is tested through other tests created for other classes, those tests are not sufficient enough for my purpose. I'm pushing the given test for DefaultUnknownHandlerManager to confirm that it is acceptable by you.

### DIFF
--- a/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
@@ -1,0 +1,69 @@
+package com.opensymphony.xwork2;
+
+import com.opensymphony.xwork2.config.entities.ActionConfig;
+import com.opensymphony.xwork2.config.providers.SomeUnknownHandler;
+
+import java.util.ArrayList;
+
+/**
+ * Partial test to the DefaultUnknownHandlerManager to understand the relationship between Manager and Handlers.
+ *
+ * @author Ziyad Alsaeed
+ */
+public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
+
+    ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
+    ActionConfig actionConfig = actionConfigBuilder.build();
+    SomeUnknownHandler someUnknownHandler = new SomeUnknownHandler();
+
+    /**
+     * Relationshsip when UnknownAction method is called.
+     *
+     * @author Ziyad Alsaeed
+     */
+    public void testHandleUnknownAction() {
+
+        DefaultUnknownHandlerManager defaultUnknownHandlerManager = new DefaultUnknownHandlerManager();
+        defaultUnknownHandlerManager.unknownHandlers = new ArrayList<>();
+        defaultUnknownHandlerManager.unknownHandlers.add(someUnknownHandler);
+
+        ActionConfig newActionConfig = defaultUnknownHandlerManager.handleUnknownAction("arbitraryNameSpace", "arbitraryActionName");
+
+        assertEquals(newActionConfig, actionConfig);
+    }
+
+    /**
+     * Relationship when UnknownActionMethod method called.
+     *
+     * @author Ziyad Alsaeed
+     */
+    public void testHandelUnknownActionMethod() throws Exception {
+        DefaultUnknownHandlerManager defaultUnknownHandlerManager = new DefaultUnknownHandlerManager();
+        defaultUnknownHandlerManager.unknownHandlers = new ArrayList<>();
+        defaultUnknownHandlerManager.unknownHandlers.add(someUnknownHandler);
+
+        String result = null;
+
+        for (int i = 0; i < 10 ; i++) {
+            result = (String) defaultUnknownHandlerManager.handleUnknownMethod(this, "someMethodName");
+            assertEquals(result, "specialActionMethod");
+        }
+
+    }
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+
+        // Make sure we are using the actionConfig we initialized.
+        someUnknownHandler.setActionConfig(actionConfig);
+        someUnknownHandler.setActionMethodResult("specialActionMethod");
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+
+    }
+
+}
--- a/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
@@ -29,9 +29,8 @@ import java.util.ArrayList;
  */
 public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
 
-    ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
-    ActionConfig actionConfig = actionConfigBuilder.build();
-    SomeUnknownHandler someUnknownHandler = new SomeUnknownHandler();
+    ActionConfig actionConfig;
+    SomeUnknownHandler someUnknownHandler;
 
     /**
      * Relationshsip when UnknownAction method is called.
@@ -71,6 +70,10 @@ public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
         super.setUp();
 
         // Make sure we are using the actionConfig we initialized.
+        ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
+        actionConfig = actionConfigBuilder.build();
+        someUnknownHandler = new SomeUnknownHandler();
+
         someUnknownHandler.setActionConfig(actionConfig);
         someUnknownHandler.setActionMethodResult("specialActionMethod");
     }
--- a/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


zalsaeed:
I'm not seeing the relationship between what I have changed and the error given by Jenkins. Can someone explain please to fix it?

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,69 @@
+package com.opensymphony.xwork2;
```
yasserzamani: Please copy and paste the LICENCE text from a neighbor test .java file to beginning of this file to pass Jenkins build :)

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.opensymphony.xwork2.config.entities.ActionConfig;
+import com.opensymphony.xwork2.config.providers.SomeUnknownHandler;
+
+import java.util.ArrayList;
+
+/**
+ * Partial test to the DefaultUnknownHandlerManager to understand the relationship between Manager and Handlers.
+ *
+ * @author Ziyad Alsaeed
```
yasserzamani: Apache doesn't accept author tag anymore. Instead see [here](https://www.apache.org/foundation/how-it-works.html#roles) if you're interested to contribute :)

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,69 @@
...
+public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
+
+    ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
+    ActionConfig actionConfig = actionConfigBuilder.build();
+    SomeUnknownHandler someUnknownHandler = new SomeUnknownHandler();
+
+    /**
+     * Relationshsip when UnknownAction method is called.
+     *
+     * @author Ziyad Alsaeed
```
yasserzamani: (same here)

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        ActionConfig newActionConfig = defaultUnknownHandlerManager.handleUnknownAction("arbitraryNameSpace", "arbitraryActionName");
+
+        assertEquals(newActionConfig, actionConfig);
+    }
+
+    /**
+     * Relationship when UnknownActionMethod method called.
+     *
+     * @author Ziyad Alsaeed
```
yasserzamani: (same here)

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
zalsaeed: thanks, this is clear and fixed.

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Partial test to the DefaultUnknownHandlerManager to understand the relationship between Manager and Handlers.
+ *
+ */
+public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
+
+    ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
+    ActionConfig actionConfig = actionConfigBuilder.build();
+    SomeUnknownHandler someUnknownHandler = new SomeUnknownHandler();
```
lukaszlenart: It's a good practice to initialise such things in `setUp` method.

zalsaeed: you are right, fixed.

lukaszlenart: 👍

lukaszlenart:
Looks good to me, let @yasserzamani to pull the trigger :)

yasserzamani:
I see there is already a dedicated test for `DefaultUnknownHandlerManager` at [/core/src/test/java/com/opensymphony/xwork2/util/UnknownHandlerManagerTest.java](https://github.com/apache/struts/blob/master/core/src/test/java/com/opensymphony/xwork2/util/UnknownHandlerManagerTest.java). I thought you may want to see it also and merge your ones into it if applicable or move the whole .java file beside it in same package. (I see that has tested `handleUnknownMethod` but missing `handleUnknownAction` which you have)

zalsaeed:
This is a tricky one. I'm not an expert in struts but this is slightly different from the `DefaultUnknownHandlerManager` I'm testing. I have ran all the available tests before and the lines of `DefaultUnknownHandlerManager` were not touched throughly. I can move the whole file under the util package, but wouldn't this be wrong? Because the class is within package `/core/src/test/java/com/opensymphony/xwork2/`.

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,87 @@
...
+import com.opensymphony.xwork2.config.entities.ActionConfig;
+import com.opensymphony.xwork2.config.providers.SomeUnknownHandler;
+
+import java.util.ArrayList;
+
+/**
+ * Partial test to the DefaultUnknownHandlerManager to understand the relationship between Manager and Handlers.
+ *
+ */
+public class DefaultUnknownHandlerManagerTest extends XWorkTestCase {
```
yasserzamani: I tested `extends TestCase` and works. It seems you can downgrade to `TestCase` to save performance.

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,87 @@
...
+        ActionConfig.Builder actionConfigBuilder = new ActionConfig.Builder( "com", "someAction", "someClass");
+        actionConfig = actionConfigBuilder.build();
+        someUnknownHandler = new SomeUnknownHandler();
+
+        someUnknownHandler.setActionConfig(actionConfig);
+        someUnknownHandler.setActionMethodResult("specialActionMethod");
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
```
yasserzamani: It seems you can downgrade to `TestCase` to save performance and delete this.

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
```diff
@@ -0,0 +1,87 @@
...
+     *
+     */
+    public void testHandelUnknownActionMethod() throws Exception {
+        DefaultUnknownHandlerManager defaultUnknownHandlerManager = new DefaultUnknownHandlerManager();
+        defaultUnknownHandlerManager.unknownHandlers = new ArrayList<>();
+        defaultUnknownHandlerManager.unknownHandlers.add(someUnknownHandler);
+
+        String result = null;
+
+        for (int i = 0; i < 10 ; i++) {
```
yasserzamani: Why 10 times? I'm just curious :)

yasserzamani:
> This is a tricky one. I'm not an expert in struts but this is slightly different from the DefaultUnknownHandlerManager I'm testing. I have ran all the available tests before and the lines of DefaultUnknownHandlerManager were not touched throughly.

Hm.. but coverall says it's almost fully covered in last master build before this pr at [here](https://coveralls.io/builds/15231921/source?filename=core%2Fsrc%2Fmain%2Fjava%2Fcom%2Fopensymphony%2Fxwork2%2FDefaultUnknownHandlerManager.java). I also reviewed that previous test case and saw it already has tested `handleUnknownMethod`. It seems you can add your new `handleUnknownAction` test to it as a new test method.

> I can move the whole file under the util package, but wouldn't this be wrong? Because the class is within package /core/src/test/java/com/opensymphony/xwork2/.

Hm.. you're right. It seems it was there previously and moved here with no moved tests :)

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
zalsaeed: I was following other test cases here. I will change it and push it if tests pass.

core/src/test/java/com/opensymphony/xwork2/DefaultUnknownHandlerManagerTest.java
zalsaeed: It is an arbitrary number. I wanted to see the difference this will make on what I'm developing.

zalsaeed:
I think wasn't clear when I said "were not touched thoroughly". What I meant is not related to if a line is touched or not (the general definition of test coverage). After all as I mentioned before, the class `DefaultUnknownHandlerManager` was tested through other tests cases that were dedicated to other classes. But that doesn't mean it was tested thoroughly. For any given source code there is the space of possible paths it could take and that space is hard to cover sometimes %100. Here I was trying to hit some of those missed paths. Please let me know if this is still not clear. 

> I also reviewed that previous test case and saw it already has tested handleUnknownMethod. It seems you can add your new handleUnknownAction test to it as a new test method.

> Hm.. you're right. It seems it was there previously and moved here with no moved tests :)

But still there is no test class with the name `DefaultUnknownHandlerManagerTest` I still believe that it much clearer if you name each test class with the name of the class you are targeting. Nevertheless, I'm not %100 sure what coding style you follow. So, what do you think, should I keep the name of the class I pushed and maybe merge the test cases from `UnknownHandlerManagerTest`, do the opposite, or keep both classes since there exist two different classes (in the sources code `DefaultUnknownHandlerManager` and `UnknownHandlerManager`) and it is good to have a test class for each one?

yasserzamani:
No I also agree and your class name is good. And you're right, let keep both files as is and as @lukaszlenart 
 confirmed, I think I can merge it now while we discussed all possible things :)

Thanks 👍

PR #206, opened 2018-01-26
WW-4912: scan package-info and class at the same time

lukaszlenart:
This change allows to define `AllowedMethods` in `package-info.java` and all class level and all those annations will be examined to build set of allowed methods.

Fixes [WW-4912](https://issues.apache.org/jira/browse/WW-4912)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
@@ -140,13 +140,16 @@ public class AnnotationUtils {
      */
     public static <T extends Annotation> T findAnnotation(Class<?> clazz, Class<T> annotationClass) {
         T ann = clazz.getAnnotation(annotationClass);
-        if (ann == null && clazz != Object.class) {
+        while (ann == null && clazz != null) {
             ann = clazz.getAnnotation(annotationClass);
             if (ann == null) {
                 ann = clazz.getPackage().getAnnotation(annotationClass);
             }
-            if (ann == null && clazz != Object.class) {
-                ann = findAnnotation(clazz.getSuperclass(), annotationClass);
+            if (ann == null) {
+                clazz = clazz.getSuperclass();
+                if (clazz != null) {
+                    ann = clazz.getAnnotation(annotationClass);
+                }
             }
         }
 
--- a/core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
@@ -18,6 +18,8 @@
  */
 package com.opensymphony.xwork2.util;
 
+import org.apache.commons.lang3.ClassUtils;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
```diff
@@ -151,4 +152,33 @@ public static String resolvePropertyName(Method method) {
...
     }
+
+    /**
+     * Returns the annotation on the given class or the package of the class. This searches up the
+     * class hierarchy and the package hierarchy for the closest match.
+     *
+     * @param <T>             class type
+     * @param clazz           The class to search for the annotation.
+     * @param annotationClass The Class of the annotation.
+     * @return The annotation or null.
```
HedjuHor: never return 'null'

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
```diff
@@ -151,4 +152,33 @@ public static String resolvePropertyName(Method method) {
...
+            anns.add(ann);
+        }
+
+        ann = clazz.getPackage().getAnnotation(annotationClass);
+        if (ann != null) {
+            anns.add(ann);
+        }
+
+        if (clazz.getSuperclass() != Object.class) {
+            anns.addAll(findAnnotations(clazz.getSuperclass(), annotationClass));
```
HedjuHor: if someone calls this method with Object.class, the recursive call ends with NPE

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
aleksandr-m: never returns `null`

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
```diff
@@ -139,19 +140,45 @@ public static String resolvePropertyName(Method method) {
      */
     public static <T extends Annotation> T findAnnotation(Class<?> clazz, Class<T> annotationClass) {
         T ann = clazz.getAnnotation(annotationClass);
-        while (ann == null && clazz != null) {
+        if (ann == null && clazz != Object.class) {
             ann = clazz.getAnnotation(annotationClass);
```
yasserzamani: As you replaced `while` with `if` then I think you can correct this method as below:

```java
    public static <T extends Annotation> T findAnnotation(Class<?> clazz, Class<T> annotationClass) {
        T ann = null;
        if (clazz != null) {
            ann = clazz.getAnnotation(annotationClass);
            if (ann == null) {
                ann = clazz.getPackage().getAnnotation(annotationClass);
            }
            if (ann == null) {
                ann = findAnnotation(clazz.getSuperclass(), annotationClass);
            }
        }

        return ann;
    }
```

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
```diff
@@ -139,19 +140,45 @@ public static String resolvePropertyName(Method method) {
...
+        if (ann != null) {
+            anns.add(ann);
+        }
+
+        ann = clazz.getPackage().getAnnotation(annotationClass);
+        if (ann != null) {
+            anns.add(ann);
+        }
+
+        if (clazz.getSuperclass() != Object.class) {
```
yasserzamani: What do you feel with this instead:

```java
        ann = findAnnotations(clazz.getSuperclass(), annotationClass);
        if (ann != null) {
            anns.add(ann);
        }
```

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
lukaszlenart: Sorry, I forgot to update the JavaDoc :)

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
lukaszlenart: I think that's good, calling this function in `Object` is a bad idea anyway.

core/src/main/java/com/opensymphony/xwork2/util/AnnotationUtils.java
lukaszlenart: `findAnnotations` does not return `null` also I don't want to scan above `Object` as it doesn't make sense

yasserzamani:
Thanks! LGTM 👍

lukaszlenart:
@yasserzamani could you hit the button? :)

yasserzamani:
Yeah I was going to push :) but said myself, now I know it works but let see once time again if any further improvements are possible. Now I'm preparing some nice improvements as another PR on this PR 😉 improvements like supporting annotations on interfaces in other packages also :) plus more strict tests e.g. one of my tests showed `findAnotations` will have duplicate results when searches in same package.

lukaszlenart:
Hm... not sure what's the problem with duplications? If we scan the whole class hierarchy having duplications is rather expected.

yasserzamani:
I meant if class A extends B and implements I and A and B and I are in same package P and P is annotated by annotation ANN then it will have three references of ANN in returned list. I felt it's somehow not nice but in another side, now, I also am worry about performance 😟

lukaszlenart:
That's why I would leave it as is and add a task to improve that in 2.6 - tbh, I would like to start a release process of 2.5.15 ASAP

yasserzamani:
👍 Just also please see [here](https://github.com/lukaszlenart/struts/pull/1/files). There are a few but nice improvements which don't need a lot time to review. Please feel free to merge them or just inform me and I'll merge this PR without them if you prefer then we can postpone them and performance improvements to `2.6` :)

yasserzamani:
Great work 👍 thanks!

PR #207, opened 2018-01-31
WW-4741: Do not create session

lukaszlenart:
Resolves [WW-4741](https://issues.apache.org/jira/browse/WW-4741)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
@@ -281,15 +281,16 @@ public class I18nInterceptor extends AbstractInterceptor {
 
         @Override
         public Locale store(ActionInvocation invocation, Locale locale) {
-            //save it in session
-            Map<String, Object> session = invocation.getInvocationContext().getSession();
+            LOG.debug("Do not create session if it doesn't exist");
+            HttpSession session = ServletActionContext.getRequest().getSession(false);
 
             if (session != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
```diff
@@ -281,15 +281,17 @@ public Locale find() {
...
+                String sessionId = session.getId();
                 synchronized (sessionId.intern()) {
-                    session.put(attributeName, locale);
+                    invocation.getInvocationContext().getSession().put(attributeName, locale);
                 }
+            } else {
+                LOG.debug("session creation avoided as it doesn't exist already");
             }
+
             return locale;
```
yasserzamani: When `shouldStore` field is not false (i.e. we should store), shouldn't we forcefully create a session and store the locale? if we should not, shouldn't we return `null` here?

lukaszlenart: Maybe we should forcefully create the session but I would leave that to a user, if there was no session just ignore (a user fault). And returning `null` can break action processing as then the null will be set in `ActionContext`

yasserzamani:
LGTM 👍 Nice catch! Thanks!

core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
yasserzamani: 👍

PR #208, opened 2018-02-05
WW-4915 Replace deprecated commons-lang3 classes

aleksandr-m:
`org.apache.commons.lang3.StringEscapeUtils.escapeXml` was replaced with `org.apache.commons.text.StringEscapeUtils.escapeXml10` which escapes more characters and potentially can break some things.
@apache/struts-committers


lukaszlenart:
This introduces a new dependency and I would wait with merging this till 2.6, wdyt?

aleksandr-m:
Sure. The `escapeXml` change will also be more reasonable then.

lukaszlenart:
LGTM 👍

PR #210, opened 2018-02-20
WW-4922: Jackson XML

lukaszlenart:
Implements [WW-4922](https://issues.apache.org/jira/browse/WW-4922)

### DIFF
--- a/apps/rest-showcase/src/main/resources/struts.xml
+++ b/apps/rest-showcase/src/main/resources/struts.xml
@@ -32,6 +32,7 @@
 
     <constant name="struts.convention.package.locators" value="example"/>
 
+<!-- Uncomment the lines below to use Jackson XML bindings instead of the XStream library to handle XML serialisations -->
 <!--
     <bean name="jacksonXml" type="org.apache.struts2.rest.handler.ContentTypeHandler" class="org.apache.struts2.rest.handler.JacksonXmlHandler" />
     <constant name="struts.rest.handlerOverride.xml" value="jacksonXml"/>
--- a/apps/rest-showcase/src/main/resources/struts.xml
+++ b/apps/rest-showcase/src/main/resources/struts.xml
@@ -32,6 +32,11 @@
 
     <constant name="struts.convention.package.locators" value="example"/>
 
+<!--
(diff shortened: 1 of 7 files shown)


yasserzamani:
LGTM :+1: I'm just curious if we can move these to `struts-examples` (because I guess user already can do these himself) ?

lukaszlenart:
Yes, they can but the long plan is to have just a one library to handle JSON/XML serialisation. Also having this in an official package gives users confidence.

yasserzamani:
So wouldn't you like to set it as default now (like what we did with json plugin) ?

apps/rest-showcase/src/main/resources/struts.xml
```diff
@@ -32,6 +32,11 @@
 
     <constant name="struts.convention.package.locators" value="example"/>
 
+<!--
+    <bean name="jacksonXml" type="org.apache.struts2.rest.handler.ContentTypeHandler" class="org.apache.struts2.rest.handler.JacksonXmlHandler" />
+    <constant name="struts.rest.handlerOverride.xml" value="jacksonXml"/>
+-->
+
```
sdutry: do these commented lines serve any purpose?

lukaszlenart: An example how to use it but maybe I should move it to struts-examples, wdyt?

yasserzamani: If so, you'll have to wait for 2.5.16 release.

lukaszlenart: Yes I know, it's not a problem.

yasserzamani: As this is a showcase, I think you also have this option to uncomment and let showcase use jackson; a showcase for how override rest content handler.

lukaszlenart: I added a comment how to use the commented configuration

yasserzamani:
Yes I am. I just was in wait for @sdutry if has something with your question:
> ... maybe I should move it to struts-examples, wdyt?

lukaszlenart:
I see no objections :)

yasserzamani:
Me too :)

sdutry:
I'm sorry for my late response. I didn't mean to be a blocker for the merge.
I've just been busy IRL lately.

lukaszlenart:
@sdutry don't worry, everything is fine :)

PR #211, opened 2018-02-20
WW-4920 fix java.net.JarURLConnection#parseSpecs

yasserzamani:
I inspired such fix from Spring Framework when I found it during my research at [here](https://github.com/spring-projects/spring-framework/blob/8d3264f680018ccbec6a40253b477283aa3a7e6e/spring-core/src/main/java/org/springframework/util/ResourceUtils.java#L316).

Oracle WebLogic may report jar urls like "zip:C:/web-app-lib-path/some-jar.jar" but JDK's JarURLConnection#parseSpecs breaks on such urls.

While JDK's JarURLConnection#parseSpecs is private then I had to extend URLConnection instead.

Struts keeps a reference of all loaded files regardless of `devMode` or `struts.configuration.xml.reload` values (i.e. user may get a warning exception if framework was not able to create a revision for loaded file). But Struts checks if reload is needed only when `struts.configuration.xml.reload` constant is set to
true (`devMode` still doesn't make sense). So, javadocs of `FileManager` interface also fixed to keep it consistent with current implementation.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
@@ -38,10 +38,10 @@ import java.util.jar.JarFile;
 /**
  * WW-4901 If was needed, decouples from underlying implementation of {@link URL#openConnection()}
  * e.g. from IBM WebSphere com.ibm.ws.classloader.Handler$ClassLoaderURLConnection
- * Also decouples from and fixes {@link JarURLConnection#parseSpecs(URL)} if was needed
+ * WW-4920 Also decouples from and fixes {@link JarURLConnection#parseSpecs(URL)} if was needed
  * e.g. from Oracle WebLogic which may report jar urls like "zip:C:/web-app-lib-path/some-jar.jar"
  * but {@link JarURLConnection#parseSpecs(URL)} breaks on such urls
- * While {@link JarURLConnection#parseSpecs(URL)} is private then we had too extends {@link URLConnection} instead
+ * While {@link JarURLConnection#parseSpecs(URL)} is private, then we had to extend {@link URLConnection} instead
  * @since 2.5.15
  */
 class StrutsJarURLConnection extends URLConnection {
--- a/core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
@@ -18,14 +18,14 @@
  */
 package com.opensymphony.xwork2.util.fs;
 
-import sun.net.www.ParseUtil;
-
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.JarURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.net.URLDecoder;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -71,7 +71,7 @@ class StrutsJarURLConnection extends URLConnection {
     /**
     * A fixed copy of {@link JarURLConnection#parseSpecs(URL)}
     */
-    private void parseSpecs(URL url) throws MalformedURLException {
+    private void parseSpecs(URL url) throws MalformedURLException, UnsupportedEncodingException {
         String spec = url.getFile();
 
         int separator = spec.indexOf("!/");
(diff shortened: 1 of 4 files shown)


lukaszlenart:
I think all the issues were addressed in the previous PR and we can merge it, @aleksandr-m any other objections?

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
```diff
@@ -20,34 +20,89 @@
...
 
 /**
- * WW-4901 Decouples from underlying implementation of {@link URL#openConnection()}
+ * WW-4901 If was needed, decouples from underlying implementation of {@link URL#openConnection()}
  * e.g. from IBM WebSphere com.ibm.ws.classloader.Handler$ClassLoaderURLConnection
+ * WW-4920 Also decouples from and fixes {@link JarURLConnection#parseSpecs(URL)} if was needed
+ * e.g. from Oracle WebLogic which may report jar urls like "zip:C:/web-app-lib-path/some-jar.jar"
+ * but {@link JarURLConnection#parseSpecs(URL)} breaks on such urls
+ * While {@link JarURLConnection#parseSpecs(URL)} is private, then we had to extend {@link URLConnection} instead
  * @since 2.5.15
```
aleksandr-m: `2.5.16` or just remove it.

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
```diff
@@ -20,34 +20,89 @@
...
     }
 
-    @Override
-    public JarFile getJarFile() throws IOException {
-        connect();
-        return jarFile;
+    /**
+    * A fixed copy of {@link JarURLConnection#parseSpecs(URL)}
+    */
+    private void parseSpecs(URL url) throws MalformedURLException, UnsupportedEncodingException {
```
aleksandr-m: No unit tests for that. Or does `JarEntryRevisionTest` covers all?

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
yasserzamani: Thanks! Did you mean only here which is modified in 2.5.16? Or all "2.5.15" occurrences? Because there are also other "2.5.15" occurrences in other .java files e.g. BackgoundProcess.java.

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
yasserzamani: Yes it covers all according to [cover all report for this file](https://coveralls.io/builds/15608083/source?filename=core%2Fsrc%2Fmain%2Fjava%2Fcom%2Fopensymphony%2Fxwork2%2Futil%2Ffs%2FStrutsJarURLConnection.java#L74). But if you still prefer, I can write a sepearete test case named StrutsJarURLConnectionTest.java. Thanks!

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
lukaszlenart: We can leave it, there is a tag for 2.5.15 and it didn't just pass the verification process to be a full release.

aleksandr-m: Ah, didn't noticed that this class was already there in 2.5.15. Then it is ok.

core/src/main/java/com/opensymphony/xwork2/util/fs/StrutsJarURLConnection.java
aleksandr-m: We can do it after the merge, but it would be great to have dedicated tests.

yasserzamani: I tried but it seems JUnit does not allow calling private methods. Then I researched and found [this answer](https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods):

> You generally don't unit test private methods directly. Since they are private, consider them an implementation detail. Nobody is ever going to call one of them and expect it to work a particular way.

rgielen:
If you feel a private method justifies testing, that is not being of trivial complexity, you might consider switching visibility to package local

yasserzamani:
@aleksandr-m , as this method is a copy from JDK with a small fix, and as it already has tested and all covered via `JarEntryRevisionTest `, then I think it may don't justify testing enough and please let me merge this as is for now :question:

aleksandr-m:
@yasserzamani As I've said, just a nitpicks. So go on.

lukaszlenart:
Looks good, LGTM 👍

PR #213, opened 2018-02-22
WW-4921 revert I18nInterceptor logic from WW-4741 changes…

yasserzamani:
… to it's previous stable logic

WW-4741 changes logic which breaks WW-4921. This commit if applied, will revert the logic to previous stable state but keeps WW-4741 fixed and also will fix WW-4921.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
@@ -282,14 +282,15 @@ public class I18nInterceptor extends AbstractInterceptor {
         @Override
         public Locale store(ActionInvocation invocation, Locale locale) {
             HttpSession session = ServletActionContext.getRequest().getSession(false);
+            Map<String, Object> invocationSession = invocation.getInvocationContext().getSession();
 
-            if (session != null) {
+            if (session != null && invocationSession != null) {
                 String sessionId = session.getId();
                 synchronized (sessionId.intern()) {
-                    invocation.getInvocationContext().getSession().put(attributeName, locale);
+                    invocationSession.put(attributeName, locale);
                 }
             } else {
-                LOG.debug("session creation avoided as it doesn't exist already");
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
```diff
@@ -282,14 +282,15 @@ public Locale find() {
         @Override
         public Locale store(ActionInvocation invocation, Locale locale) {
             HttpSession session = ServletActionContext.getRequest().getSession(false);
+            Map<String, Object> invocationSession = invocation.getInvocationContext().getSession();
 
-            if (session != null) {
+            if (session != null && invocationSession != null) {
```
lukaszlenart: I'm not sure if this is the best approach, `invocationSession` cannot be normally `null`, it's always initialised with `Dispatcher`. It can be empty but not `null`. I would rather solve this by initialising a test environment properly.

yasserzamani: :+1: thanks! then I think a similar fix like [https://issues.apache.org/jira/browse/WW-3442](WW-3442) seems good.

lukaszlenart: Yes and know, the my fix had introduced a regression and I wonder if we can fix that without introducing the above change.

yasserzamani: > the my fix

Did you mean WW-3442? Or WW-4741?

lukaszlenart: WW-4741

yasserzamani: That is a good fix :+1: I thought when httpSession is null, then invocationSession also will be null but as you corrected me, it won't be but will be empty. So this PR isn't valid anymore and I like your fix instead :)

PR #214, opened 2018-02-27
WW-4921 initialize session like WW-3442

yasserzamani:


plugins/junit/src/main/java/org/apache/struts2/StrutsJUnit4TestCase.java
```diff
@@ -170,6 +168,20 @@ protected ActionProxy getActionProxy(String uri) {
         return proxy;
     }
 
+    private void initActionContext(ActionContext actionContext) {
```
lukaszlenart: Could you make this method `protected`? This will allow override it when needed

plugins/junit/src/main/java/org/apache/struts2/StrutsJUnit4TestCase.java
```diff
@@ -170,6 +168,20 @@ protected ActionProxy getActionProxy(String uri) {
...
     }
 
+    private void initActionContext(ActionContext actionContext) {
+        actionContext.setParameters(HttpParameters.create(request.getParameterMap()).build());
+        initSession(actionContext);
+        // set the action context to the one used by the proxy
+        ActionContext.setContext(actionContext);
+    }
+
+    private void initSession(ActionContext actionContext) {
```
lukaszlenart: Same here, making methods protected allows users to override them.

plugins/junit/src/main/java/org/apache/struts2/StrutsJUnit4TestCase.java
yasserzamani: :+1: thanks!

plugins/junit/src/main/java/org/apache/struts2/StrutsJUnit4TestCase.java
yasserzamani: Ah... again I forgot this practice. Those are IntelliJ auto created methods. Thanks! fixed :)

lukaszlenart:
LGTM 👍

PR #215, opened 2018-03-02
WW-685 Add generic type conversion error message

aleksandr-m:
Another one for 2.6. :)

Note: Because of the `Object` in conversion errors map (`Map<String, Object> conversionErrors`) there is no type check when switching to `Map<String, ConversionData>`, which can lead to some errors.

Note: The `ConversionData` can be improved and used later in other places.


lukaszlenart:
LGTM 👍

PR #219, opened 2018-05-10
WW-4937 Add SortedSet support to JSON plugin

loomchild:
Currently when I try to deserialize a bean that contains fields of type SortedSet, I get the following exception: `java.lang.IllegalArgumentException: argument type mismatch`. This is due to the fact that JSONPopulator does not distinguish between Set and SortedSet interface. 

I added this support by creating a TreeSet for SortedSet fields. I also added a couple of tests to show the issue and confirm that it's fixed. 

I tried to follow the local code style, but since it's my first contribution to Struts project, I might have made some mistakes. I would be very happy to hear any comments and make necessary changes to this pull request. I truly appreciate your hard work of maintaining this project.

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
@@ -207,6 +207,8 @@ public class JSONPopulator {
                     newCollection = new TreeSet();
                 } else if (Set.class.isAssignableFrom(clazz)) {
                     newCollection = new HashSet();
+                } else if (Queue.class.isAssignableFrom(clazz)) {
+                    newCollection = new ArrayDeque();
                 } else {
                     newCollection = new ArrayList();
                 }
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
@@ -203,7 +203,9 @@ public class JSONPopulator {
                 newCollection = (Collection) clazz.newInstance();
             } catch (InstantiationException ex) {
                 // fallback if clazz represents an interface or abstract class
-                if (Set.class.isAssignableFrom(clazz)) {
+                if (SortedSet.class.isAssignableFrom(clazz)) {
+                    newCollection = new TreeSet();
+                } else if (Set.class.isAssignableFrom(clazz)) {
                     newCollection = new HashSet();
                 } else {
                     newCollection = new ArrayList();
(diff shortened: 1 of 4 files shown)


lukaszlenart:
Nice, could you create a ticket in [JIRA](https://issues.apache.org/jira/projects/WW/)? This will allow control us what has changed. Thanks!

loomchild:
Sure, done. Sorry, I somehow missed that you use JIRA.

yasserzamani:
Thanks! Good catch! Does it fail also with `java.util.NavigableSet` :thinking: And `java.util.Queue` and `java.util.Deque`.

loomchild:
It works with NavigableSet out of the box, since NavigableSet is SortedSet and TreeSet is both NavigableSet and SortedSet (added a test).

I added support for Deque and Queue via ArrayDeque in separate commit.

loomchild:
Yes, it won't work. However, I believe supporting standard library types is a step in the right direction before implementing a complete solution covering all possible cases, don't you think?

yasserzamani:
As it seems it isn't possible, then yes, I think so. However, if you mean jdk types by standard library types, then it seems there are a lot of :disappointed: I see following for `java.util.collection` at [here](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html):

> All Known Subinterfaces:
>     BeanContext, BeanContextServices, BlockingDeque<E>, BlockingQueue<E>, Deque<E>, List<E>, NavigableSet<E>, Queue<E>, Set<E>, SortedSet<E>, TransferQueue<E> 

Each of them also has subinterfaces if click.

loomchild:
Blocking collections are a bit specific - I don't think anyone would expose them via getters and setters, but who knows. I don't know what BeanContext is, but it's not a standard collection I believe. 

Everything is possible (for example by allowing library clients to register their own factories for given interfaces), but it will require more work than I can put in it at the moment and it will require some architectural choices in the project that I can't make. Besides, nobody asked for it, so probably nobody needs it. I would like to solve a specific requirement of a project I am currently working on for my client by adding support for SortedSet.

lukaszlenart:
I think this PR is ok, I mean it fixes a basic problem. And based on that we can think of of a potential solution how support other types, but this will be out of scope of the PR.

LGTM 👍

yasserzamani:
Yeah I agree @loomchild, let keep it simple :) until potential user demand. @lukaszlenart , In future, we can provide user an annotation which can being used to specify concrete class of the interface. e.g.

```
@deserialize(as = TreeSet.class)
interface MySet extends SortedSet
```
then in that catch block, we can read it and create a new instance from that instead of interface :)

loomchild:
Thanks a lot for merging it guys. Keep up with the good work.

PR #224, opened 2018-05-21
WW-4779 Remove profiling layer

sdutry:


aleksandr-m:
Do we need all these empty `finally` blocks?

sdutry:
@aleksandr-m 
you're right, i'll remove the try blocks with only an empty finally

sdutry:
and it seems i also missed removing profileKey Strings

lukaszlenart:
Astonish work! I wonder why coverage decreased so much ... you have dropped a bunch of lines and I would expect that the coverage should go high 🤔

yasserzamani:
I clicked on "coverage/coveralls.....[Details](https://coveralls.io/builds/17093747)" link below. It's seems there is a problem with our configuration as coveralls thinks previous un-covered lines are new for this PR. See [here](https://coveralls.io/builds/17093747/source?filename=core%2Fsrc%2Fmain%2Fjava%2Fcom%2Fopensymphony%2Fxwork2%2FDefaultActionInvocation.java#L258).

lukaszlenart:
It's ok, the code was re-formatted only after removing `try-finally` block.

sdutry:
@yasserzamani 

> Maybe your master branch was not updated when you created this pull.

Don't worry, the start situation before my changes was the state of master branch yesterday morning.

It's just the github diff that doesn't correctly show indentation changes after removing try blocks which no longer had a catch or finally

yasserzamani:
You both are right :+1: It seems coveralls also isn't able to work correctly when there are such changes.

lukaszlenart:
LGTM 👍 (if no objections)

yasserzamani:
Same here :+1: Thanks @sdutry

aleksandr-m:
Coveralls results are still weird (https://markmail.org/message/lc6yl2zhihgznwfu).

sdutry:
@aleksandr-m 
> Coveralls results are still weird.

Can i merge the pull request or do you need it open to investigate the coveralls issue?

yasserzamani:
@aleksandr-m I think it isn't same as our first issue with coveralls. I think in this issue, coveralls isn't able to distinguish new lines correctly. Don't you?

aleksandr-m:
@sdutry 

> Can i merge the pull request or do you need it open to investigate the coveralls issue?

Merge it. :+1: 

@yasserzamani 

> I think in this issue, coveralls isn't able to distinguish new lines correctly.

But they were covered before, right?

I suggest to try another coverage tool or just don't use coverall. Currently it is useless.

yasserzamani:
@aleksandr-m I personally like it a lot as doesn't allow us to push not tested new stuffs. However there are some limitations on very specific conditions which I mentioned [here](https://github.com/apache/struts/pull/226#issuecomment-391674949). I think they're acceptable. Do you know any better free tool which doesn't have such limitations?

PR #226, opened 2018-05-24
dummy temporal commit to test coveralls on ambiguous changes

yasserzamani:


yasserzamani:
Results:

1.  first commit shows coveralls may consider moved already not covered lines in some changes (rearranging methods locations) as new not covered lines (however doesn't fail the build).
2. second commit shows coveralls fails section 1 if new not covered lines are also added.
3. other commits including previous ones show it seems coveralls only decide on covered lines increase or decrease. e.g. maybe coveralls pass a build when you delete some covered functionality and add some new not covered functionality.

PR #229, opened 2018-05-28
WW-4796 Rename springconstants to have a uniform naming pattern

sdutry:


yasserzamani:
If you're sure (how did you?) there are no further constants in core or else where, then looks good to me :+1: thanks!

sdutry:
> If you're sure (how did you?) there are no further constants in core or else where, then looks good to me

I'm not sure i found all of them. 
I just started from SpringConstants and used the `findstr` to find those in there.

lukaszlenart:
I assume we are good to merge this PR if no objections, from my side LGTM 👍

PR #230, opened 2018-06-08
Travis CI build: add 'oraclejdk9'

sullis:


lukaszlenart:
This won't work, Struts 2 won't compile on JDK9

aleksandr-m:
@lukaszlenart What is the problem with compiling with jdk9? Are we tracking this issue somewhere?

lukaszlenart:
It's here https://issues.apache.org/jira/browse/WW-4845

PR #233, opened 2018-06-18
add 'java.rmi.' to struts.excludedPackageNames

sullis:


sullis:
@lukaszlenart

lukaszlenart:
You don't have to mention me in each PR, I get notification when you create them anyway.

How this package can increase application's vulnerability? Do you have an example?

PR #234, opened 2018-06-18
add 'org.codehaus.groovy.' to struts.excludedPackageNames

sullis:


sullis:
@lukaszlenart

yasserzamani:
But Struts hasn't any dependency to `org.codehaus.groovy`.

sullis:
Motivation:  **Secure By Default** philosophy

https://en.wikipedia.org/wiki/Secure_by_default

"Security by default, in software, means that the default configuration settings are the most secure settings possible"

lukaszlenart:
We won't block the whole world and I agree with Yasser in this case. This is also developer/devops responsibility to properly exclude potentially vulnerable classes.

PR #235, opened 2018-06-19
add 'org.apache.commons.io.' to struts.excludedPackageNames

sullis:


sullis:
@lukaszlenart

lukaszlenart:
Do you have a case how to use this package in potential security breach?

sullis:
The blog post uses 'org.apache.commons.io.IOUtils'  to copy data to the servlet OutputStream.

https://www.synopsys.com/blogs/software-security/cve-2017-5638-anatomy-apache-struts-vulnerability/

aleksandr-m:
Data can be copied w/o `IOUtils`.

We cannot exclude everything. What we have right now is enough for most cases and devs can add additional pattern if they want to. Same goes for #233 and #234

lukaszlenart:
In that case (the blog post), `IOUtils` was used just as an example. As Aleksander mentioned, some classes can be used in the same way.

yasserzamani:
Yes and also this package isn't the root cause of the vuln but maybe @sullis meant this is the only class in Struts dependencies which could be used to copy data after a successful attack ?

sullis:
for Struts 2.x applications, [commons-io] is always present on the runtime classpath.   

'org.apache.commons.io.IOUtils' has the potential to be incorporated in an OGNL exploit.

Excluding package [org.apache.commons.io] will reduce the attack surface.

lukaszlenart:
Every class can be incorporated into an OGNL exploit and can be potentially dangerous, blocking everything isn't a way to go. We should rather focus on securely handling incoming values and marking them properly which isn't possible right now with `String`s.

cnenning:
Well, `IOUtils` contain handy utils even for attackers. Blocking it would not increase security but at least it would make attacker's live a little harder 😉

sullis:
Handle incoming values properly ---  yes, absolutely.

First line of defense:         handle incoming values correctly
Second line of defense:    block access to sensitive packages

IMHO, we can do both.

lukaszlenart:
With the first, the second won't be needed. Bear in mind that this impacts an overall application performance. Each and every expression is checked by the `SecurityMemberAccess` and expressions are everywhere inside Struts (e.g. setting actions' properties are expressions, interceptors' configurations are expressions, and so on). No one will use super secure framework which is super slow.

Please treat this as a general comment not related to this `IOUtils` case.

aleksandr-m:
Closing for the reasons above.

PR #240, opened 2018-07-26
fix typos in README

ThrawnCA:


ThrawnCA:
I think there's a problem with the build. All I changed was the README text. The error message suggests that it's trying to use an outdated SSL version to talk to Maven Central maybe?

lukaszlenart:
Thanks, I have fixed the job. Could you touch your PR with a small commit to init the build again?

yasserzamani:
I see SSL has fixed. Merged, Thanks!

PR #244, opened 2018-08-30
WW-4954 Moves XWorkList out of util package

lukaszlenart:
Resolves [WW-4954](https://issues.apache.org/jira/browse/WW-4954)


yasserzamani:
Although we should correctly test if this fix fixes this issue really but unfortunately, as current tests failed to recognize this issue sooner, furthermore, I cannot be sure if any other compatibility issues exists or not.

As I mentioned before, I'm still would hard code these critical exclusions inside pure java code outside of the configuration. I know this has a lot false positives fails on current test but we must fix them rather than testing things not rely on "struts-default.xml".

It would be great if we add these to this PR then we will be able to merge this with a more probability of non-existence of other incompatibility issues :)

aleksandr-m:
LGTM :+1:

yasserzamani:
LGTM :+1: as my locally executed above idea didn't report any further issue :)

aleksandr-m:
@lukaszlenart What about master?

lukaszlenart:
Good point, I will prepare a PR ASAP :)

lukaszlenart:
@aleksandr-m #249

PR #247, opened 2018-09-01
hard code critical exclusions

yasserzamani:
Benefits:
1) Disables user to override them (which puts him/herself on risk).
2) Keeps user safe even if injection mechanism failed temporary at that time.
3) Puts all tests under default exclusions (so discover incompatibilities).

@apache/struts-committers , it toked time but I love and enjoyed this PR. There are huge changes but most of them (maybe 99%) has been done automatically via IntelliJ powerful refactoring.


yasserzamani:
@apache/struts-committers , to facilitate your review: What made me more sure about of lgtm of this PR is: As you can see, I didn't change any test assertion, body or logic. I or better I can say, IntelliJ just moved some classes to a not excluded package and all tests still passing. I strongly like to export those benefits to 2.5.x please. If this PR merged, I will do same for 2.3.x.

aleksandr-m:
> Disables user to override them (which puts him/herself on risk).

Don't treat users as stupid, they are not.

If right now we had it hard-coded then current release would be total fail.
There are valid use cases where excluded classes / packages must be allowed.

> Keeps user safe even if injection mechanism failed temporary at that time.

There are bigger problems at hand if injection fails, probably. :)

> Puts all tests under default exclusions (so discover incompatibilities).

Current tests will still not reveal this issue.
Problem is in the lack of proper integration tests.

Bottom line is: This is not backward compatible and cannot be released in next 2.5.x series.

yasserzamani:
> Don't treat users as stupid, they are not.

I didn't say anything about 'stupid'! AFAIK it's a usability benefit to minimize user mistakes (just think about a developer who in a morning is angry against employer or wife and sees an error about exclusions, he or she simply overrides them I think)

And I remember a user asking at user list that what happens if he removes Object from excluded classes. What we should say? you can but at your risk!

> If right now we had it hard-coded then current release would be total fail.
> There are valid use cases where excluded classes / packages must be allowed.

If I was those users, I was moving those logics into my actions (outside of the excluded packages) instead of removing an exclusion globally.

> There are bigger problems at hand if injection fails, probably. :)

I remember several issues about not working validations which after a lot investigating were because of silently failed injections.

> Current tests will still not reveal this issue.
> Problem is in the lack of proper integration tests.

Because Lukasz already has moved the XWorkList. If you revert back XWorkList, this PR fails with:

```
2018-09-03 13:28:45,626 WARN  [main] ognl.SecurityMemberAccess (SecurityMemberAccess.java:78) - Package of target [[1, 2, 3]] or package of member [public boolean com.opensymphony.xwork2.util.XWorkList.contains(java.lang.Object)] are excluded!
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   XWorkListPropertyAccessorTest.testContains:48 NullPointer
[INFO] 
[ERROR] Tests run: 1794, Failures: 1, Errors: 0, Skipped: 0
```
As you see, if it was hard coded then it was discovering this issue sooner than user :(

> Bottom line is: This is not backward compatible and cannot be released in next 2.5.x series.

This is backward compatible but only for users who has overrided these important exclusions which is not bad to meet them, I think :)

lukaszlenart:
It's rather a matter of using `XWorkTestCase` instead of `StrutsInternalTestCase` - and this is a leftover of merging XWork project into Struts project - XWork and Struts were to separated projects.

aleksandr-m:
Confirming, using `StrutsInternalTestCase` reveals this issue as expected. We should fix the tests. There are about 99 tests which extend `XWorkTestCase`.

yasserzamani:
Thanks for your replies I learnt. So I close this one :)

PR #250, opened 2018-09-06
WW-4948 delete temp files on close instead of on JVM exists

yasserzamani:


lukaszlenart:
Looks good to me 👍 if no objections it can be merged.

yasserzamani:
@apache/struts-committers , I and also [the issue reporter](https://issues.apache.org/jira/browse/WW-4948?page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel) are agree.

lukaszlenart:
No objections, LGTM 👍

lukaszlenart:
Can you port this into `master` branch?

lukaszlenart:
This fix introduced a small problem https://github.com/struts-community-plugins/struts2-jquery/issues/165

sdutry:
minimal project for this new issue:
[https://github.com/sdutry/struts2-staticFiles-devMode](https://github.com/sdutry/struts2-staticFiles-devMode)

Starting up the jetty server ( `mvn jetty:run` ) and going to [localhost:8080/index.action] shows you it fails to load the `styles.css` and `utils.js` resources (originating from the <s:head /> tag).

problem:
loading a static resource with `struts.devMode` set to `true`

It throws a `NullPointerException` in `DefaultStaticContentLoader` because it creates a new instance on each request, but doesn't call setHostConfig other than on the initialy created object => leaving `pathPrefixes` null .

in version `2.5.17` it was the same object and not a new instance

lukaszlenart:
It's because we properly reloads JAR now, but I think this can be improved by initiating `FileManager` in a proper way, right now it will reload everything at least once even if not needed.

lukaszlenart:
Hmm... Yasser's change looks good, I have no idea why this could break anything.
- revision are created only if `reloadConfigs` is set to `true` (in `devMode`)
- then `FileManager` checks if given file need to be reloaded (and this work ok)

I cannot reproduce the issue with my local AppEngine setup.

sdutry:
@yasserzamani 
Does this mean that certain classes should not be reloaded even if reloadingConfigs is true or does this mean that they should be able to and therefore should be fully re-initialised when it happens?

(sorry if it's a dumb question, i've not been active on the project for quite some time, trying to get back into it)

yasserzamani:
@sdutry , you're welcome back thank you very much :) No it's a good question.

I rethought and am thinking about these. According to last thoughts, It seems as currently already behavior, it's nice and OK to load Struts internal configuration i.e. `struts-default.xml` with `reloadingConfigs=false` rather if user has set it or `devMode` to true in his/her app config or not. Until this, Struts is correct and doesn't create a revision for `struts-default.xml`, which I like, because it's not a good practice to monitor it for changes and reload!; It always is not modified even between some Struts versions!

So I think the problem is with [DefaultFileManager.java#L63](https://github.com/apache/struts/blob/6f56e14de93e35b76bcac96207934d3663b4ca57/core/src/main/java/com/opensymphony/xwork2/util/fs/DefaultFileManager.java#L63). When user sets `reloadingConfigs=true` then that line always return true then Struts reloads it's internal `struts-default.xml` which not only is a bad practice, but also even isn't needed as it's not modified.

Kind regards.

lukaszlenart:
but when `reloadConfigs` is `true`, the revision is there [DefaultFileManager.java#L73-L75](https://github.com/apache/struts/blob/6f56e14de93e35b76bcac96207934d3663b4ca57/core/src/main/java/com/opensymphony/xwork2/util/fs/DefaultFileManager.java#L73-L75), so this shouldn't be the case

yasserzamani:
No. If a break point putted there, it's false firstly then after processing user xml, it will be set to true. For struts-default and user struts xml URLs it's false even if user has set it true. So a revision will not created for them. For struts-default xml this is a good behaviour but isn't for user xml and etc.

PR #252, opened 2018-09-19
WW-4963 Implement new Aware interfaces

lukaszlenart:
This PR implements a new set of *Aware interfaces that use `with` instead of `set` to avoid setting values via request.

Implements [WW-4963](https://issues.apache.org/jira/browse/WW-4963)


yasserzamani:
@apache/struts-committers , it's a great work. While there are several changes, LGTM; increased passed test coverage beside Lukasz as author made me more sure ;)

PR #256, opened 2018-10-19
Local s2 2 5 x branch

JCgH4164838Gh792C124B5:
Proposed fix for WW-4971 (broken includes for non-UTF8 content), against 2.5.x branch.
Built successfully with local Maven (2 unrelated test failures on Windows - see below).
Interactive tests indicate the fix is working as intended.

Maven test issue:  EnvsValueSubstitutorTest.testSimpleValue and XmlConfigurationProviderEnvsSubstitutionTest.testSubstitution 
tests fail during Maven build on Windows.  It may be related to assumptions concerning "${env.USER}", ${env.HOME} that may only be true on Linux/Unix.


lukaszlenart:
This PR won't be merged, please use the latest revision of the struts-2-5 branch to prepare a new branch.

PR #257, opened 2018-10-19
Proposed fix for WW-4971 (broken includes for non-UTF8 content).

JCgH4164838Gh792C124B5:
Adds a config flag to make all s:include tags use response page encoding.
Adds an encoding override property to the s:include tag (flexibility).
Adds a warning log output to FastByteArrayBuffer when decoding has errors.

Fixes [WW-4971](https://issues.apache.org/jira/projects/WW/issues/WW-4971)

### DIFF
--- a/core/src/main/java/org/apache/struts2/StrutsConstants.java
+++ b/core/src/main/java/org/apache/struts2/StrutsConstants.java
@@ -41,10 +41,10 @@ public final class StrutsConstants {
     /** The URL extension to use to determine if the request is meant for a Struts action */
     public static final String STRUTS_ACTION_EXTENSION = "struts.action.extension";
 
-	/** Comma separated list of patterns (java.util.regex.Pattern) to be excluded from Struts2-processing */
-	public static final String STRUTS_ACTION_EXCLUDE_PATTERN = "struts.action.excludePattern";
+    /** Comma separated list of patterns (java.util.regex.Pattern) to be excluded from Struts2-processing */
+    public static final String STRUTS_ACTION_EXCLUDE_PATTERN = "struts.action.excludePattern";
 
-    /** Whether to use the alterative syntax for the tags or not */
+    /** Whether to use the alternative syntax for the tags or not */
     public static final String STRUTS_TAG_ALTSYNTAX = "struts.tag.altSyntax";
 
     /** Whether to use the response encoding (JSP page encoding) for s:include tag processing (false - use STRUTS_I18N_ENCODING - by default) */
@@ -59,7 +59,7 @@ public final class StrutsConstants {
     /** The default includeParams method to generate Struts URLs */
     public static final String STRUTS_URL_INCLUDEPARAMS = "struts.url.includeParams";
 
-	public static final String STRUTS_URL_RENDERER = "struts.urlRenderer";
+    public static final String STRUTS_URL_RENDERER = "struts.urlRenderer";
 
     /** The com.opensymphony.xwork2.ObjectFactory implementation class */
     public static final String STRUTS_OBJECTFACTORY = "struts.objectFactory";
@@ -94,7 +94,7 @@ public final class StrutsConstants {
     /** The org.apache.struts2.views.freemarker.FreemarkerManager implementation class */
     public static final String STRUTS_FREEMARKER_MANAGER_CLASSNAME = "struts.freemarker.manager.classname";
 
-    /** Update freemarker templates cache in seconds*/
+    /** Update freemarker templates cache in seconds */
     public static final String STRUTS_FREEMARKER_TEMPLATES_CACHE_UPDATE_DELAY = "struts.freemarker.templatesCache.updateDelay";
     
     /** Cache model instances at BeanWrapper level */
@@ -223,12 +223,12 @@ public final class StrutsConstants {
     public static final String STRUTS_LOCALE_PROVIDER_FACTORY = "struts.localeProviderFactory";
 
     /** The name of the parameter to create when mapping an id (used by some action mappers) */
-	public static final String STRUTS_ID_PARAMETER_NAME = "struts.mapper.idParameterName";
-	
-	/** The name of the parameter to determine whether static method access will be allowed in OGNL expressions or not */
-	public static final String STRUTS_ALLOW_STATIC_METHOD_ACCESS = "struts.ognl.allowStaticMethodAccess";
+    public static final String STRUTS_ID_PARAMETER_NAME = "struts.mapper.idParameterName";
 
-	/** The com.opensymphony.xwork2.validator.ActionValidatorManager implementation class */
+    /** The name of the parameter to determine whether static method access will be allowed in OGNL expressions or not */
+    public static final String STRUTS_ALLOW_STATIC_METHOD_ACCESS = "struts.ognl.allowStaticMethodAccess";
+
+    /** The com.opensymphony.xwork2.validator.ActionValidatorManager implementation class */
     public static final String STRUTS_ACTIONVALIDATORMANAGER = "struts.actionValidatorManager";
 
     /** The {@link com.opensymphony.xwork2.util.ValueStackFactory} implementation class */
@@ -239,7 +239,7 @@ public final class StrutsConstants {
 
     /** The {@link com.opensymphony.xwork2.util.reflection.ReflectionContextFactory} implementation class */
     public static final String STRUTS_REFLECTIONCONTEXTFACTORY = "struts.reflectionContextFactory";
-    
+
     /** The {@link com.opensymphony.xwork2.util.PatternMatcher} implementation class */
     public static final String STRUTS_PATTERNMATCHER = "struts.patternMatcher";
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/struts2/util/FastByteArrayBufferTest.java
```diff
@@ -0,0 +1,238 @@
...
+    String utf8_String = null;
+    String iso8859_1_String = null;
+
+
+    protected void setUp() throws Exception {
+        super.setUp();
+        
+        try {
+            if (utf8_String == null) {
+                utf8_String = new String(utf16_String.getBytes("UTF-8"), "UTF-8");
```
sepe81: You could use `public String(byte bytes[], Charset charset)` with `java.nio.charset.StandardCharsets`.

According to the Javadoc "These charsets are guaranteed to be available on every implementation of the Java platform."

Benefit: No need to handle `UnsupportedEncodingException`.

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -41,12 +41,15 @@
     /** The URL extension to use to determine if the request is meant for a Struts action */
     public static final String STRUTS_ACTION_EXTENSION = "struts.action.extension";
 
-	/** Comma separated list of patterns (java.util.regex.Pattern) to be excluded from Struts2-processing */
```
sepe81: Would it be possible to separate those formatting and "fix typo" changes into one separate commit?

This would simplify the review.

core/src/test/java/org/apache/struts2/util/FastByteArrayBufferTest.java
```diff
@@ -0,0 +1,238 @@
...
+            if (iso8859_1_String == null) {
+                iso8859_1_String = new String(utf16_String.getBytes("ISO-8859-1"), "ISO-8859-1");
+            }
+        }
+        catch(UnsupportedEncodingException ue) {
+            fail("Unable to set up UTF-8 and ISO-8859-1 strings for test: " + ue);
+        }
+    }
+    
+    protected void tearDown() throws Exception {
```
sepe81: no need for empty override of `tearDown`

core/src/test/java/org/apache/struts2/util/FastByteArrayBufferTest.java
```diff
@@ -0,0 +1,238 @@
...
+ */
+public class FastByteArrayBufferTest extends TestCase {
+
+    final String utf16_String = new String("Standard string with accented characters \u00E7\u00E8\u00E9 in the middle");
+    String utf8_String = null;
+    String iso8859_1_String = null;
+
+
+    protected void setUp() throws Exception {
+        super.setUp();
```
sepe81: no need for super call b/c impl in abstract super class is emtpy

core/src/main/java/org/apache/struts2/components/Include.java
```diff
@@ -90,27 +92,50 @@
...
     private static final Logger LOG = LogManager.getLogger(Include.class);
 
-    private static String systemEncoding = System.getProperty("file.encoding");
+    private static final String systemEncoding = System.getProperty("file.encoding");
 
     protected String value;
     private HttpServletRequest req;
     private HttpServletResponse res;
-    private static String defaultEncoding;
+    private String defaultEncoding;   // Made non-static (during WW-4971 fix)
```
sepe81: IMO the comments shouldn't contain to many of those ticket references. Therefore we have the git history.

core/src/main/java/org/apache/struts2/components/Include.java
```diff
@@ -90,27 +92,50 @@
...
+    private String overrideEncoding;  // Added with WW-4971 fix (allow direct encoding override via tag)
+    private boolean usePageEncoding;  // Added with WW-4971 fix (allow usage of page or default encoding)
 
     public Include(ValueStack stack, HttpServletRequest req, HttpServletResponse res) {
         super(stack);
         this.req = req;
         this.res = res;
+        defaultEncoding = "UTF-8";  // Set UTF-8 for defaultEncoding, when not set in configuration
```
aleksandr-m: Why to override `defaultEncoding` in constructor? It will be set in `setDefaultEncoding`.

aleksandr-m:
@JCgH4164838Gh792C124B5  I've run your tests in old version of Struts and all of them succeed. Can you include some test that fails with current version?

core/src/test/java/org/apache/struts2/util/FastByteArrayBufferTest.java
JCgH4164838Gh792C124B5: Fair point, as using the static `StandardCharsets` would allow for immediate assignment in the initializer block.  That would permit elimination of the `setup()` method.

core/src/main/java/org/apache/struts2/StrutsConstants.java
JCgH4164838Gh792C124B5: The "fix typo" changes were minimal enough that I figured it would be OK.  It might be possible to separate those changes out, but I'm pretty new to Git.  Presumably that would involve a new commit/push/pull request sequence (after trying to undo the formatting fixes) ?

core/src/test/java/org/apache/struts2/util/FastByteArrayBufferTest.java
JCgH4164838Gh792C124B5: The `tearDown()` could be eliminated along with `setup()` in relation to your earlier comments.

core/src/main/java/org/apache/struts2/components/Include.java
JCgH4164838Gh792C124B5: I think there are a few spots in the source that refer to ticket references,  After seeing that, it seemed like a good idea to identify the ticket reference for the change via comment.  If the Git history is considered sufficient/preferred those comment references could be eliminated.

core/src/main/java/org/apache/struts2/components/Include.java
JCgH4164838Gh792C124B5: Just a case of being extra defensive (in case `setDefaultEncoding()` injection parameter was ever changed to optional).  It's probably unlikely to occur, in which case that assignment in the constructor could be eliminated.

JCgH4164838Gh792C124B5:
> @JCgH4164838Gh792C124B5 I've run your tests in old version of Struts and all of them succeed. Can you include some test that fails with current version?

Hello @aleksandr-m.  Unfortunately the only way I could find to demonstrate the failure was interactively (with a parent page in ISO-8859-1 using s:include on a child page with ISO-8859-1 high-ASCII accented characters).  The unit tests that were added were primarily to demonstrate the proposed changes to `Include` and `FastByteArrayBuffer`  don't break any existing functionality.

The `FastByteArrayBufferTest.testISO8859_1WriteToUTF8` unit test demonstrates the truncation failure that can occur when input is considered "isMalformed" by the decoder, and causes the log warning to be produced now in that circumstance.  Although the test will pass in older Struts 2.5.x or 2.3.x versions the failure is "silent" in the older versions (can only be noticed by noting the result is truncated visually in the case of an s:include for instance).

Unfortunately I couldn't figure out a mock-test that would pass with the proposed fix, but fail in previous versions.  However I was able to demonstrate it interactively (truncation on older 2.5.x, but success using the appropriate encoding override with the fixed version).  Do you think there's another angle we could use in this scenario ?

core/src/main/java/org/apache/struts2/StrutsConstants.java
sepe81: The following steps will do it.

```
# reset last commit
git reset HEAD~

# see your changes unstaged
git status

# partially add formatting and typo changes
# see for example https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/
git add -p .
git commit -m"Fix formatting and some typos"

# partially add main changes
git add -p .
git commit -m"WW-4971 Fix broken includes for non-UTF8 content"

# check that all wanted changes are included
git status

# log should contain two new commits
git log

# http://weiqingtoh.github.io/force-with-lease/
git push --force-with-lease

```

sepe81: Hint: If anything goes wrong or is unclear, don't do the push and instead repull from guthub.

lukaszlenart: I wrote an instruction how to use GitHub, let me know if anything is unclear and I will be happy to improve it :)
https://struts.apache.org/submitting-patches.html#contributing-with-github

JCgH4164838Gh792C124B5:
Hello @aleksandr-m.  
I've attempted to create a simple reproducer app (WAR) demonstrating this issue using Struts 2.5.18.  For some reason Github doesn't like it as an attachment (even with a .zip extension), so the reproducer was added to the WW-4791 Jira issue instead.
In my case the reproducer demonstrates the issue in this environment: Windows 10 (64-bit), Java 1.7.0_79 (32-bit), Tomcat 7.0.91.  Hopefully it will demonstrate the issue on other environments as well, please let me know how it behaves for you.

core/src/main/java/org/apache/struts2/StrutsConstants.java
JCgH4164838Gh792C124B5: Thank you for the helpful Git instructions, @sepe81 and @lukaszlenart.  The instructions on the Struts website seem to be pretty clear for "normal" contribution usage, and @sepe81 's instructions here seemed to work well for breaking up the changes into two sections (something well beyond my Git skills without such help).  
I ended up doing them in the reverse order he suggested (1st commit was for the proposed fix only, 2nd commit overlays typo fixes on top of that one), so hopefully that will be OK.   :)  :) 
Let me know if it worked out, if not I'll try again (this time around the 1st commit also includes some of the suggested changes from comments, hopefully it's a bit cleaner).

core/src/main/java/org/apache/struts2/components/Include.java
lukaszlenart: Yes, this is a good practice and if you feel it is needed that's ok - but we can also depend on git history - your choice :)

lukaszlenart: Yes, this is a good practice and if you feel it is needed that's ok - but we can also depend on git history - your choice :)

aleksandr-m:
It is the encoding of response that matters. I don't see point in having different encoding in the `s:include`. Is there any?

> The proposed fix attempts to preserve the current 2.5.x behaviour

It is 2.6 we can break some things. :)

lukaszlenart:
Yeah... it would be better to target 2.6 (the `master` branch) instead of 2.5.x

JCgH4164838Gh792C124B5:
Hello again.  
Having an override available in the `s:include` tag provides targeted flexibility (sometimes you can encounter weird encoding/decoding scenarios with source/destination and a localized override can get things to "just work" for such outlier cases).  
As far as 2.5.x vs. 2.6 consideration, applications using 2.5.x won't necessarily be able to move to 2.6.x right away due to various factors.  Since the current (non-patched) behaviour of 2.5.18 (and I believe 2.3.36) for `s:include` isn't working in the identified scenario I think there's value in fixing it in 2.5.x at least.  :)
The proposed `s:include` fix (or variant thereof) could be applied to 2.5.x (2.5.19) and either forward-ported to 2.6 master as-is or as a "stripped-down" version (apply response encoding if available/valid, then fallback to default encoding, then fallback to system encoding).
Thanks for your consideration.

lukaszlenart:
Ok, looks good to me. We can release one more version of 2.5.x :)

sepe81:
@JCgH4164838Gh792C124B5 Would it be possible to include some parts of your reasoning into the commit comment? IMHO this would be the best place for later reference.

@lukaszlenart Not sure how you handle this for the project.

sepe81:
@lukaszlenart I don't mean the cherry-picking. With "include some parts of your reasoning into the commit comment" I intend to enrich the implicit documentation of each commit as to be found under [7. Use the body to explain what and why vs. how](https://chris.beams.io/posts/git-commit).

lukaszlenart:
Ach... you meant JCgH4164838Gh792C124B5's comment - it would be better to put it in the issue on JIRA but it can stay here as well. There is a direct reference in the issue back to this PR.

https://issues.apache.org/jira/browse/WW-4971

aleksandr-m:
Well, we can merge this into 2.5.x branch but that means additional work in 2.6, if we want to make it properly. Deprecation of constant and attribute in tag, etc.

As I see it, it is just a bug that needs to be fixed. Using different encoding than in http response seems to be meaningless and leads to errors.

> sometimes you can encounter weird encoding/decoding scenarios with source/destination and a localized override can get things to "just work" for such outlier cases

@JCgH4164838Gh792C124B5 Can you think of any real world scenarios where it is applicable?

JCgH4164838Gh792C124B5:
Hi @aleksandr-m and @lukaszlenart.
The only definite "real-world" scenario I can provide would be isolating change (only applying an override to single problematic `s:include` minimizes change risk for existing apps to a single JSP).  However since there are concerns about the value/utility of that feature, it could certainly be omitted.
Taking into consideration your comments about how having the tag-based override making things more complicated for inclusion in 2.6, I have a suggestion for a possible way forward:

1) I could attempt another 2-part update (logic change, then typo cleanup) that drops the tag override and only implements the configuration-setting option (`usePageEncoding`) fix for 2.5.x.  As before, it would be false by default (opt-in) to avoid inadvertently impacting/breaking existing 2.5.x apps.
2) The logic in 1), if accepted, could be pulled into 2.6 and then updated to invert the `usePageEncoding` initial state (i.e. have `usePageEncoding` set true by default, but allow a user to revert to defaultEncoding by setting the flag to false in struts.xml configuration).

The above strategy might be sufficient to mitigate the concerns @aleksandr-m outlined, while still providing a relatively safe and flexible solution for both 2.5.x and 2.6 users.  If you think it's reasonable then I will go ahead and attempt the steps outlined in 1) above.  Please let me know what you think.

aleksandr-m:
@JCgH4164838Gh792C124B5 I think that is better. Maybe we should rename `usePageEncoding` to `useResponseEncoding`. Wdyt?

lukaszlenart:
I think we are good to merge this, right?

aleksandr-m:
@lukaszlenart Already done. :) Cherry picked it to master and changed `useResponseEncoding` in `Include` to `true`.

lukaszlenart:
@aleksandr-m osm!!!

PR #258, opened 2018-10-27
[WW-4973] Upgrades to OGNL 3.2.8

lukaszlenart:
Refs [WW-4973](https://issues.apache.org/jira/browse/WW-4973)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -81,8 +81,9 @@ public class SecurityMemberAccess implements MemberAccess {
     @Override
     public boolean isAccessible(Map context, Object target, Member member, String propertyName) {
         LOG.debug("Checking access for [target: {}, member: {}, property: {}]", target, member, propertyName);
-        
-        Class targetClass = target.getClass();
+
+        // target can be null in case of accessing static fields, since OGNL 3.2.8
+        Class targetClass = target != null ? target.getClass() : member.getDeclaringClass();
         Class memberClass = member.getDeclaringClass();
         
         if (checkEnumAccess(target, member)) {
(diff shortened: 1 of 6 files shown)


lukaszlenart:
@apache/struts-committers I opened this PR to review a new approach in OGNL to allow control access to static fields. Till now OGNL didn't support this so you could access any static field. As from OGNL 3.2.8 a check was added if access to the static field is allowed, but there is one thing: in such case `target` is null. It is possible to pass a class as the target but then it will be narrowed to `Class` class (`target.getClass()` will return `Class`).

I assumed it is better to use `null` in this case. Opinions?

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -81,8 +81,9 @@ public void restore(Map context, Object target, Member member, String propertyNa
     @Override
     public boolean isAccessible(Map context, Object target, Member member, String propertyName) {
         LOG.debug("Checking access for [target: {}, member: {}, property: {}]", target, member, propertyName);
-        
-        Class targetClass = target.getClass();
+
+        // target can be null in case of accessing static fields, since OGNL 3.2.8
+        Class targetClass = target != null ? target.getClass() : member.getDeclaringClass();
         Class memberClass = member.getDeclaringClass();
```
sepe81: As a small improvement you could switch L:86 and L:87, declare `Class memberClass` as `final` and reuse `memberClass` within the ternary expression.

lukaszlenart: Thank you, done :)

yasserzamani:
if `target` could be null, I see it also is passed to `ProxyUtil.isProxyMember` which should be fixed also to accept null without exception.

> I assumed it is better to use null in this case. Opinions?

Did you mean in OGNL? because I see Struts doesn't call it directly anywhere so no control to pass null or class as target.

lukaszlenart:
Yes, it is in OGNL itself, when a call to a static field is performed from an expression
https://github.com/jkuhnert/ognl/blob/master/src/main/java/ognl/OgnlRuntime.java#L2066

lukaszlenart:
@yasserzamani I have fixed `ProxyUtil` but to be honest I have no idea how to write a test for that :(

yasserzamani:
It's same as `testAccessStaticField` but with `sma.setDisallowProxyAccess(true)`. i.e. it should return true without encounter an NPE :)

lukaszlenart:
@yasserzamani I have added two more tests but one isn't too realistic ;-)

PR #263, opened 2018-11-12
[WW-4977] Fixing flaky test in Jsr168DispatcherTest and Jsr286DispatcherTest.

OrDTesters:
Issue link: https://issues.apache.org/jira/browse/WW-4977

testRender_ok and testProcessAction_ok in Jsr168DispatcherTest have test-order dependencies and fail when they are run after testModeChangeUsingPortletWidgets. Similarly, testRender_ok and testProcessAction_ok in Jsr286DispatcherTest also have this problem.

These tests can be fixed by adding the entry {STRUTS_ALWAYS_SELECT_FULL_NAMESPACE=“true”} to initParams in testModeChangeUsingPortletWidgets. The fix will enable the tests in these classes to now pass in any order. Let me know if you want to discuss more.


yasserzamani:
Yes you're right. thanks 👍

PR #265, opened 2018-11-13
Update Struts 2.6.0 build with some newer (compatible) library versions.

JCgH4164838Gh792C124B5:
Update Struts 2.6.0 build with some newer (compatible) library versions.

- Add missing multiple-success.jsp for Showcase application.
- Enhance main pom.xml maven-surefire-plugin configuration to avoid some test-ng related duplication failures:  "Cannot find JUnit method class junit.framework.TestSuite$1.warning".
- Remove unnecessary plugins/embeddedjsp/pom.xml (struts2-embeddedjsp-plugin) maven-surefire-plugin version override (noted with preceding change).
- Update Struts 2.6.0 build to use some newer (compatible) library versions.
  Change the main pom.xml library versions for the following:
  - spring.platformVersion 4.3.13.RELEASE -> 4.3.20.RELEASE
  - oval 1.31 -> 1.90  (Note: required unit test fix for OValValidationInterceptorTest.java AND code fix for OvalValidationInterceptor.java) (Additional Note: 1.70 was the most recent that functioned without a fix to OvalValidationInterceptor)
  - jackson 2.9.6 -> 2.9.7
  - fluido-skin.version 1.6 -> 1.7
  - slf4j (slf4j-api, slf4j-simple) 1.7.12 -> 1.7.25
  - xstream 1.4.10 -> 1.4.11.1
  - jetty 6.1.9 -> 6.1.26 (last in 6.1.x line)
  - xerces 2.10.0 - > 2.12.0
  - org.owasp 3.1.1 -> 3.3.4
  - versions-maven-plugin 2.5 -> 2.7
  - doxia-core 1.7 -> 1.8
  - doxia-module-markdown 1.3 -> 1.7
  - org.apache.felix.main 4.0.3 -> 4.6.1  (Note: most recent 4.x)
  - easymock 3.4 -> 3.5.1
  - javax.el 3.0 -> 3.0.1-b10
  - jasper 6.0.18 -> 6.0.53  (Note: most recent 6.0.x)
  - juli 6.0.18 -> 6.0.53    (Note: most recent 6.0.x)
  - commons-logging 1.1.3 -> 1.2
  - commons-collections4 4.1 -> 4.2
  - commons-io 2.5 -> 2.6
  - commons-lang3 3.6 -> 3.8.1
  - commons-text 1.2 -> 1.3  (Note: most recent compatible with Java 7)
  - commons-validator 1.5.1 -> 1.6
  - mockito 1.9.5 -> 1.10.19            (Note: most recent 1.x)
  - cdi-api 1.0-SP1 -> 1.0-SP4          (Note: most recent 1.0.x)
  - weld-core 1.0.1-Final -> 1.0.1-SP4  (Note: most recent 1.0.x)
  - cglib 2.2 -> 2.2.2                  (Note: most recent 2.2.x)
  Note for cglib-nodep:  Build succeeded with cglib-nodep 2.2.2 and 3.2.5.  Although
    cglib 3.2.5 matches ASM 5.2, jmock-cglib 1.2.0 identifies cglib-nodep 2.1_3 as its
    requirement (on the JMock http://jmock.org/download.html download page) so it should
    probably remain as long as JMock 1.2.0 is used.

Fixes [WW-4978](https://issues.apache.org/jira/browse/WW-4978)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,16 +97,16 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <spring.platformVersion>4.3.13.RELEASE</spring.platformVersion>
+        <spring.platformVersion>4.3.20.RELEASE</spring.platformVersion>
         <ognl.version>3.2.8</ognl.version>
         <asm.version>7.0</asm.version>
         <tiles.version>3.0.8</tiles.version>
         <tiles-request.version>1.0.7</tiles-request.version>
         <log4j2.version>2.11.1</log4j2.version>
-        <jackson.version>2.9.6</jackson.version>
+        <jackson.version>2.9.7</jackson.version>
 
         <!-- Site generation -->
-        <fluido-skin.version>1.6</fluido-skin.version>
+        <fluido-skin.version>1.7</fluido-skin.version>
 
         <!-- Sonar -->
         <sonar.host.url>https://builds.apache.org/analysis/</sonar.host.url>
@@ -234,6 +234,13 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.20.1</version>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.apache.maven.surefire</groupId>
+                            <artifactId>surefire-junit47</artifactId>
+                            <version>2.20.1</version>
+                        </dependency>
+                    </dependencies>
                     <configuration>
                         <argLine>-Duser.language=en -Duser.region=US</argLine>
                         <includes>
@@ -242,6 +249,12 @@
                         <excludes>
                             <exclude>**/TestBean.java</exclude>
                         </excludes>
+                        <properties>
+                          <property>
+                            <name>junit</name>
+                            <value>false</value>
+                          </property>
+                        </properties>
                     </configuration>
                 </plugin>
                 <plugin>
@@ -312,7 +325,7 @@
                 <plugin>
                     <groupId>org.owasp</groupId>
                     <artifactId>dependency-check-maven</artifactId>
-                    <version>3.1.1</version>
+                    <version>3.3.4</version>
                     <configuration>
                         <suppressionFiles>
                             <suppressionFile>src/etc/project-suppression.xml</suppressionFile>
@@ -373,12 +386,12 @@
                     <dependency>
                         <groupId>org.apache.maven.doxia</groupId>
                         <artifactId>doxia-core</artifactId>
-                        <version>1.7</version>
+                        <version>1.8</version>
                     </dependency>
                     <dependency>
                         <groupId>org.apache.maven.doxia</groupId>
                         <artifactId>doxia-module-markdown</artifactId>
-                        <version>1.3</version>
+                        <version>1.7</version>
                     </dependency>
                 </dependencies>
             </plugin>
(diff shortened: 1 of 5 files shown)


JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

A Jira WW-4979 for S2 2.6.x has been opened to match this PR (similar to the WW-4978 opened for S2 2.5.x).  It lists the proposed library updates and has some further comments (such as the potential to reduce the number of libraries involved if anyone identifies particular risks).  It also identifies that the struts2-rest-showcase initialization fails, but that it also failed before the library version changes.  If there's anything additional you'd like in the Jira entry please just let me know.

Thanks,
James.

pom.xml
```diff
@@ -1048,7 +1062,8 @@
             <dependency>
                 <groupId>cglib</groupId>
                 <artifactId>cglib</artifactId>
-                <version>2.2</version>
+                <version>2.2.2</version>
```
sepe81: @lukaszlenart With https://github.com/apache/struts/commit/518854a08565c894a18168db7777136c9f3b8f99 we got the update to asm:7 which seems to be compatible with Java 11. 

cglib:3.2.9, according to https://github.com/cglib/cglib/releases/tag/RELEASE_3_2_9, seems to try the same and therefore may be compatible with Java 11.

@JCgH4164838Gh792C124B5 do you see any chance to use the latest version of cglib by the aim of Java 11 support?

JCgH4164838Gh792C124B5: Hello @sepe81 .

I noted that the main pom.xml has two cglib entries (one for cglib-nodep with 2.1_3 - which matches what JMock 1.2.0 indicates as a requirement) and another separate cglib entry (was 2.2, updated to 2.2.2 with this PR).

Interestingly no cglib-x.x.jar is actually packaged and the various "DEPENDENCIES" files in the jars produced by the build only list the 2.1_3 dependency.  I'm still a relative Maven newbie, but is it possible that the non-"nodep" cglib entry is a redundant entry (leftover from a past compile-time dependency) ?

Perhaps the cglib-nodep entry dependency-scope could be set to "test", since it appears to only be needed for jmock-cglib during the test regression series.  That might help with JDK 11 runtime compatibility (if not build compatibility).

Given the above, it seems that migrating to JMock 2.x (current appears to be 2.9.0) might allow upgrading cglib to the latest version (or possibly eliminate the dependency as they mention switching to ByteBuddy for JMock 2.9 - although their release comment mentions "JDK 11 still failing).

Migrating the unit tests to JMock 2.x in the Struts 2.6.x series might be an idea for the future (and should help with a JDK 11 migration goal).

JCgH4164838Gh792C124B5:
Hello @sepe81 .

The inclusion of the multiple-success.jsp JSP in this PR was pretty much incidental, but intended to help verify the Struts2-Showcase was working after building with the new libraries.  

I found that the multiple-success.jsp was referenced-but-missing for the multiple file upload example (and not due to the library changes), but figured it should be implemented for completeness.  Considering it a "test artifact" related to this PR is a bit of a stretch, but made it less likely I'd forget/loose it.    :)

pom.xml
lukaszlenart: I don't think we use CGLIB anywhere in production code, it was used sometime ago in the Convention Plugin (as far I recall)

lukaszlenart:
LGTM 👍

pom.xml
sepe81: Thanks for your investigation. In another step I will try to remove the explicit dependency cglib-nodep. If this works and the tests are fine, I'll create a new PR.

sepe81: see https://github.com/apache/struts/pull/268

PR #266, opened 2018-11-13
Update maven-wrapper to 3.5.4 and add maven-wrapper.jar to .gitignore

sepe81:
The MavenWrapperDownloader downloads the JAR if it is missing locally.

Fixes [WW-4980](https://issues.apache.org/jira/browse/WW-4980)


lukaszlenart:
Could you create a JIRA ticket to cover this change? We use that to automatically generate change log when preparing a release and announcement. Thanks in advance!

lukaszlenart:
LGTM 👍

PR #268, opened 2018-11-14
Remove dependencies cglib and cglib-nodep

sepe81:
* jmock-cglib brings cglib-nodep as a transitive dependency -- no need for explicit declaration
* cglib isn't needed at all -- plugins/portlet build runs perfectly fine without it (`mvn install`)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -289,12 +289,6 @@
         </dependency>
 
         <dependency>
-            <groupId>cglib</groupId>
-            <artifactId>cglib-nodep</artifactId>
-            <optional>true</optional>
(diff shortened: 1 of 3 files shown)


sepe81:
I'd like to see this as a further step for https://issues.apache.org/jira/browse/WW-4979

@lukaszlenart Is it ok to reuse this JIRA reference?

lukaszlenart:
> @lukaszlenart Is it ok to reuse this JIRA reference?

Yeah, we can, this won't impact users

lukaszlenart:
LGTM 👍

PR #269, opened 2018-11-15
Update various plugin dependencies

sepe81:


pom.xml
```diff
@@ -342,6 +342,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
```
sepe81: the explicit version was missing at all

yasserzamani:
Hi Sebastian,

They aren't missing. I remember some common usual maven dependencies like compiler plugin and etc come from parent module, the struts-master module, and it inherits them from it's parent module, the apache-master module. I.e. to change them, we should upgrade apache-master version in struts-master module, then struts-master version in struts root module.

Those dependencies that have versions in strue, it's not a bad idea to check if they are presented already in these parent modules and fixing them.

Kind Regards.
On Nov 16, 2018, at 12:32 AM, Sebastian Peters <notifications@github.com<mailto:notifications@github.com>> wrote:

@sepe81 commented on this pull request.

________________________________

In pom.xml<https://github.com/apache/struts/pull/269#discussion_r234006271>:

> @@ -342,6 +342,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>


the explicit version was missing at all

bundles/pom.xml
```diff
@@ -66,7 +66,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>rat-maven-plugin</artifactId>
-                <version>1.0-alpha-2</version>
+                <version>1.0-alpha-3</version>
```
lukaszlenart: Ups... this is wrong, we are using `apache-parent:21` and the rat plugin is defined there are follow:
```
        <plugin>
          <groupId>org.apache.rat</groupId>
          <artifactId>apache-rat-plugin</artifactId>
          <version>0.12</version>
        </plugin>
```

So it would be better to just redefine this section to:
```
        <plugin>
          <groupId>org.apache.rat</groupId>
          <artifactId>apache-rat-plugin</artifactId>
          ...
```

The same for the pom in the sitegraph plugin

pom.xml
lukaszlenart: 👏

yasserzamani:
Hi Lukasz,


I just afraid if overriding maven compiler version is a good practice or not. Shouldn't we upgrade to most recent version of apache-master and stick together with it? Or it's optional to stick with master? I guess apache-master is here to keep things compatible with each other and overriding them may break it.


Kind Regards.


On Nov 16, 2018, at 10:34 AM, Lukasz Lenart <notifications@github.com<mailto:notifications@github.com>> wrote:

@lukaszlenart requested changes on this pull request.

Looks good, left one comment

________________________________

In bundles/pom.xml<https://github.com/apache/struts/pull/269#discussion_r234106316>:

> @@ -66,7 +66,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>rat-maven-plugin</artifactId>
-                <version>1.0-alpha-2</version>
+                <version>1.0-alpha-3</version>


Ups... this is wrong, we are using apache-parent:21 and the rat plugin is defined there are follow:

        <plugin>
          <groupId>org.apache.rat</groupId>
          <artifactId>apache-rat-plugin</artifactId>
          <version>0.12</version>
        </plugin>


So it would be better to just redefine this section to:

        <plugin>
          <groupId>org.apache.rat</groupId>
          <artifactId>apache-rat-plugin</artifactId>
          ...


The same for the pom in the sitegraph plugin

________________________________

In pom.xml<https://github.com/apache/struts/pull/269#discussion_r234106485>:

> @@ -342,6 +342,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>


👏

lukaszlenart:
Basically those are suggestions, the most important part for us is the release process which heavily depends on apache-master pom - that's the main case we are using it.

sepe81:
> apache-master

sepe81:
@yasserzamani I understand your concerns about versions from parent poms and know this from other projects. But for now I'm unsure how we want to deal with it. Could you please point me to the "apache-master" project -- is it also on GitHub? I only find struts-master on GitHub. If we change any versions there, it has implications for other -- presumably many ;-) -- projects, which I can not overlook. Any ideas?

bundles/pom.xml
sepe81: Oh, I see. the "zero-versions" seem to be newer than 1.x. ;-)

I try it out later.

see http://creadur.apache.org/rat/apache-rat-plugin/changes-report.html

yasserzamani:
It seems we already reference latest apache master module which is configured to use java 7 for source and target. Maybe we should probe what other similar open source projects has done to support java 7,8,9 and 10 in same time?
On Nov 16, 2018, at 12:34 PM, Sebastian Peters <notifications@github.com<mailto:notifications@github.com>> wrote:

@yasserzamani<https://github.com/yasserzamani> I understand your concerns about versions from parent poms and know this from other projects. But for now I'm unsure how we want to deal with it. Could you please point me to the "apache-master" project -- is it also on GitHub? I only find struts-master on GitHub. If we change any versions there, it has implications for other -- presumably many ;-) -- projects, which I can not overlook. Any ideas?

bundles/pom.xml
sepe81: @lukaszlenart I removed the explicit version for `apache-rat-plugin`.

pom.xml
yasserzamani:
```suggestion
```
I think we can and better to delete this one because #270 can pass all tests with all jdks without enforcing this version :)

bundles/pom.xml
```diff
@@ -66,7 +66,6 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>rat-maven-plugin</artifactId>
-                <version>1.0-alpha-2</version>
```
lukaszlenart: It would be better to use this plugin below which is successor 

```
<groupId>org.apache.rat</groupId>
<artifactId>apache-rat-plugin</artifactId>
```

plugins/sitegraph/pom.xml
```diff
@@ -151,7 +151,6 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>rat-maven-plugin</artifactId>
-                <version>1.0-alpha-2</version>
```
lukaszlenart: Same here

```
<groupId>org.apache.rat</groupId>
<artifactId>apache-rat-plugin</artifactId>
```

sepe81: Successfully migrated to apache-rat-plugin.

yasserzamani:
Wow! everything is green :) thanks! LGTM 👍

lukaszlenart:
Nice :) 👍

PR #270, opened 2018-11-16
[WW-4981] Adds support for JDK11 

lukaszlenart:
Refs [WW-4981](https://issues.apache.org/jira/browse/WW-4981)


yasserzamani:
I just pushed some fixes directly into your branch. Please pull them into your local. Now, here at my local, Struts can pass all tests with all java 7,8,9,10 and 11 :)

yasserzamani:
Hmm! FYI travis says green but actually it's java 9 and 11 builds have failed with following error. I'm surprised because at my local all jdks pass all tests!

> goal org.codehaus.mojo:cobertura-maven-plugin:2.7:instrument failed: Plugin org.codehaus.mojo:cobertura-maven-plugin:2.7 or one of its dependencies could not be resolved: Could not find artifact com.sun:tools:jar:0 at specified path /usr/lib/jvm/java-9-oracle/../lib/tools.jar

yasserzamani:
OK. No I was wrong. there are two builds in each travis build. The first one has passed :) The second one `after_success` to report coverage to coveralls has failed. It seems we should disable `after_success` for java 9 and later as coverage maven plugins break with them.

lukaszlenart:
Yes, we must disable that.

yasserzamani:
I disabled already in last commit and verify it. I think your branch is ready for merge now :)

lukaszlenart:
@apache/struts-committers any other objections?

yasserzamani:
Great job Lukasz, thanks a lot! According to your works, do you feel these are importable for 2.5.x branch also?

lukaszlenart:
As far I know, you can run current Struts versions on Java 9 and the same should be possible on Java 11. I would focus on releasing 2.6 and just pushing important fixes into 2.5.x. Basically switch 2.5.x into maintenance mode without new features.

yasserzamani:
I just imported some of these fixes into 2.5.x branch. Works great with all jdks :) I just added these to 2.5.x branch because of users demand on user mail-list , after this one, I also would like switch 2.5.x into maintenance mode without new features 👍  It doesn't seem nice ans is too hard and risky to import everything into 2.5.x branch. Would you like (or should we) to announce these into dev list to inform all including new contributors?

yasserzamani:
Aha after release 👍 I agree. Thanks for clarification.

PR #271, opened 2018-11-17
Back-port enhancement/fix to 2.5.19 unit tests (permits build on Windows)

JCgH4164838Gh792C124B5:
Back-port enhancement/fix from 2.6 to 2.5.x to allow Struts 2.5.x build to complete on Windows.  Without the changes the build fails on Windows due to differences in Linux/Unix vs. Windows environments.

- Involves changes to: EnvsValueSubstitutorTest, XmlConfigurationProviderEnvsSubstitutionTest, xwork-test-envs-substitution.xml.
- Note: 2.5.x EnvValueSubstitutor doesn't support system properties (only env properties) so some tests were not back-ported.

### DIFF
--- a/core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
@@ -20,16 +20,55 @@ package com.opensymphony.xwork2.config.providers;
 
 import org.apache.struts2.StrutsInternalTestCase;
 
-
 public class EnvsValueSubstitutorTest extends StrutsInternalTestCase {
 
+    private boolean osIsWindows = false;  // Assume Linux/Unix environment by default
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+
+        final String os = System.getProperty("os.name");
+        if (os != null && os.startsWith("Windows")) {
+            osIsWindows = true;   // Determined that the OS is Windows (must use different environment variables)
+        }
+        else {
+            osIsWindows = false;  // Assume Linux/Unix environment by default
+        }
+    }
+
     public void testSimpleValue() throws Exception {
+
+        String expected;
+        String actual;
+        final ValueSubstitutor substitutor = new EnvsValueSubstitutor();
+
+        if (osIsWindows) {
+            // given
+            expected = System.getenv("USERNAME");
+
+            // when
+            actual = substitutor.substitute("${env.USERNAME}");
+        }
+        else {
+            // given
+            expected = System.getenv("USER");
+
+            // when
+            actual = substitutor.substitute("${env.USER}");
+        }
+
+        // then
+        assertEquals(expected, actual);
+    }
+
+    public void testDefaultValue() throws Exception {
         // given
-        String expected = System.getenv("USER");
+        String expected = "defaultValue";
         ValueSubstitutor substitutor = new EnvsValueSubstitutor();
 
         // when
-        String actual = substitutor.substitute("${env.USER}");
+        String actual = substitutor.substitute("${env.UNKNOWN:" + expected + "}");
 
         // then
         assertEquals(expected, actual);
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
```diff
@@ -20,16 +20,55 @@
...
-
 public class EnvsValueSubstitutorTest extends StrutsInternalTestCase {
 
+    private boolean osIsWindows = false;  // Assume Linux/Unix environment by default
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+
+        final String os = System.getProperty("os.name");
```
sepe81: We have `commons-lang3` on the classpath -- IMHO it would be better to use `SystemUtils.IS_OS_WINDOWS` for that check as a one-liner at L:25.

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
```diff
@@ -20,16 +20,55 @@
...
+        String actual;
+        final ValueSubstitutor substitutor = new EnvsValueSubstitutor();
+
+        if (osIsWindows) {
+            // given
+            expected = System.getenv("USERNAME");
+
+            // when
+            actual = substitutor.substitute("${env.USERNAME}");
+        }
```
sepe81: I think we should aim for consistent style here and use 

```
if () {
} else {
}
```
instead of

```
if () {
}
else {
}
```

Unfortunately https://cwiki.apache.org/confluence/display/S2WIKI/Struts+2+Coding+Conventions has out-of-date links. But I see this style in most struts2 classes.

@lukaszlenart Are there any up-to-date coding conventions for struts2?

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
JCgH4164838Gh792C124B5: Hello @sepe81 .

This PR copies the equivalent code from previous PR#261 (into 2.6/master), and in both cases I had not considered using additional imports/libraries for the Windows OS check.  If you think using `SystemUtils.IS_OS_WINDOWS` from `commons-lang3` is preferable that is certainly doable (in which case it should probably also be done for the same tests in 2.6/master).  

I'm not sure if it's cleaner to incorporate the fix as-is (to mirror 2.6/master more closely), and then follow-up with an "improve/cleanup" version for 2.5.x and master branches.

Please let me know what you think.  :)

sepe81: Hey @JCgH4164838Gh792C124B5 

you are right. In this case it's just kind of cherry-picking. Better we try some improvement exclusively for the master branch. :+1:

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
JCgH4164838Gh792C124B5: Hello @sepe81 .

Maybe an if/else coding-conventions cleanup of these two tests would be better served with a pair of follow-up PRs that apply the convention to the two tests (along with the `commons-lang3` enhancement you identified above) both the 2.5.x and master branches ?

Yourself and @lukaszlenart can advise which convention is preferred and I'll do my best to remember.  For whatever reason, I seem to default to lining up the if-else elements column-wise, even if the { } positioning varies.  :)  :)

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
sepe81: see https://github.com/apache/struts/pull/272

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
sepe81: see https://github.com/apache/struts/pull/272

lukaszlenart:
👏

core/src/test/java/com/opensymphony/xwork2/config/providers/EnvsValueSubstitutorTest.java
yasserzamani: 👍 I would like an automatic style checker tool like [apache/commons-lang](https://github.com/apache/commons-lang/blob/a668a4798163cbf2ca33584270a439369d33fedc/pom.xml#L620) but I'm not sure if any tool exists that supports checking only PR added code lines because it's hard to fix all project code styles.

PR #276, opened 2018-11-19
WW-4983 Set private access modifier for HttpParameters.toMap

sepe81:
This method can be potentially dangerous as it allows access to raw
parameter values.


JCgH4164838Gh792C124B5:
👍

lukaszlenart:
LGTM 👍

PR #279, opened 2018-11-20
Update persistence-api from 1.0 to 1.0.2

sepe81:


lukaszlenart:
I wonder if we should create a JIRA ticket for such a minor change 🤔

sepe81:
Fixes https://issues.apache.org/jira/browse/WW-4985

PR #280, opened 2018-11-21
Enhance after_success for Travis CI

sepe81:
Use default bash if-then-else block with semicolon after each command and fix spelling for "incompatibility".

Don't do `mvn clean` b/c we need the test results for the following tasks and don't need to rebuild everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,10 @@ script: mvn test -DskipAssembly
 
 after_success:
   # TODO delete following if statement after fix of https://github.com/cobertura/cobertura/issues/271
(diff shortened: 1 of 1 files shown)


sepe81:
@lukaszlenart the api key for UpdateImpact seems to be ignored. At the end of the relevant Travis CI build I see `Please define the api key. You can find it on UpdateImpact.com`.

lukaszlenart:
@sepe81 we can ignore the UpdateImpact plugin, I mostly added it to have a badge with number of dependencies ;-)

yasserzamani:
osm! about two minutes save in jdk8 job, thanks a lot!

PR #283, opened 2018-11-26
Upgrade DWR from 1.x to 2.x

JCgH4164838Gh792C124B5:
Upgrade `DWR` from 1.x to 2.x
- intended to resolve Jenkins build failure with old DWR version.

`pom.xml` changes:
- entry change in `struts2-parent pom.xml`
  - upgrade 1.1.1 (uk.ltd.getahead) -> 2.0.11-RELEASE (org.directwebremoting)
- entry change in `struts2-dwr-plugin pom.xml`
- entry change in `struts2-showcase pom.xml`

### DIFF
--- a/apps/showcase/pom.xml
+++ b/apps/showcase/pom.xml
@@ -133,7 +133,7 @@
             <artifactId>sitemesh</artifactId>
         </dependency>
         <dependency>
-            <groupId>uk.ltd.getahead</groupId>
+            <groupId>org.directwebremoting</groupId>
             <artifactId>dwr</artifactId>
         </dependency>
         <dependency>
--- a/plugins/dwr/pom.xml
+++ b/plugins/dwr/pom.xml
@@ -33,7 +33,7 @@
 
     <dependencies>
         <dependency>
-            <groupId>uk.ltd.getahead</groupId>
+            <groupId>org.directwebremoting</groupId>
             <artifactId>dwr</artifactId>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -648,9 +648,9 @@
             </dependency>
 
             <dependency>
-                <groupId>uk.ltd.getahead</groupId>
+                <groupId>org.directwebremoting</groupId>
                 <artifactId>dwr</artifactId>
-                <version>1.1.1</version>
+                <version>2.0.11-RELEASE</version>
             </dependency>
 
             <dependency>


yasserzamani:
Unfortunately this plugin has [zero test coverage](https://coveralls.io/builds/20283669/source?filename=plugins/dwr/src/main/java/org/apache/struts2/validators/DWRValidator.java) and so we cannot verify if this upgrade breaks anything or not :( there are 8 years between those versions!

lukaszlenart:
We can use the Showcase app to manually test if it works or not.

yasserzamani:
How do you test `dwr` functionality via `showcase`? I'm not able to find related ui in main page. I only found `http://localhost:8080/tags/ui/showDynamicTreeAction.action` which truly fails with `HTTP Status 404 - /WEB-INF/tags/ui/treeExampleDynamic.jsp`.

Anyway IMHO it seems undeniable to add a minimal main scenario covering unit test to this PR when we found some time.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

I know next-to-nothing about `DWR`, but from the overview page it mentions browser/web-container interaction with `AJAX`.  Not sure how an automated test might work for it ...

As for the `ShowDynamicTreeAction` in the showcase there's apparently no way to "mouse-navigate" to it.
Looks like the `treeExampleDynamic.jsp` is missing too (both from the .war and the app source) - which explains the `404 return`.

In terms of `DWR`, the `/dwr/*` servlet mapping should be in effect (based on configuration).
After re-checking, it seems the DWR example is interfered-with by one of the many `/*` `filter-mappings` defined in the showcase `web.xml`.

By _commenting out_ all of the `/*` `filter-mappings` in the showcase `web.xml` I was able to
navigate to: `http://localhost:8080/struts2-showcase/dwr/index.html`.  That displays a page with
a `validator` link which when clicked launches a test page of sorts.

Don't know how the `DWR` validator page is supposed to work, but _it seems to look/behave similarly
for both the 2.6-snapshot and the 2.5.19-snapshot_ ...

What do you think ?

lukaszlenart:
It looks good to me, just a reference to a JIRA ticket is missing :) Thank you!

JCgH4164838Gh792C124B5:
A `JIRA` ticket (https://issues.apache.org/jira/browse/WW-4988) has been created which references back to this PR, as requested.  :)

yasserzamani:
Thank you! yesterday I tried to add an unit-test but was so complicated that I thought. It was needing mocking servletConfig, servletContext, request, response and etc for DWRServlet and in same time, to be able to execute a validation aware action outside of the core tests. So let's keep showcase manual test enough and merge :)

PR #285, opened 2018-12-03
WW-4069 upgrade to DWR 3 with a showcase

yasserzamani:


plugins/dwr/src/main/java/org/apache/struts2/validators/DWRValidator.java
```diff
@@ -32,7 +32,7 @@
 import org.apache.struts2.dispatcher.RequestMap;
 import org.apache.struts2.dispatcher.SessionMap;
 
-import uk.ltd.getahead.dwr.WebContextFactory;
+import org.directwebremoting.WebContextFactory;
```
lukaszlenart: It was that easy??? wow :)

pom.xml
```diff
@@ -650,7 +650,7 @@
             <dependency>
                 <groupId>org.directwebremoting</groupId>
                 <artifactId>dwr</artifactId>
-                <version>2.0.11-RELEASE</version>
+                <version>3.0.2-RELEASE</version>
```
lukaszlenart: 👏

plugins/dwr/src/main/java/org/apache/struts2/validators/DWRValidator.java
yasserzamani: Fortunately yes, I enjoyed it too :)

lukaszlenart:
LGTM 👍

PR #286, opened 2018-12-04
WW-4987 Add list attributes to optgroup tag

aleksandr-m:


core/src/main/resources/template/simple/optgroup.ftl
```diff
@@ -35,10 +35,37 @@
 	<#assign tmpKey=stack.findValue(optGroupInternalListUiBean.parameters.listKey) />
 	<#assign tmpValue=stack.findValue(optGroupInternalListUiBean.parameters.listValue) />
 	<#assign tmpKeyStr = tmpKey.toString() />
+	<#assign optGroupItemCssClass = ''/>
+	<#if optGroupInternalListUiBean.parameters.listCssClass??>
+		<#if stack.findString(optGroupInternalListUiBean.parameters.listCssClass)??>
+			<#assign optGroupItemCssClass= stack.findString(optGroupInternalListUiBean.parameters.listCssClass)/>
```
lukaszlenart: Such repeatit evaluation can be expensive, why not to do it once and assign the value? In case of missing expression, a `null` will be returned which should work with `?has_content`.

core/src/main/resources/template/simple/optgroup.ftl
```diff
@@ -35,10 +35,37 @@
...
+	<#if optGroupInternalListUiBean.parameters.listTitle??>
+		<#if stack.findString(optGroupInternalListUiBean.parameters.listTitle)??>
+			<#assign optGroupItemTitle= stack.findString(optGroupInternalListUiBean.parameters.listTitle)/>
+		</#if>
+	</#if>
 	<option value="${tmpKeyStr?html}"<#rt>
 	<#if tag.contains(parameters.nameValue, tmpKey) == true>
 	selected="selected"<#rt>
 	</#if>
+	<#if optGroupItemCssClass?has_content>
```
lukaszlenart: hm... `?has_content` is pretty much the same as `??` but it supports collections, do we expect collections here? Or did I miss something?

core/src/main/resources/template/simple/optgroup.ftl
aleksandr-m: Copy paste from select template actually. I've seen this and thought about improving it later in separate commit. :)
Freemarker doesn't allow to assign null / undefined, so the default value operator should be used.

core/src/main/resources/template/simple/optgroup.ftl
aleksandr-m: It will be assigned empty string in case there is no attribute in the tag and `??` doens't check for empty strings.

aleksandr-m:
@lukaszlenart Thanks. I've removed multiple calls to `findString` and improved test a bit.

core/src/main/resources/template/simple/optgroup.ftl
lukaszlenart: 👏

core/src/main/resources/template/simple/optgroup.ftl
lukaszlenart: 👍

lukaszlenart:
@aleksandr-m thanks a lot! I have restarted Travis and waiting for sign off :)

lukaszlenart:
LGTM 👍

PR #287, opened 2018-12-06
[WW-4889] Adds XML handler based on Apache Juneau

lukaszlenart:
Implements [WW-4889](https://issues.apache.org/jira/browse/WW-4889)

### DIFF
--- a/plugins/rest/src/test/java/org/apache/struts2/rest/handler/JuneauXmlHandlerTest.java
+++ b/plugins/rest/src/test/java/org/apache/struts2/rest/handler/JuneauXmlHandlerTest.java
@@ -82,8 +82,8 @@ public class JuneauXmlHandlerTest extends XWorkTestCase {
 
         // then
         assertNotNull(obj);
-        assertEquals(obj.getName(), "Jan");
-        assertEquals(obj.getAge().longValue(), 12L);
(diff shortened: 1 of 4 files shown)


plugins/rest/src/test/java/org/apache/struts2/rest/handler/JuneauXmlHandlerTest.java
```diff
@@ -0,0 +1,93 @@
...
+        // given
+        SimpleBean obj = new SimpleBean();
+
+        // when
+        Reader in = new StringReader(xml);
+        handler.toObject(ai, in, obj);
+
+        // then
+        assertNotNull(obj);
+        assertEquals(obj.getName(), "Jan");
```
sepe81: The params order for `assertEquals` should be expected, actual.

lukaszlenart:
Just an alternative, I rather thought about switching to JacksonXML as a default to handle XMLs - it's already there.

yasserzamani:
A test is failing. It seems generated xml elements order is different than your local one pc.

lukaszlenart:
@yasserzamani thanks, fixed :)

yasserzamani:
LGTM 👍

PR #289, opened 2018-12-12
[WW-4972] Switches incompatible_improvements to VERSION_2_3_28

lukaszlenart:
Implements [WW-4972](https://issues.apache.org/jira/browse/WW-4972)

### DIFF
--- a/apps/showcase/src/main/resources/template/ajaxErrorContainers/actionerror.ftl
+++ b/apps/showcase/src/main/resources/template/ajaxErrorContainers/actionerror.ftl
@@ -23,21 +23,21 @@
 -->
 <ul<#rt/>
 <#if parameters.id??>
- id="${parameters.id?html}"<#rt/>
+ id="${parameters.id}"<#rt/>
 </#if>            
 <#if parameters.cssClass??>
- class="${parameters.cssClass?html}"<#rt/>
+ class="${parameters.cssClass}"<#rt/>
 <#else>
  class="errorMessage"<#rt/>
 </#if>
 <#if parameters.cssStyle??>
- style="${parameters.cssStyle?html}"<#rt/>
+ style="${parameters.cssStyle}"<#rt/>
 </#if>
 >
 <#if (actionErrors?? && actionErrors?size > 0)>
 	<#list actionErrors as error>
 		<#if error??>
-            <li><span><#if parameters.escape>${error!?html}<#else>${error!}</#if></span><#rt/></li><#rt/>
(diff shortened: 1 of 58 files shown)


yasserzamani:
Travis jdk7 fails on updateimpact submit - I have no vision on updateimpact

aleksandr-m:
@lukaszlenart What about `?html` in other files?

lukaszlenart:
@aleksandr-m to be honest I'm not sure if this is a proper solution at all... for other side, those are only JS related attributes.

aleksandr-m:
@lukaszlenart Me neither. Assuming everything works as it is now, maybe we can postpone removing of `?html` until it is removed from freemarker?

lukaszlenart:
It's rather an issue with changed behaviour of `?html` - when I leave it, all single quotes will be escaped with a proper entity. 

And now I'm reading about `?html` and it was [marked as deprecated](https://freemarker.apache.org/docs/ref_builtins_string.html#ref_builtin_html), I think I will extend this PR and will throw out all those in return using [auto-escaping](https://freemarker.apache.org/docs/dgui_misc_autoescaping.html)

lukaszlenart:
@aleksandr-m I have dropped all the `?html` and switched into auto escaping mode by default. I was wonder if instead of hardcoding auto escape into `FreemarkerManager` it would be better to rename all `.ftl` files into `.ftlh` - thus enables HTML auto escaping by a file type.

aleksandr-m:
@lukaszlenart I think it is ok. Renaming templates means that users also must rename overridden templates in their apps. It is not a big deal, but still one more thing to do upon upgrading. Besides each template can change these settings with `<#ftl output_format=...>` and `<#ftl auto_esc=...>`.

lukaszlenart:
I just figured out that I should also review all the JS related attributes to be sure to use
```
<#outputformat 'JavaScript'>${parameters....}</#outputformat>
```
instead of enabling JS escaping for the whole template.

lukaszlenart:
@apache/struts-committers this is ready for another round

yasserzamani:
To be frank I didn't pursue due to lack of knowledge about `freemarker backward compatibility` concept e.g. fall back to latest version doesn't make sense for me, it looks like an upgrade for me. So I thought let @aleksandr-m to handle this please :) or if you prefer additional my review, it's OK also :)

lukaszlenart:
Yes, this an upgrade but thanks to `incompatible improvements` it was possible to do those changes in 2.6 instead of doing them in 2.5.x or something :)

PR #291, opened 2018-12-13
Updates and cleanup for some configuration-related items (2.5.x).

JCgH4164838Gh792C124B5:
Updates and cleanup for some configuration-related items.
- Made several attributes final.
- Added limiter logic to the setting of certain configuration elements, and increased logging around them.
- Updated unit tests impacted by the limiter logic, including minor typo fixes and locale consistency.
- Minor cleanup while making the changes.

PR #292, opened 2018-12-13
Updates and cleanup for some configuration-related items (2.6.x)

JCgH4164838Gh792C124B5:
Updates and cleanup for some configuration-related items.
- Made several attributes final.
- Added limiter logic to the setting of certain configuration elements, and increased logging around them.
- Updated unit tests impacted by the limiter logic, including minor typo fixes and locale consistency.
- Minor cleanup while making the changes.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
@@ -224,6 +224,7 @@ public class XWorkConfigurationProvider implements ConfigurationProvider {
         props.setProperty(StrutsConstants.STRUTS_ENABLE_OGNL_EXPRESSION_CACHE, Boolean.TRUE.toString());
         props.setProperty(StrutsConstants.STRUTS_ENABLE_OGNL_EVAL_EXPRESSION, Boolean.FALSE.toString());
         props.setProperty(StrutsConstants.STRUTS_CONFIGURATION_XML_RELOAD, Boolean.FALSE.toString());
+        props.setProperty(StrutsConstants.STRUTS_ALLOW_STATIC_METHOD_ACCESS, Boolean.FALSE.toString());
     }
 
 }
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -57,21 +57,17 @@ public class OgnlUtil {
     private final ConcurrentMap<Class, BeanInfo> beanInfoCache = new ConcurrentHashMap<>();
     private TypeConverter defaultConverter;
 
-    private boolean devMode;                       // Defaults to false
-    private boolean devModeValueSet;               // Defaults to false (not explicitly set yet)
+    private boolean devMode;
     private boolean enableExpressionCache = true;
-    private boolean enableEvalExpression;          // Defaults to false
-    private boolean enableEvalExpressionValueSet;  // Defaults to false (not explicitly set yet)
+    private boolean enableEvalExpression;
 
     private Set<Class<?>> excludedClasses;
     private Set<Pattern> excludedPackageNamePatterns;
     private Set<String> excludedPackageNames;
 
     private Container container;
-    private boolean allowStaticMethodAccess;            // Defaults to false
-    private boolean allowStaticMethodAccessValueSet;    // Defaults to false (not explicitly set yet)
-    private boolean disallowProxyMemberAccess;          // Defaults to false
-    private boolean disallowProxyMemberAccessValueSet;  // Defaults to false (not explicitly set yet)
+    private boolean allowStaticMethodAccess;
+    private boolean disallowProxyMemberAccess;
 
     public OgnlUtil() {
         excludedClasses = new HashSet<>();
@@ -80,49 +76,35 @@ public class OgnlUtil {
     }
 
     @Inject
-    public void setXWorkConverter(XWorkConverter conv) {
+    protected void setXWorkConverter(XWorkConverter conv) {
         this.defaultConverter = new OgnlTypeConverterWrapper(conv);
     }
 
     @Inject(StrutsConstants.STRUTS_DEVMODE)
-    public void setDevMode(String mode) {
-        final boolean booleanMode = BooleanUtils.toBoolean(mode);
-        if (!this.devModeValueSet) {
-            this.devMode = booleanMode;
-            this.devModeValueSet = true;  // Permit setting devMode only once
-            if (this.devMode) {
-                LOG.warn("Setting development mode [{}] affects the safety of your application!",
-                            this.devMode);
-            }
-        }
-        else {
-            LOG.debug("Error setting development mode value [{}], already previously set to [{}]", mode, this.devMode);
+    protected void setDevMode(String mode) {
+        this.devMode = BooleanUtils.toBoolean(mode);
+        if (this.devMode) {
+            LOG.warn("Setting development mode [{}] affects the safety of your application!",
+                        this.devMode);
         }
     }
 
     @Inject(StrutsConstants.STRUTS_ENABLE_OGNL_EXPRESSION_CACHE)
-    public void setEnableExpressionCache(String cache) {
+    protected void setEnableExpressionCache(String cache) {
         enableExpressionCache = BooleanUtils.toBoolean(cache);
     }
 
     @Inject(value = StrutsConstants.STRUTS_ENABLE_OGNL_EVAL_EXPRESSION, required = false)
-    public void setEnableEvalExpression(String evalExpression) {
-        if (!this.enableEvalExpressionValueSet) {
-            this.enableEvalExpression = "true".equals(evalExpression);
-            this.enableEvalExpressionValueSet = true;  // Permit setting enableEvalExpression only once
-            if (this.enableEvalExpression) {
-                LOG.warn("Enabling OGNL expression evaluation may introduce security risks " +
-                        "(see http://struts.apache.org/release/2.3.x/docs/s2-013.html for further details)");
-            }
-        }
-        else {
-            LOG.debug("Error setting enable OGNL expression evaluation value [{}], already previously set to [{}]",
-                    evalExpression, this.enableEvalExpression);
+    protected void setEnableEvalExpression(String evalExpression) {
+        this.enableEvalExpression = BooleanUtils.toBoolean(evalExpression);
+        if (this.enableEvalExpression) {
+            LOG.warn("Enabling OGNL expression evaluation may introduce security risks " +
+                    "(see http://struts.apache.org/release/2.3.x/docs/s2-013.html for further details)");
         }
     }
 
     @Inject(value = StrutsConstants.STRUTS_EXCLUDED_CLASSES, required = false)
-    public void setExcludedClasses(String commaDelimitedClasses) {
+    protected void setExcludedClasses(String commaDelimitedClasses) {
         Set<Class<?>> excludedClasses = new HashSet<>();
         excludedClasses.addAll(this.excludedClasses);
         excludedClasses.addAll(parseExcludedClasses(commaDelimitedClasses));
@@ -145,7 +127,7 @@ public class OgnlUtil {
     }
 
     @Inject(value = StrutsConstants.STRUTS_EXCLUDED_PACKAGE_NAME_PATTERNS, required = false)
-    public void setExcludedPackageNamePatterns(String commaDelimitedPackagePatterns) {
+    protected void setExcludedPackageNamePatterns(String commaDelimitedPackagePatterns) {
         Set<Pattern> excludedPackageNamePatterns = new HashSet<>();
         excludedPackageNamePatterns.addAll(this.excludedPackageNamePatterns);
         excludedPackageNamePatterns.addAll(parseExcludedPackageNamePatterns(commaDelimitedPackagePatterns));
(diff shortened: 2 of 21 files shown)


core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
```diff
@@ -224,6 +224,7 @@ public void register(ContainerBuilder builder, LocatableProperties props)
         props.setProperty(StrutsConstants.STRUTS_ENABLE_OGNL_EXPRESSION_CACHE, Boolean.TRUE.toString());
         props.setProperty(StrutsConstants.STRUTS_ENABLE_OGNL_EVAL_EXPRESSION, Boolean.FALSE.toString());
         props.setProperty(StrutsConstants.STRUTS_CONFIGURATION_XML_RELOAD, Boolean.FALSE.toString());
+        props.setProperty(StrutsConstants.STRUTS_ALLOW_STATIC_METHOD_ACCESS, Boolean.FALSE.toString());
```
aleksandr-m: Why? Is it not working?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -82,7 +86,18 @@ public void setXWorkConverter(XWorkConverter conv) {
 
     @Inject(StrutsConstants.STRUTS_DEVMODE)
     public void setDevMode(String mode) {
-        this.devMode = BooleanUtils.toBoolean(mode);
+        final boolean booleanMode = BooleanUtils.toBoolean(mode);
```
aleksandr-m: Why? If I remember correctly then injection can happen multiple times and some plugins, for example, can override already injected properties. That is how some things work (`disallowProxyMemberAccess` is one of this things).

core/src/main/resources/struts-default.xml
```diff
@@ -61,6 +61,7 @@
                 freemarker.ext.rhino.,
                 sun.reflect.,
                 javassist.,
+                org.objectweb.asm.,
```
aleksandr-m: This must be in separate commit / PR.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlUtilTest.java
```diff
@@ -375,7 +381,7 @@ public void testSetPropertiesDate() {
         assertEquals(cal.getTime(), foo.getBirthday());
 
         //UK style test
-        cal = Calendar.getInstance();
+        cal = Calendar.getInstance(Locale.UK);
```
aleksandr-m: Why?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -57,17 +57,21 @@
...
+    private boolean enableEvalExpressionValueSet;  // Defaults to false (not explicitly set yet)
 
     private Set<Class<?>> excludedClasses;
     private Set<Pattern> excludedPackageNamePatterns;
     private Set<String> excludedPackageNames;
 
     private Container container;
-    private boolean allowStaticMethodAccess;
-    private boolean disallowProxyMemberAccess;
+    private boolean allowStaticMethodAccess;            // Defaults to false
```
aleksandr-m: No need for the comments here.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
yasserzamani: @apache/struts-committers , I think here @JCgH4164838Gh792C124B5 is trying to prevent hacker from the ability of resetting those values that menaces overall security. Whether it works or breaks here, In general it seems a nice devised practice :+1: thank you!

aleksandr-m: @yasserzamani How it is helpful if basic stuff isn't working?

If this is really the case then they can be protected in some other way.

yasserzamani: Yeah basic stuff should work. I actually meant the idea itself - to prevent hacker from the ability of resetting those values that menaces overall security.

core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
JCgH4164838Gh792C124B5: Hello @aleksandr-m .
Are you asking why the new entry on line 227 ?  Simply to ensure the props includes a set property for that element, in line with the other significant ones already present.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
JCgH4164838Gh792C124B5: Hello Struts Team.

The primary intention for this PR was only to try and limit the ability to change certain elements (such as devMode), for which there didn't seem to be a good reason for them to change after the 1st (initial) configuration.  

It was more aimed at avoiding inadvertent/accidental state changes for important state than anything else.

None of the regression tests, nor interaction with the Showcase application encountered any issues so it seemed like a reasonable approach to take.

If there are certain properties (such as `disallowProxyMemberAccess`) for which subsequent changes are necessary/valid, then they could be omitted by this or another mechanism that attempts to limit such changes.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlUtilTest.java
JCgH4164838Gh792C124B5: Hello @aleksandr-m .

I found during debugging within an IDE that the `OgnlUtilTest` test `testSetPropertiesDate` was failing (even though it worked when run in the Maven build - on Windows in both cases).  It might be a combination of OS/environment conditions for the default Java locale - so I explicitly set the locale values (based on the existing comments) so that the test worked consistently both inside and outside the Maven build.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
JCgH4164838Gh792C124B5: Hello @aleksandr-m  and @yasserzamani .  

Do either of you think _any_ of the properties I attempted to apply change-limiters to could/should have the change-limiter applied to them ?

I could eliminate some or all of the change-limiters in a revision of this PR (and PR #291) that only retains the additional logging around property changes.  Please let me know what you think.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
aleksandr-m: Don't be. :) Unnecessary comments just clutter code and can take focus away from the important stuff.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlUtilTest.java
aleksandr-m: Ok.

core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
aleksandr-m: But there is more constants. Why to add especially this one?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
aleksandr-m: > Do either of you think any of the properties I attempted to apply change-limiters to could/should have the change-limiter applied to them ?

Well, currently I don't see what benefits does it bring.

core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
JCgH4164838Gh792C124B5: Properties like devMode, DMI, eval_expression are all "privilege-related" (and set to false), just like static method access.  Since the others were present, it seemed that static method access should be there too (for completeness, also set to false).
That was my reasoning anyway.  :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
yasserzamani: Hi @JCgH4164838Gh792C124B5 , I like your idea but to be frank, I'm not able to have a vision on future side effects.

But I remembered one of my old ideas that not only has somehow same benefits but also is lower risky and is more readable: **Refactoring all methods annotated with `@com.opensymphony.xwork2.inject.Inject` from `public setXXX` to `private injectXXX`.** This preserves those secure against OGNL access. As a glimpse, with IntelliJ IDE's refactoring tools, I did it on `OgnlUtil.setDevMode` and worked  :)

core/src/main/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsChecker.java
```diff
@@ -66,9 +76,14 @@ public void setAdditionalExcludePatterns(String excludePatterns) {
     @Inject(StrutsConstants.STRUTS_ENABLE_DYNAMIC_METHOD_INVOCATION)
     public void setDynamicMethodInvocation(String dmiValue) {
         if (!BooleanUtils.toBoolean(dmiValue)) {
-            LOG.debug("DMI is disabled, adding DMI related excluded patterns");
+            if (LOG.isDebugEnabled()) {
```
yasserzamani: Please revise all `if (LOG...` statements in this PR according to [this answer](https://stackoverflow.com/a/29561054/1362623).

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -92,10 +107,17 @@ public void setEnableExpressionCache(String cache) {
 
     @Inject(value = StrutsConstants.STRUTS_ENABLE_OGNL_EVAL_EXPRESSION, required = false)
     public void setEnableEvalExpression(String evalExpression) {
-        enableEvalExpression = "true".equals(evalExpression);
-        if(enableEvalExpression){
-            LOG.warn("Enabling OGNL expression evaluation may introduce security risks " +
-                    "(see http://struts.apache.org/release/2.3.x/docs/s2-013.html for further details)");
+        if (!this.enableEvalExpressionValueSet) {
```
yasserzamani: Sorry you did a lot of work but please replace all change-limiters logic including their unit test with proposed solution above. Thanks in advance!

JCgH4164838Gh792C124B5:
Hello @yasserzamani  and @aleksandr-m .

This PR has been updated with changes that hopefully addresses most of the concerns that were provided in your feedback.

The limiter logic was removed and instead of the full-refactoring of the injectable methods it only reduces the access level of most of the relevant methods to `protected` (improves access safety with less chance of impacting plugins or other elements negatively).  It's not the level of refactoring that @yasserzamani  suggested above, but is _a small step in that direction_ (although slightly less risky in terms of potential impacts).

Let me know if the changes in the PR look better now.

core/src/main/java/com/opensymphony/xwork2/security/DefaultAcceptedPatternsChecker.java
```diff
@@ -70,7 +70,14 @@ public void setAcceptedPatterns(String[] additionalPatterns) {
     }
 
     public void setAcceptedPatterns(Set<String> patterns) {
-        LOG.trace("Sets accepted patterns [{}]", patterns);
+        if (acceptedPatterns == null) {
+            LOG.debug("Replacing accepted patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
```
yasserzamani: It should be `LOG.debug("Sets accepted patterns [{}]", patterns);`

core/src/main/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsChecker.java
```diff
@@ -46,9 +46,15 @@ public DefaultExcludedPatternsChecker() {
...
-    public void setOverrideExcludePatterns(String excludePatterns) {
-        LOG.warn("Overriding excluded patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
-                    StrutsConstants.STRUTS_OVERRIDE_EXCLUDED_PATTERNS, excludePatterns);
+    protected void setOverrideExcludePatterns(String excludePatterns) {
+        if (excludedPatterns != null && excludedPatterns.size() > 0) {
+            LOG.warn("Overriding excluded patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
+                        excludedPatterns, excludePatterns);
+        }
+        else {
+            LOG.debug("Overriding excluded patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
```
yasserzamani: It should be `LOG.debug("Overriding excluded patterns with [{}]", excludePatterns);`

core/src/main/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsChecker.java
```diff
@@ -80,7 +89,14 @@ public void setExcludedPatterns(String[] patterns) {
     }
 
     public void setExcludedPatterns(Set<String> patterns) {
-        LOG.trace("Sets excluded patterns [{}]", patterns);
+        if (excludedPatterns != null && excludedPatterns.size() > 0) {
+            LOG.warn("Replacing excluded patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
+                        excludedPatterns, patterns);
+        }
+        else {
+            LOG.debug("Replacing excluded patterns [{}] with [{}], be aware that this affects all instances and safety of your application!",
```
yasserzamani: It should be `LOG.debug("Sets excluded patterns [{}]", patterns);`

JCgH4164838Gh792C124B5: Hello @yasserzamani .
The two `debug` log statements have been updated in the most recent commit based on your comments/request.  I left in the extra warning text information but the wording and parameters of the start of the log output match what you requested.  :)

yasserzamani: Yes it's great thank you :)

@aleksandr-m , if you consent please approve your review, then we're OK to squash and merge this one :)

aleksandr-m: @yasserzamani Done. Squashing would be great. ;)

PR #296, opened 2018-12-17
Enhancement for s:set tag for 2.5.x to allow better tag body whitespace control.

JCgH4164838Gh792C124B5:
Enhancement for s:set tag to allow better tag body whitespace control.
- Allow disabling of the default body whitespace trim behaviour.  
- Allow stripping of all line breaks (carriage returns and linefeeds) from the body (vertical collapse).
- Both flags can be applied separately or together.
- _Default behaviour of_ `s:set` _tag remains unchanged_ (these are opt-in flags).

Implements [WW-4995](https://issues.apache.org/jira/browse/WW-4995)


JCgH4164838Gh792C124B5:
A Jira WW-4995 has been opened in relation to this PR.  Thanks for consideration in advance,

core/src/main/java/org/apache/struts2/components/Set.java
```diff
@@ -83,6 +83,8 @@
 public class Set extends ContextBean {
     protected String scope;
     protected String value;
+    protected boolean doNotTrimBody;
```
lukaszlenart: What about using `trimBody` with default set to `true`?

core/src/main/java/org/apache/struts2/components/Set.java
```diff
@@ -83,6 +83,8 @@
...
+    protected boolean stripBodyLineBreaks;
```
lukaszlenart: Shouldn't this be set to `true` by default to persist current behaviour?

core/src/main/java/org/apache/struts2/components/Set.java
```diff
@@ -136,6 +138,16 @@ public void setValue(String value) {
         this.value = value;
     }
 
+    @StrutsTagAttribute(description="Set to true to prevent the default whitespace-trim of this tag's body content")
+    public void setDoNotTrimBody(String doNotTrimBody) {
```
aleksandr-m: You should use `boolean` with correct `type` set in annotation. Take a look how it is done in `Property` and `PropertyTag` for example.

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
```diff
@@ -59,4 +61,32 @@ public void setScope(String scope) {
     public void setValue(String value) {
         this.value = value;
     }
+
+    public void setDoNotTrimBody(String doNotTrimBody) {
```
aleksandr-m: Same here: use boolean.

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
```diff
@@ -59,4 +61,32 @@ public void setScope(String scope) {
...
+        this.doNotTrimBody = Boolean.parseBoolean(doNotTrimBody);
+    }
+
+    public void setStripBodyLineBreaks(String stripBodyLineBreaks) {
+        this.stripBodyLineBreaks = Boolean.parseBoolean(stripBodyLineBreaks);
+    }
+
+    @Override
+    protected String getBody() {
+        if (bodyContent == null) {
```
aleksandr-m: This can be probably rewritten to use `super.getBody()` instead of duplicating code here.

core/src/main/java/org/apache/struts2/components/Set.java
aleksandr-m: Isn't this `stripBodyLineBreaks` too much specific to make additional attribute in `set` tag?

core/src/main/java/org/apache/struts2/components/Set.java
JCgH4164838Gh792C124B5: The requested change (switch to positive value with a default of true) has been made with the latest commit.

core/src/main/java/org/apache/struts2/components/Set.java
JCgH4164838Gh792C124B5: This has been addressed in the most recent commit (the setters and tag annotation now use boolean, as requested).

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
JCgH4164838Gh792C124B5: This has been addressed in the most recent commit (the setters and tag annotation now use boolean, as requested).

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
JCgH4164838Gh792C124B5: This has been addressed in the most recent commit (the `super.getBody()` is now called in `getBody()`, along with improved logic to ensure behaviour for nulls is the same as the original tag).

core/src/main/java/org/apache/struts2/components/Set.java
JCgH4164838Gh792C124B5: The `doNotTrimBody `was replaced with `trimBody `(set to true by default to mimic current behaviour) in the most recent commit.

The `stripBodyLineBreaks` , in my mind, had more immediate applicability to the `s:set` tag than most others.  I figured it could be made available to the `set` tag initially (a small and relatively safe change), then possibly considered for a more general future refactoring.

JCgH4164838Gh792C124B5:
Hello Apache Struts Team members.

This enhancement/fix should be OK for 2.5.19 since the 2 new attributes are optional and the pre-existing behaviour of any `s:set` tags in use by people's 2.5.x application will be preserved.  Not being able to `set` the _body content _as-is__ is something I was hoping to address in the 2.5.x series (and if accepted, forward-ported to 2.6). 

The line-break stripping is more enhancement than fix, but both were white-space related so I grouped them together.

Hopefully this change can be considered for 2.5.19 (and if so, then ported to 2.6).

lukaszlenart:
Looks good to me, no more objections.

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
```diff
@@ -59,4 +61,32 @@ public void setScope(String scope) {
...
+        this.trimBody = trimBody;
+    }
+
+    public void setStripBodyLineBreaks(boolean stripBodyLineBreaks) {
+        this.stripBodyLineBreaks = stripBodyLineBreaks;
+    }
+
+    @Override
+    protected String getBody() {
+        if (trimBody == true && stripBodyLineBreaks == false) {
```
sepe81: IMO `trimBody && !stripBodyLineBreaks` would be more concise.

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
```diff
@@ -59,4 +61,32 @@ public void setScope(String scope) {
...
+            return super.getBody();
+        } else {
+            String whitespaceModifiedBodyContent = (bodyContent == null ? "" : bodyContent.getString());
+            if (whitespaceModifiedBodyContent == null) {
+                whitespaceModifiedBodyContent = "";
+            } else if (whitespaceModifiedBodyContent.length() > 0) {
+                if (stripBodyLineBreaks) {
+                    whitespaceModifiedBodyContent = whitespaceModifiedBodyContent.replaceAll("[\r\n]+?", "");
+                }
+                if (trimBody == true) {
```
sepe81: same here: `trimBody` to be consistent with L:82

core/src/main/java/org/apache/struts2/components/Set.java
aleksandr-m: @JCgH4164838Gh792C124B5 Can you show some use case for removing line breaks?
There is a good reason to introduce not trimming because there is no way to get real value in set tag. But `stripBodyLineBreaks` looks way too specific. It is like replacing `a` letters with `b`. Should we create an attribute for that too? :)
Besides this really should be done in the controller not in the view. And even if you want to do this in the view then you can call `replaceAll` directly in the jsp.

JCgH4164838Gh792C124B5: Hello @aleksandr-m .
The `stripBodyLineBreaks` was more of a convenience enhancement idea than a necessity.  The use cases for it are not very strong, as you pointed out above.  Given the concerns raised for the `stripBodyLineBreaks` attribute it has been removed in the latest commit - which also simplifies the `getBody()` logic and test changes too.  :)

core/src/main/java/org/apache/struts2/views/jsp/SetTag.java
JCgH4164838Gh792C124B5: Hello @sepe81 .
The `stripBodyLineBreaks` attribute was removed in the latest commit (as per @aleksandr-m 's comments), which allowed for a significant cleanup of the `getBody()` method.  The updated code has the cleaner-style `if` block usage you had asked for.  :)

PR #297, opened 2018-12-18
[WW-4996] Refactor DefaultTypeConverterCreator to use ObjectFactory#buildConverter

lukaszlenart:
Implements [WW-4996](https://issues.apache.org/jira/browse/WW-4996)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/factory/DefaultConverterFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/factory/DefaultConverterFactory.java
@@ -42,7 +42,7 @@ public class DefaultConverterFactory implements ConverterFactory {
 
     public TypeConverter buildConverter(Class<? extends TypeConverter> converterClass, Map<String, Object> extraContext) throws Exception {
         LOG.debug("Creating converter of type [{}]", converterClass.getCanonicalName());
-        return container.getInstance(converterClass);
+        return container.inject(converterClass);
     }
 
 }
--- a/core/src/main/java/org/apache/struts2/conversion/StrutsTypeConverterCreator.java
+++ b/core/src/main/java/org/apache/struts2/conversion/StrutsTypeConverterCreator.java
@@ -22,7 +22,6 @@ import com.opensymphony.xwork2.ObjectFactory;
 import com.opensymphony.xwork2.conversion.TypeConverter;
 import com.opensymphony.xwork2.conversion.TypeConverterCreator;
 import com.opensymphony.xwork2.inject.Inject;
-import com.opensymphony.xwork2.ognl.XWorkTypeConverterWrapper;
 
 /**
  * Default implementation of {@link TypeConverterCreator}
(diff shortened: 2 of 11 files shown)


aleksandr-m:
Just curious. What is wrong with `Default*`? :) There are lot of them in configuration. Are there any plans to change all of them to `Struts*`?

lukaszlenart:
`Default*` means nothing tbh, I've started with this pattern some time ago but after using it in many places I noticed that I don't really understand what `Default*` means. Using `Struts*` instead gives a clear message that this is a Struts implementation (whatever that means). Compare this with `ObjectFactory` and `SpringObjectFactory` and `CdiObjectFactory`. And now renaming `ObjectFactory` to `DefaultObjectFactory` means nothing, you still don't know what mechanism was used inside, but using `StrutsObjectFactory` at least you understand that this is a Struts-based implementation of the factory. Also it's far more readable when you inherits from `StrutsObjectFactory` than from `DefaultObjectFactory`.

Also, all the default beans defined in `struts-default.xml` will use `struts` as a bean name, this corresponds to internal DI mechanism which will use those `struts`' beans as default implementations.

lukaszlenart:
@aleksandr-m any other objections before merging this?

aleksandr-m:
@lukaszlenart Nope.

lukaszlenart:
@apache/struts-committers can someone hit the button? :)

core/src/main/java/org/apache/struts2/conversion/StrutsTypeConverterCreator.java
```diff
@@ -37,24 +36,22 @@ public void setObjectFactory(ObjectFactory objectFactory) {
...
     public TypeConverter createTypeConverter(String className) throws Exception {
-        Object obj = objectFactory.buildBean(className, null);
-        return getTypeConverter(obj);
-    }
+        Class<?> classInstance = objectFactory.getClassInstance(className);
 
-    public TypeConverter createTypeConverter(Class<?> clazz) throws Exception {
-        Object obj = objectFactory.buildBean(clazz, null);
-        return getTypeConverter(obj);
+        if (TypeConverter.class.isAssignableFrom(classInstance)) {
```
yasserzamani: I think this if-else can be replaced with `return createTypeConverter(classInstance)`

core/src/main/java/com/opensymphony/xwork2/factory/DefaultConverterFactory.java
```diff
@@ -42,7 +42,7 @@ public void setContainer(Container container) {
 
     public TypeConverter buildConverter(Class<? extends TypeConverter> converterClass, Map<String, Object> extraContext) throws Exception {
         LOG.debug("Creating converter of type [{}]", converterClass.getCanonicalName());
-        return container.getInstance(converterClass);
+        return container.inject(converterClass);
```
yasserzamani: why?

core/src/main/java/org/apache/struts2/conversion/StrutsTypeConverterCreator.java
lukaszlenart: 👍

core/src/main/java/com/opensymphony/xwork2/factory/DefaultConverterFactory.java
lukaszlenart: `getInstance()` can only create instances of classes already registered in container, thus means you would have to register all the converters used in your app as beans in `struts.xml`.

`inject()` is a factory method, it will create an instance of the class and inject all the marked dependencies.

`buildConverter()` was not used (a dead code), as `buildBean` was used instead to create converters, that's why that wasn't discovered early.

yasserzamani:
LGTM 👍 thank you!

PR #298, opened 2018-12-18
Fix sonar warnings for DefaultTypeConverter

sepe81:


core/src/main/java/com/opensymphony/xwork2/conversion/impl/DefaultTypeConverter.java
```diff
@@ -63,7 +63,7 @@
         map.put(Float.TYPE, new Float(0.0f));
         map.put(Double.TYPE, new Double(0.0));
         map.put(BigInteger.class, new BigInteger("0"));
-        map.put(BigDecimal.class, new BigDecimal(0.0));
+        map.put(BigDecimal.class, BigDecimal.valueOf(0.0));
```
aleksandr-m: Tiny nitpick: there are `BigDecimal.ZERO` and `BigInteger.ZERO`.

sepe81: I considered it. But b/c of https://stackoverflow.com/questions/9992006/equals-operator-for-zeros-bigdecimal-double-in-java I was not quite sure if 0 is the right substitute for 0.0 at this place. What do you think - should we try anyways?

aleksandr-m: [`equals` shouldn't be used to compare BigDecimals](http://www.opentaps.org/docs/index.php/How_to_Use_Java_BigDecimal:_A_Tutorial)

And it is seems you've already changed it :)
```
BigDecimal.ZERO.equals(BigDecimal.valueOf(0.0)) // false
BigDecimal.ZERO.equals(new BigDecimal(0.0)) // true
```

sepe81: @aleksandr-m Considering that no tests are failing, what do you prefer? Should I revert this line (no risk of changed behaviour in depending code) or maybe better switch to `BigDecimal.ZERO` + `BigInteger.ZERO` for L:66.

@lukaszlenart do you have any thoughts about that?

core/src/test/java/com/opensymphony/xwork2/conversion/impl/XWorkBasicConverterTest.java
```diff
@@ -228,6 +229,37 @@ public void testPositiveDoubleValue() throws Exception {
         assertEquals(94.1231233d, (Double) convertedObject, 0.0001);
     }
 
+    public void testBigDecimal() {
+        Object convertedObject = basicConverter.convertValue(null, BigDecimal.class);
```
sepe81: @aleksandr-m @lukaszlenart I added a new test that also succeeds for the current master branch.

Conclusion: IMHO we can (should) remove all non-primitive types from the `Map` of `primitiveDefaults` within `DefaultTypeConverter.class b/c they won't ever be used. ;-)

The only access to `primitiveDefaults is in case of `value == null` AND if `toType.isPrimitive()`. This is not the case for `BigDecimal` therefore `null` will be returned -> see L:233-234

sepe81: Even more: non of those types are primitive b/c ALL are class types. Do I miss something? Otherwise the whole map + check could be removed... :confused:

sepe81: Also see https://stackoverflow.com/questions/709961/determining-if-an-object-is-of-primitive-type

lukaszlenart: I would say go ahead and we will see :)

core/src/main/java/com/opensymphony/xwork2/conversion/impl/DefaultTypeConverter.java
lukaszlenart: I would assume `BigDecimal.valueOf(0.0)` is better here

core/src/test/java/com/opensymphony/xwork2/conversion/impl/XWorkBasicConverterTest.java
aleksandr-m: @sepe81 :) Nice find. Go ahead.

sepe81: I concluded my new findings within the commit comment of d09d3dc.

core/src/main/java/com/opensymphony/xwork2/conversion/impl/DefaultTypeConverter.java
sepe81: I think we can go for `BigInteger.ZERO` and `BigDecimal.ZERO as @aleksandr-m proposed.

see commit d09d3dc

core/src/test/java/com/opensymphony/xwork2/conversion/impl/XWorkBasicConverterTest.java
aleksandr-m: Well, I'm curious now, why `BigInteger` and `BigDecimal` default to `0` instead of `null`. @lukaszlenart Do you have any idea?

Regarding this PR it looks good. :+1:

lukaszlenart: No idea, this class is one of the oldest and I would like to finish cleaning up the conversion logic at some point :)

lukaszlenart:
If no objections LGTM 👍

PR #299, opened 2018-12-21
Enhancement for s:set tag to allow better tag body whitespace control.

JCgH4164838Gh792C124B5:
Enhancement for `s:set` tag to allow better tag body whitespace control.
- Allow disabling of the default body whitespace trim behaviour.  
- Forward port change from PR #296 for WW-4995 into master (2.6)


aleksandr-m:
@JCgH4164838Gh792C124B5 Thanks. BTW if the code is the same then you don't have to open separate PR for each branch. We can cherry pick commits into another branch.

aleksandr-m:
@lukaszlenart Do we need to push generated tags html to repo? Is it used for something? Maybe we should exclude it from repo.

lukaszlenart:
I think we use them in our docs

PR #307, opened 2018-12-31
Update Embedded JSP Plugin (more current Juli/Jasper versions)

JCgH4164838Gh792C124B5:
Update Embedded JSP Plugin to utilize the most current Tomcat Juli/Jasper still compatible with JDK 7.

- Replaced AnnotationHelper with InstanceHelper due to removal of AnnotationProcessor in Tomcat 7+.
- Tomcat Juli/Jasper Maven artifact ids changed with Tomcat 7+.
- Addition of tomcat-api dependency due to changes in Jasper for Tomcat 7+.
- Applied changes to embedded plugin modules required due to Jasper API changes.
- Additional unit tests for better test coverage.


JCgH4164838Gh792C124B5:
Jenkins build is failing.
- Although a new length test in a new unit tests worked properly locally (Windows 10), it appears to be failing on the remote Jenkins build. 
- Will have to look to see if the size variation can be explained (the old tests didn't attempt any similar length checks) and the relative position checks all passed since that was the last check ...

JCgH4164838Gh792C124B5:
Found the issue (length miscalculation for crlf in Windows vs. lf-only in LInux/Unix) and updated the unit test to compensate (and actually output the index/lengths if things fail).  The Jenkins build now succeeds for this PR.

Guess this is a reminder that trying to commit-when-tired might not be a good idea.  :) :)

lukaszlenart:
I'm ok with merging this in even if this plugin will be dropped and replaced by native support of embedded JSPs in Servlet 3.0 API (as we switch to it at some time)

yasserzamani:
I afraid new changes always potentially will arise bugs which forces us to support a deprecated plugin, BUT beside this, @JCgH4164838Gh792C124B5's fantastic works are too good to not merging them :(

lukaszlenart:
LGTM 👍

PR #308, opened 2019-01-03
[WW-4943] Tries to use parent CL if no bundle for current CL

lukaszlenart:
Fixes [WW-4943](https://issues.apache.org/jira/browse/WW-4943)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
@@ -90,7 +90,30 @@ abstract class AbstractLocalizedTextProvider implements LocalizedTextProvider {
     }
 
     protected List<String> getCurrentBundleNames() {
-        return classLoaderMap.get(getCurrentThreadContextClassLoader().hashCode());
+        int currentKey = getCurrentThreadContextClassLoader().hashCode();
+
+        List<String> currentBundleName = classLoaderMap.get(currentKey);
+
+        if (currentBundleName == null) {
+            currentBundleName = getParentBundleNames(currentKey);
+        }
+
+        return currentBundleName;
+    }
+
+    private List<String> getParentBundleNames(int currentKey) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
```diff
@@ -90,7 +90,30 @@ public void addDefaultResourceBundle(String resourceBundleName) {
...
+    }
+
+    private List<String> getParentBundleNames(int currentKey) {
+        List<String> parentBundleName = null;
+
+        ClassLoader parentCL = getCurrentThreadContextClassLoader().getParent();
+        if (parentCL != null) {
+            LOG.debug("No bundle name for current ClassLoader, trying with parent ClassLoader");
+            parentBundleName = classLoaderMap.get(parentCL.hashCode());
+            if (parentBundleName != null && !delegatedClassLoaderMap.containsKey(currentKey)) {
```
yasserzamani: I think `parentBundleName` always will be null because `classLoaderMap` is private and all of it's `put` calls are only with `getCurrentThreadContextClassLoader().hashCode()`.

JCgH4164838Gh792C124B5: Hmmm...I missed a few nuances in the earlier read-through.

Maybe the WW-4943 report was indicating that there is a use case where the `getCurrentThreadContextClassLoader()` result is a child classloader of the parent that initialized the web-application classloader/context , but still somehow executing requests through the parent's instance of `AbstractLocalizedTextProvider` ?

Assuming that condition could arise (contrived or not), then the logic for this PR would appear to work (since the parent cl's key would match the original key used to store original bundle names).  It would be a strange condition though ...

lukaszlenart: Yeah... I have the same thoughts and that's why I stopped working on this issue till I get more info from the reporter.

PR #309, opened 2019-01-04
Address devMode log flooding concern raised by Greg Huber (on Dev List)

JCgH4164838Gh792C124B5:
Address `devMode` log flooding concern raised by Greg Huber in the Struts Dev List review of the 2.5.19 test build.
Changes:
- Introduce a static count-limiter for the `devMode` set warning in `OgnlValueStack`.  It _limits the total number of log warnings produced to 250 total_, after which it switches to trace level logs to avoid log flooding.

This is a single change which should be safe enough to include in a revised 2.5.19 build (and cherry-picked into the 2.6 master branch as well).  Since the log flooding could be somewhat annoying to developers, a compromise like this might be a reasonable response.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -40,6 +40,7 @@ import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
 
 /**
@@ -60,6 +61,10 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     private static final long serialVersionUID = 370737852934925530L;
 
     private static final String MAP_IDENTIFIER_KEY = "com.opensymphony.xwork2.util.OgnlValueStack.MAP_IDENTIFIER_KEY";
+
+    private static final int MAX_DEVMODESET_WARNCOUNT = 250;
+
+    private static final AtomicInteger DEVMODESET_WARNCOUNT = new AtomicInteger(0);
 
     protected CompoundRoot root;
     protected transient Map<String, Object> context;
@@ -105,8 +110,19 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     protected void setDevMode(String mode) {
         this.devMode = BooleanUtils.toBoolean(mode);
         if (this.devMode) {
-            LOG.warn("Setting development mode [{}] affects the safety of your application!",
-                        this.devMode);
+            int devModeWarnCount = DEVMODESET_WARNCOUNT.get();
+            if (devModeWarnCount < MAX_DEVMODESET_WARNCOUNT) {
+                devModeWarnCount = DEVMODESET_WARNCOUNT.incrementAndGet();
+                LOG.warn("Setting development mode [{}] affects the safety of your application!",
+                            this.devMode);
+                if (devModeWarnCount >= MAX_DEVMODESET_WARNCOUNT) {
+                    LOG.warn("Last warning concerning development mode [{}]!  These messages will now be trace level only!",
+                                this.devMode);
+                }
+            } else {
+                LOG.trace("Setting development mode [{}] affects the safety of your application!",
+                            this.devMode);
+            }
         }
     }
 
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -40,7 +40,7 @@ import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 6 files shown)


JCgH4164838Gh792C124B5:
The limit value of 250 was a best-guess at balanced number to apply the limit.  If the Apache Struts Team has a preferred limit value to use instead, the value can be changed to match that.  

A value of 100 or 150 seemed too low (quickly buried in historical logs as the number of log-warns-per-request can get there fairly quickly).  A configurable value might be nice, but is too complicated for a "quick-fix" response to the concern raised.

aleksandr-m:
Too much hustle for the logs. Just remove it or change to `trace`. I would also suggest to do the same for all other logs introduced in previous commit.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -105,8 +110,19 @@ protected void setRoot(XWorkConverter xworkConverter, CompoundRootAccessor acces
     protected void setDevMode(String mode) {
         this.devMode = BooleanUtils.toBoolean(mode);
         if (this.devMode) {
-            LOG.warn("Setting development mode [{}] affects the safety of your application!",
-                        this.devMode);
+            int devModeWarnCount = DEVMODESET_WARNCOUNT.get();
+            if (devModeWarnCount < MAX_DEVMODESET_WARNCOUNT) {
+                devModeWarnCount = DEVMODESET_WARNCOUNT.incrementAndGet();
+                LOG.warn("Setting development mode [{}] affects the safety of your application!",
```
sepe81: The message could be extracted and reused to avoid string duplication.

JCgH4164838Gh792C124B5:
Hello @aleksandr-m  and @sepe81 .
The previous commit has been updated with changes that should address the concern for

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -61,6 +62,12 @@
 
     private static final String MAP_IDENTIFIER_KEY = "com.opensymphony.xwork2.util.OgnlValueStack.MAP_IDENTIFIER_KEY";
 
+    private static final String EXCEPTION_EVALUATING_EXPRESSION_STR = "Caught an exception while evaluating expression '{}' against value stack";
+
+    private static final String DEVMODE_SAFETY_STR = "Setting development mode [{}] affects the safety of your application!";
+
+    private static final AtomicBoolean INITIAL_WARNING_FOR_DEV_MODE = new AtomicBoolean(true);
```
yasserzamani: Thank you @JCgH4164838Gh792C124B5 but still it seems too complicated logic for me for just a log line. I think we can revert it to what it was before. The phrase `devMode` I think already conventionally WARNs user it's for development environment.

And could you please revise all other logs introduced in same PR's commits if they may raise similar issues? Thanks in advance!

JCgH4164838Gh792C124B5: Hello Again.

During the review of logs introduced in the previous PR, I decided to go ahead and apply the same only-warn-once-logic to the relevant modules (OgnlUtil, OgnlValueStackFactory, CompoundRootAccessor, DefaultAcceptedPatternsChecker, DefaultExcludedPatternsChecker).  Also simplified the `devMode` log warn for OgnlValueStack, and committed the end-result for consideration.

If the reasoning in the previous comment is considered sufficient (as it applies to all of the preceding), then some or all of the changes could be applied ... and if nothing else it makes the the logging locations are easier to track for this PR.  :)  :)

yasserzamani: Thank you @JCgH4164838Gh792C124B5 for your suggested benefits for Struts :) yes I already knew and understand what are you're trying to achieve. Actually I not only was worry for performance but also for somehow more important thing, the code readability and nice logic/design. Benefit#1 is good but I think the injection point isn't good place to log because of three reasons: 1) In future, anyone who creates a new class that needs devMode has to remember or simply will forget to consider this log consideration 2) It violates DRY (Don't Repeat Yourself) 3) For not singleton beans, it causes log floods and an angry user :) I guess not only your new added log statements but also already existent e.g. [here](https://github.com/apache/struts/pull/291/files#diff-ac4c516bdbcef4755bd30af8d088e5b5L51) may suffer from this issue. It depends, it may work on singleton beans but still suffers from reason#1&2. Benefit#2 already has been addressed as much as was logically possible by refactoring public to protected. Any further improvements are welcomed only if they don't violate reason#1&2 or any other best practice :)

I think reviewing #291 and reverting potential issues for now. Then we can consider these benefits in a more appropriate place (rather than a bottle neck like prototype beans injection points). wdyt?

JCgH4164838Gh792C124B5: Hello @yasserzamani .
Thanks for the review.  :)

I just wanted to make sure of my points before going ahead and reverting most of the log entries introduced with #291 .  Thanks for taking the time to provide the counterpoints to the ones I raised and giving good reasons as to why the added logging is more-trouble-than-its-worth.  :) :)

The last commit has been amended again and effectively removes all new logging introduced (except for the accepted/excluded locations, that if I understood correctly, @lukaszlenart indicated should probably remain).  If I misunderstood that point it's a quick fix.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .
Thanks for providing your feedback/comments as well!

Based on the last sentence of your comment I interpreted that to mean that the "extra" logging added for the setExcludedPatterns/setAcceptedPatterns still serves a purpose and so should remain.  The latest amended commit reflects that, while removing all the other added logging from the original PR.

If I mis-understood your comment, just let me know and I can revert those two elements back to the original trace output as well !

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
yasserzamani: Thanks again! I reviewed `AcceptedPatternsChecker` and `ExcludedPatternsChecker` interfaces and how they work. I feel their new warn logs should be removed or changed to debug level as well. Because firstly `this affects all instances and safety of your application` seems isn't quite true. Those interfaces are `prototype` beans injected to parameters and cookie interceptors. Maybe user adds an accepted pattern to only one specific action. This doesn't affects all instances and safety of application but user sees numerous of this warning; each action call prints this :(

lukaszlenart: @yasserzamani those are prototypes on purpose - interceptors are singletons per package which means those instance will be injected just once into the interceptors.

aleksandr-m:
Master branch should be updated as well.

JCgH4164838Gh792C124B5:
For the updates to the master branch, let me know if you prefer to "cherry-pick" from this PR or would prefer creation of a second PR with equivalent changes for the master (2.6) branch.

lukaszlenart:
LGTM 👍

lukaszlenart:
@JCgH4164838Gh792C124B5 I think a new PR is preferable in this case

PR #310, opened 2019-01-07
WW-4998 I18nInterceptor's default storage should store locale

aleksandr-m:


core/src/main/java/org/apache/struts2/interceptor/I18nInterceptor.java
```diff
@@ -281,15 +282,13 @@ public Locale find() {
 
         @Override
         public Locale store(ActionInvocation invocation, Locale locale) {
-            HttpSession session = ServletActionContext.getRequest().getSession(false);
+            Map<String, Object> session = invocation.getInvocationContext().getSession();
```
lukaszlenart: Why not to call `HttpSession session = ServletActionContext.getRequest().getSession();` directly? The the `if` below can be dropped.

aleksandr-m: I think the `invocation.getInvocationContext().getSession()` can be `null` in some cases, can't it? And we need it to put locale.

lukaszlenart: ah, right :)

lukaszlenart:
@apache/struts-committers if no objections, LGTM?

yasserzamani:
LGTM too 👍

lukaszlenart:
LGTM 👍

PR #311, opened 2019-01-10
Equivalent to PR#309, but for the master (2.6.x) branch.

JCgH4164838Gh792C124B5:
Equivalent to PR#309 (for 2.5.x), but this one is for the master (2.6.x) branch.
- Removed most of the logging added in PR#292.
- Left the added logging for the `setExcludedPatterns `and `setAcceptedPatterns `methods in their respective modules (please see PR#309 for justification).  Made the if-else blocks consistent with preferred styling for the log blocks.
- Removed a dangling "/\*\*" start comment tag with no proper close (and the whitespace between it and the next "/\*\*") in the `OgnlValueStack `module.


lukaszlenart:
LGTM 👍

PR #312, opened 2019-01-12
 Switch to Java 8 

aleksandr-m:
[WW-5010](https://issues.apache.org/jira/browse/WW-5010)


aleksandr-m:
@apache/struts-committers Seems like jenkins doing all PR-s on jdk7. Can someone do something with that?

yasserzamani:
I configured that job to use java 8 as a quick fix to see what happens but about what we should do with 2.5.x branch PRs, I thought it should be fine to share the job with them also because 2.5.x already seems fine with java 8.

aleksandr-m:
Shouldn't Jenkinsfile take care of that? Is it used at all, currently?

lukaszlenart:
Jenkinsfile doesn't cover all the builds, it's a plan for the future to have all of them configure throughout the Jenkinsfile.

aleksandr-m:
@lukaszlenart Is there any problems using Jenkinsfile? We need some solution to build PR-s in 2.5.x and 2.6 branches.

yasserzamani:
We are already OK with PRs. This one will build also if you push a new one :)

lukaszlenart:
@aleksandr-m it supposed to work but as you can notice it doesn't and I don't know why (I must ask our DevOps guru ;-)

aleksandr-m:
> We are already OK with PRs. This one will build also if you push a new one :)

@yasserzamani What do you mean?

How Jenkins will know which jdk to use for PR-s for different branches?

lukaszlenart:
@aleksandr-m it's possible as Jenkins can create a different builds based on Jenkinsfile found in a branch - so we can have two different Jenkinsfile to support Struts 2.5.x with JDK7 and Struts 2.6.x with JDK8.

And I finally have an initial Jenkinsfile working
https://builds.apache.org/view/S-Z/view/Struts/job/Struts-Seed/

lukaszlenart:
Jenkins rebuild this

aleksandr-m:
> it's possible as Jenkins can create a different builds based on Jenkinsfile found in a branch

Yes. That is exactly what I referred to. :)

> And I finally have an initial Jenkinsfile working

Great!

aleksandr-m:
@yasserzamani I don't think that building 2.5.x branch with jdk8 is a good idea. It is easy to miss some stuff that can only be build/run in java8.

Now when Jenkinsfile is used we can rely on that.

aleksandr-m:
@apache/struts-committers If no objections, I'm going to merge this in a few days.

yasserzamani:
Me too 👍

PR #315, opened 2019-01-23
WW-5005 upgrade to ASM 7

yasserzamani:
fixes WW-5005

According to ASM7's source codes, it seems it uses passed `api` version only to check if it should throw an unsupported exception or not (i.e. doesn't have any effect on functionality and somehow is just a validation). So, as we upgrade to asm.version=7.0 for default then I think it's OK to upgrade `Opcodes.ASM5` to `Opcodes.ASM7` as well. It only breaks when user manually exclude ASM7 dependency with an older version which is not usually expected.


yasserzamani:
Thanks a lot for your time to explore, review and confirmation! It makes us more confident to merge :)

lukaszlenart:
@yasserzamani conflicts :(

lukaszlenart:
LGTM 👍

lukaszlenart:
Could you cherry-pick this change to the master?

PR #316, opened 2019-01-24
Fix for NPE issue discovered in WW-5004.

JCgH4164838Gh792C124B5:
Fix for `NPE` issue discovered in WW-5004.
- Guard fix for a `NPE` that can arise under certain conditions, identified by _A. Mashchenko_.
- This guard (or something similar) will probably be required for future non-public field access logic ... may want to hold-off on this PR and instead include the fix in a "fuller solution".

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
@@ -19,6 +19,7 @@
 package com.opensymphony.xwork2.ognl;
 
 import com.opensymphony.xwork2.ActionContext;
+import com.opensymphony.xwork2.XWorkConstants;
 import com.opensymphony.xwork2.TextProvider;
 import com.opensymphony.xwork2.conversion.NullHandler;
 import com.opensymphony.xwork2.conversion.impl.XWorkConverter;
@@ -61,7 +62,7 @@ public class OgnlValueStackFactory implements ValueStackFactory {
         this.textProvider = textProvider;
     }
     
-    @Inject(value="allowStaticMethodAccess", required=false)
+    @Inject(value = XWorkConstants.ALLOW_STATIC_METHOD_ACCESS, required = false)
     protected void setAllowStaticMethodAccess(String allowStaticMethodAccess) {
         this.allowStaticMethodAccess = BooleanUtils.toBoolean(allowStaticMethodAccess);
     }
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -49,7 +49,7 @@ public class SecurityMemberAccess extends DefaultMemberAccess {
     /**
      * SecurityMemberAccess
      *   - access decisions based on whether member is static (or not)
-     *   - block or allow access to properties (configureable-after-construction)
+     *   - block or allow access to properties (configurable-after-construction)
      * 
      * @param allowStaticMethodAccess
      */
@@ -104,7 +104,7 @@ public class SecurityMemberAccess extends DefaultMemberAccess {
         }
 
         boolean allow = true;
-        if (!checkStaticMethodAccess(member)) {
+        if (!checkStaticMemberAccess(member)) {
             LOG.warn("Access to static [{}] is blocked!", member);
             allow = false;
         }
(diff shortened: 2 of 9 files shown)


lukaszlenart:
That's how it was fixed (by you btw ;-) in the master branch
https://github.com/apache/struts/blob/master/core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java#L90

I think it would be better to use the same approach

yasserzamani:
Yes I agree. I remember those improvements but I just wonder why I didn't import them into 2.5 as well :((

lukaszlenart:
Too many versions to support :)

lukaszlenart:
I would like to have a stable 2.5.x version and then we can focus on 2.6 :)

JCgH4164838Gh792C124B5:
Hello @lukaszlenart and @yasserzamani .

I wasn't sure if it was better to leave the NPE in 2.5.x that arises with OGNL 3.1.19-3.1.21 or not.  Assuming the fix for 3.1.22 restores the original functionality that was available in Struts 2.5.18 for public static fields, I guess it would be OK for the NPE to still occur (which is caught and just causes the access to fail) for non-public static fields.

Once 3.1.22 is available I could create a PR that upgrades to that version and attempts to introduce unit tests that confirm the public access works and the non-public access fails (as it did with Struts 2.5.18).

What do you think ?

JCgH4164838Gh792C124B5:
Hello.

The most recent commit appears to resolve the NPE issue found (and the related issues as well).
The new unit test indicates success using OGNL 3.1.15, 3.1.18 and 3.1.22 (and failures for 3.1.19-3.1.21, detecting previously missed failure for static public field access in some circumstances).
_Note_:  Since non-public static fields were not previously accessible with 2.5.x, this commit attempts to ensure that remains the case for 2.5.21+ (leaving such access to be considered as possible future feature for 2.6.x).

Please advise if you think it looks OK or if some changes should be considered.  :)

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -121,6 +121,9 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
     protected boolean checkStaticMethodAccess(Member member) {
         int modifiers = member.getModifiers();
         if (Modifier.isStatic(modifiers)) {
+            if (member instanceof Field) {
+                return false;  // Refuse access to all static fields (even if allowStaticMethodAccess true)
+            }
```
lukaszlenart: but access to public static fields were allowed even before

JCgH4164838Gh792C124B5: Hello @lukaszlenart .

Access to public static fields is still allowed (the unit tests for this PR check that), but it works again due to the OGNL 3.1.22 fix (which restored the 3.1.18 public static field access behaviour - no isAccessible check for public static fields).
Basically since the OGNL 3.1.x "public static field access behaviour contract" means checkStaticMethodAccess should never be called in the flow for public static fields, the above was just to block all field access. 

But now that you draw attention to it again, it would probably be clearer/better to base it on the public state, even if the check "never should happen".  I'll update the PR to reflect that.

JCgH4164838Gh792C124B5: Hello.

This PR has been updated in response to the comment above (so the method will return true for static public fields, even if that path is not normally called with most OGNL 3.1.x versions).  The `SecurityMemberAccess` module `checkStaticMethodAccess` method was deprecated and replaced with the more appropriately named `checkStaticMemberAccess` (so that in the unlikely event someone extended the class things should still work OK).

One additional cleanup was added as well (noticed during the similar PR for 2.6.x), to utilize a constant instead of a literal for the static access injection name in OgnlValueStackFactory.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -61,7 +62,7 @@ protected void setTextProvider(TextProvider textProvider) {
         this.textProvider = textProvider;
     }
     
-    @Inject(value="allowStaticMethodAccess", required=false)
+    @Inject(value = XWorkConstants.ALLOW_STATIC_METHOD_ACCESS, required = false)
```
lukaszlenart: 👏

JCgH4164838Gh792C124B5: Hello @lukaszlenart .

The only other consideration might be whether to add in the changes being reviewed for SecurityMemberAccess from PR#320 (the elements valid for 2.5.x anyway ... minus the new field access flag), provided they make sense for 2.5.x.

If you think those changes (the ordering optimization for public state checks, et. al.) would be valid here as well I can make a single commit update for SecurityMemberAccess to this PR.

Please let me know what you think.  :)

lukaszlenart: This PR is already huge with a long discussion, let's merge it first then we can work on another issue.

PR #317, opened 2019-01-25
[WW-5004] Fixes changing "struts.ognl.allowStaticMethodAccess"

lukaszlenart:
This PR relates to [WW-5004](https://issues.apache.org/jira/browse/WW-5004) to allow change the flag to a non-default value.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
@@ -24,23 +24,29 @@ import com.opensymphony.xwork2.XWorkConstants;
 import com.opensymphony.xwork2.conversion.NullHandler;
 import com.opensymphony.xwork2.conversion.impl.XWorkConverter;
 import com.opensymphony.xwork2.inject.Container;
+import com.opensymphony.xwork2.inject.EarlyInitializable;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.ognl.accessor.CompoundRootAccessor;
+import com.opensymphony.xwork2.ognl.accessor.ParameterPropertyAccessor;
 import com.opensymphony.xwork2.util.CompoundRoot;
 import com.opensymphony.xwork2.util.ValueStack;
 import com.opensymphony.xwork2.util.ValueStackFactory;
 import ognl.MethodAccessor;
 import ognl.OgnlRuntime;
 import ognl.PropertyAccessor;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.apache.logging.log4j.message.SimpleMessage;
+import org.apache.struts.util.PropertyMessageResources;
 
 import java.util.Set;
 
 /**
  * Creates an Ognl value stack
  */
-public class OgnlValueStackFactory implements ValueStackFactory {
+public class OgnlValueStackFactory implements ValueStackFactory, EarlyInitializable {
     
     private static final Logger LOG = LogManager.getLogger(OgnlValueStackFactory.class);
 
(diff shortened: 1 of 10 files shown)


lukaszlenart:
I will try to add a test case for this change

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -61,20 +59,17 @@ protected void setTextProvider(TextProvider textProvider) {
...
+        String allowStaticMethodAccess = container.getInstance(String.class, XWorkConstants.ALLOW_STATIC_METHOD_ACCESS);
+        ValueStack stack = new OgnlValueStack(xworkConverter, compoundRootAccessor, textProvider, Boolean.valueOf(allowStaticMethodAccess));
         container.inject(stack);
         stack.getContext().put(ActionContext.CONTAINER, container);
         return stack;
     }
 
     public ValueStack createValueStack(ValueStack stack) {
-        ValueStack result = new OgnlValueStack(stack, xworkConverter, compoundRootAccessor, allowStaticMethodAccess);
+        String allowStaticMethodAccess = container.getInstance(String.class, XWorkConstants.ALLOW_STATIC_METHOD_ACCESS);
```
JCgH4164838Gh792C124B5: Hello @lukaszlenart.

Based on an inconsistency found as part of PR #320, it appears that  `XWorkConstants.ALLOW_STATIC_METHOD_ACCESS` should be replaced with `StrutsConstants.STRUTS_ALLOW_STATIC_METHOD_ACCESS` (as STRUTS_ALLOW_STATIC_METHOD_ACCESS is the value used elsewhere in 2.6.x for the static access control flag).

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -61,20 +59,17 @@ protected void setTextProvider(TextProvider textProvider) {
...
     }
     
-    @Inject(value="allowStaticMethodAccess", required=false)
-    protected void setAllowStaticMethodAccess(String allowStaticMethodAccess) {
-        this.allowStaticMethodAccess = BooleanUtils.toBoolean(allowStaticMethodAccess);
-    }
-
     public ValueStack createValueStack() {
-        ValueStack stack = new OgnlValueStack(xworkConverter, compoundRootAccessor, textProvider, allowStaticMethodAccess);
+        String allowStaticMethodAccess = container.getInstance(String.class, XWorkConstants.ALLOW_STATIC_METHOD_ACCESS);
```
JCgH4164838Gh792C124B5: The same comment holds here.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
lukaszlenart: But this change targets 2.5.x, those constants were dropped in 2.6 only (`master` branch)

JCgH4164838Gh792C124B5: Hello @lukaszlenart .

Ahhh .... sorry ... I totally missed that (jumping around between the two branches for some reason I had assumed this PR was referring to 2.6 as opposed to 2.5.x).  If/when this change moves to the 2.6 branch the updated constant would naturally be used.

So please ignore my previous comment-in-error.  :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -61,20 +59,17 @@ protected void setTextProvider(TextProvider textProvider) {
...
+        ValueStack stack = new OgnlValueStack(xworkConverter, compoundRootAccessor, textProvider, Boolean.valueOf(allowStaticMethodAccess));
```
JCgH4164838Gh792C124B5:
```
final boolean allowStaticMethodAccess = BooleanUtils.toBoolean(container.getInstance(String.class, XWorkConstants.ALLOW_STATIC_METHOD_ACCESS)); 
ValueStack stack = new OgnlValueStack(xworkConverter, compoundRootAccessor, textProvider, allowStaticMethodAccess);
```
Now that I'm referring to the correct branch ( :) ) as per my original comment this might be an alternate form for the 2 new code blocks.  
Also ... I forgot that BooleanUtils was present, so it might be better to use it rather than Boolean.parseBoolean.

lukaszlenart:
Hm... something went wrong with merges, I will pick changes up to another branch

PR #323, opened 2019-01-31
Improve SecurityMemberAccess

aleksandr-m:
Move public member check to be the first performed
Remove final-s from methods to allow overriding
Split check for statics into two separate methods

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -108,7 +108,7 @@ public class SecurityMemberAccess implements MemberAccess {
         final int memberModifiers = member.getModifiers();
 
         if (!checkPublicMemberAccess(memberModifiers)) {
-            LOG.trace("Access to non-public [{}] is blocked!", member);
+            LOG.warn("Access to non-public [{}] is blocked!", member);
             return false;
         }
 
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -105,19 +105,25 @@ public class SecurityMemberAccess implements MemberAccess {
     public boolean isAccessible(Map context, Object target, Member member, String propertyName) {
         LOG.debug("Checking access for [target: {}, member: {}, property: {}]", target, member, propertyName);
 
+        final int memberModifiers = member.getModifiers();
+
+        if (!checkPublicMemberAccess(memberModifiers)) {
+            LOG.trace("Access to non-public [{}] is blocked!", member);
+            return false;
+        }
+
+        if (!checkStaticFieldAccess(member, memberModifiers)) {
+            LOG.warn("Access to static field [{}] is blocked!", member);
+            return false;
+        }
+
         if (checkEnumAccess(target, member)) {
             LOG.trace("Allowing access to enum: target [{}], member [{}]", target, member);
             return true;
         }
 
-        final int memberModifiers = member.getModifiers();
-        if (!checkStaticMemberAccess(member, memberModifiers)) {
-            LOG.warn("Access to static [{}] is blocked!", member);
-            return false;
-        }
-
-        if (!checkPublicMemberAccess(memberModifiers)) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -105,19 +105,25 @@ public void restore(Map context, Object target, Member member, String propertyNa
     public boolean isAccessible(Map context, Object target, Member member, String propertyName) {
         LOG.debug("Checking access for [target: {}, member: {}, property: {}]", target, member, propertyName);
 
-        if (checkEnumAccess(target, member)) {
-            LOG.trace("Allowing access to enum: target [{}], member [{}]", target, member);
-            return true;
+        final int memberModifiers = member.getModifiers();
+
+        if (!checkPublicMemberAccess(memberModifiers)) {
+            LOG.trace("Access to non-public [{}] is blocked!", member);
```
lukaszlenart: Should we warn users about that?

JCgH4164838Gh792C124B5: Hello @lukaszlenart .
I'm not sure what @aleksandr-m  thinks, but the log-level for failure due to non-public members might be a "balancing act".  
My understanding is that prior to this the framework would just return false for non-public elements (so downstream processing would determine whether any logging happened in such cases) ?
A _warn_ for this condition might end up being noisy for some apps when migrating apps from 2.5.x to 2.6.  Maybe a level of _debug_ (or even _info_) might be a better balance in this case ?

lukaszlenart: I meant, accessing non-public methods/fields is a visible sign that something is wrong. And even before the change, access to non-publics was blocked so I see no problems with migrations.

JCgH4164838Gh792C124B5: OK ... I think I see your point.  So maybe a warn-level for 2.6 for failing the initial public member check (but if this check was back-ported to 2.5.x it would be trace or debug ...).

lukaszlenart: Ok, but I still don't see a valid case, even in Struts 2.5.x you are not allowed to access non-public methods, you can only switch on/off access to static methods.

yasserzamani: I prefer to warn i.e. a `LOG.warn` statement - it warns user about a misuse in underlying app or about an attack try on it.

aleksandr-m: Let's make it simple. :)

@lukaszlenart 

1. Remove log at this line
2. Change log level to `warn`
3. Leave as it is now

Choose one.

lukaszlenart:
I'm good, LGTM 👍

yasserzamani:
I see supporting for access static methods is deprecated long time ago (see [here](https://github.com/apache/struts/blob/e9d92f5000a827ac83b63aada624946854a77fef/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java#L65)). Let's drop supporting for access to any static (including method, field and etc) in 2.6. wdyt? I myself strongly like it to be more secure and clean code.

lukaszlenart:
@yasserzamani it's not that easy... a lot of app will break, also with did a lot of good work to improve security in case of accessing statics.

yasserzamani:
Hmm you're right :) I explored it is deprecated starting from about 2.3.25. Usually we should wait for 2 or 3 versions for delete. So let's keep it maybe until 2.7. Thanks!

aleksandr-m:
@yasserzamani Access to static fields is pretty convenient, think about constants. I wouldn't remove this, but maybe we can restrict access to static fields based on type (primitives / strings are ok).

yasserzamani:
I don't have a strict idea but just two comments: Sometimes changes to a string, boolean and etc primitive types can also cause DOS and etc attacks. And in above solution ognl blocks user also on access to a primitive type but inside a static class.

Anyway I think it's OK to merge this one for now :) wdyt?

yasserzamani:
No objections/suggestions... let's merge :)

PR #326, opened 2019-02-02
Potential mitigation for WW-5009

JCgH4164838Gh792C124B5:
Potential mitigation for WW-5009
- Add synchronization guard to problematic JSONUtil instances in JSONInterceptor, JSONResult:
  - Guarding the two calls to jsonUtil.serialize() appears sufficient based on WW-5009 interactive reproducer test.
- Should preserve functionality for WW-4034 (custom JSONWriter).

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/DefaultJSONWriter.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/DefaultJSONWriter.java
@@ -60,32 +60,33 @@ public class DefaultJSONWriter implements JSONWriter {
 
     private static final Logger LOG = LogManager.getLogger(DefaultJSONWriter.class);
 
-    private static char[] hex = "0123456789ABCDEF".toCharArray();
+    private static final char[] HEX = "0123456789ABCDEF".toCharArray();
 
     private static final ConcurrentMap<Class<?>, BeanInfo> BEAN_INFO_CACHE_IGNORE_HIERARCHY = new ConcurrentHashMap<>();
     private static final ConcurrentMap<Class<?>, BeanInfo> BEAN_INFO_CACHE = new ConcurrentHashMap<>();
 
-    private StringBuilder buf = new StringBuilder();
-    private Stack<Object> stack = new Stack<>();
     private boolean ignoreHierarchy = true;
-    private Object root;
-    private boolean buildExpr = true;
-    private String exprStack = "";
-    private Collection<Pattern> excludeProperties;
-    private Collection<Pattern> includeProperties;
     private DateFormat formatter;
     private boolean enumAsBean = ENUM_AS_BEAN_DEFAULT;
-    private boolean excludeNullProperties;
     private boolean cacheBeanInfo = true;
     private boolean excludeProxyProperties;
 
+
+    /**
+     * Set whether to exclude proxy properties or not.
+     * 
+     * @param excludeProxyProperties
+     */
     @Inject(value = JSONConstants.RESULT_EXCLUDE_PROXY_PROPERTIES, required = false)
     public void setExcludeProxyProperties(String excludeProxyProperties) {
         setExcludeProxyProperties(Boolean.parseBoolean(excludeProxyProperties));
     }
 
     /**
+     * Perform a write (serialize) of the given object into JSON.
+     * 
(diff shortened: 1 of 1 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

Due to what appear to be multiple thread-safety issues with DefaultJSONWriter (when a single JSONUtil instance and JSONWriter instance is used by the interceptor instance), there doesn't seem to be much that could be done without significant refactoring.

This PR's solution attempts to isolate the entry points for JSONUtil instances that could present concurrency issues for the interceptor.  Unfortunately it requires some synchronization (resulting in reduced concurrency), but it should have the advantage of also protecting custom JSONWriter usage too.

So far the only test available is an interactive test project by the user who opened the JIRA ticket, but this mitigation appears to survive the test (where the JSON plugin from 2.5.20 does not).

JCgH4164838Gh792C124B5:
Hello @aleksandr-m and @yasserzamani .
I'm not in favour of a _synchronization_ solution either, but it is always a fall-back in the event a re-factoring solution isn't feasible for one reason or another.
This was a quick initial mitigation for the current state (which from what I can see can result in intermittent inconsistent processing, even without seeing `EmptyStackException`).
I'll make an attempt for a minimalist refactoring to try and isolate things and see what both of you think (unless you're already working on something ...).

JCgH4164838Gh792C124B5:
Hello again.

The previous commit has been amended to achieve the following:
- Eliminated the synchronization usage.
- Re-factored DefaultJSONWriter in such a way that it should now be thread-safe (preserving the bodies of preexisting methods for the most part, moving write-related processing to use per-call handling).
- In theory this should eliminate the possibility of thread-related contention, while still allowing for a custom JSONWriter (but those authors would have to ensure thread-safety).

The unit tests all passed successfully and using the reproducer project I was not able to break this implementation using multiple tabs/browsers accessing at the same time (as described by the user who reported the issue).

There's probably room-for-improvement, but it avoids synchronization.  Let me know what you think?

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

I've taken a look at @aleksandr-m 's solution in PR#328, and it is pretty cool/clever.  It preserves the capability of a custom JSONWriter and avoids thread-collisions by generating a new JSONUtil instance on every call to intercept().

I copied his implementation locally and it survived the interactive reproducer testing (for me at least with 4+ tabs in one browser, 4+ tabs in different browser, multiple retries).  :)

The only minor downside I can see is the overhead from the JSONUtil construction/injection on every intercept call (unless there's a delayed/lazy-initialization benefit I'm overlooking, the configuration values for the JSONUtil instance won't change after the interceptor is constructed and initialized).

Looking at the implementation is looks like both PRs could actually be implemented at the same time (to gain the advantages of both).  @aleksandr-m 's solution provides isolation at the interceptor-level, while this one provides isolation at the writer-level (plus some improved comments and more accurate method-naming).

In the end it will be your decision and @aleksandr-m 's decision as to how you would like to proceed.

Thanks for taking a look.  :)

yasserzamani:
I personally would keep it simple to understand and more readable - Struts already performs a lot of injections on each request I think, and we don't have any user's objection about it's performance. Nowadays hardwares are too fast and powerful, let vote to humans and keep it simple to understand even with some overheads ;) Additionally I would minimize changes to 2.5.x as much as possible to achieve a stable release as simple and sooner as possible because it is going to go under maintenance mode.

At bottom, these are my not-complete thoughts - please wait for @apache/struts-committers if they vote to merge both :)

Thanks!

aleksandr-m:
I would keep it simple, especially for 2.5.x branch. As for the 2.6, it would be better to create a new json plugin instead of patching the current one.

JCgH4164838Gh792C124B5:
Both of your comments above make sense to me ...  minimize the changes for 2.5.x, consider a new JSON plugin for the 2.6 branch sometime in the future.  Leaving this PR aside for now should minimize the risk of introducing an unintentional bug to 2.5.x.

The important thing is that the concurrency bug is resolved by the other PR ...
... and we learned the DefaultJSONWriter in 2.5.x isn't thread-safe (so no one should try and use it directly without taking that into consideration).

PR #327, opened 2019-02-03
[WW-5004] Fixes changing "struts.ognl.allowStaticMethodAccess"

lukaszlenart:
This PR relates to [WW-5004](https://issues.apache.org/jira/browse/WW-5004) to allow change the flag to a non-default value.

See #317  for discussion

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
@@ -25,23 +25,29 @@ import com.opensymphony.xwork2.XWorkConstants;
 import com.opensymphony.xwork2.conversion.NullHandler;
 import com.opensymphony.xwork2.conversion.impl.XWorkConverter;
 import com.opensymphony.xwork2.inject.Container;
+import com.opensymphony.xwork2.inject.EarlyInitializable;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.ognl.accessor.CompoundRootAccessor;
+import com.opensymphony.xwork2.ognl.accessor.ParameterPropertyAccessor;
 import com.opensymphony.xwork2.util.CompoundRoot;
 import com.opensymphony.xwork2.util.ValueStack;
 import com.opensymphony.xwork2.util.ValueStackFactory;
 import ognl.MethodAccessor;
 import ognl.OgnlRuntime;
 import ognl.PropertyAccessor;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.apache.logging.log4j.message.SimpleMessage;
+import org.apache.struts.util.PropertyMessageResources;
 
 import java.util.Set;
 
 /**
  * Creates an Ognl value stack
  */
-public class OgnlValueStackFactory implements ValueStackFactory {
+public class OgnlValueStackFactory implements ValueStackFactory, EarlyInitializable {
     
     private static final Logger LOG = LogManager.getLogger(OgnlValueStackFactory.class);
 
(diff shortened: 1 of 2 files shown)


lukaszlenart:
@yasserzamani you are right, there is no issue. I double checked against 2.5.20 and 2.5.18 and everything is ok. I will apply only those refactors to Struts 2.6 in another PR.

Thanks for keeping your hand on the pulse!

lukaszlenart:
After moving my changes to 2.6, the positive test case stopped working ... so we have an issue here :)

yasserzamani:
So I think it's an another issue not related to WW-5004. Let register another ticket targeting 2.6 and focus on a stable 2.5 release instead, then we can probe and fix it when we found some time for it. wdyt? Or we can force us for it's resolution prior to next 2.5 release because we afraid probing into it will reveal something to fix in 2.5 as well?

lukaszlenart:
There is no need to push this now, I will work on that in meantime to fix that in 2.6 - I was just using a junit test to confirm this but maybe it will work in a proper app.

JCgH4164838Gh792C124B5:
If the positive test case in 2.6 failed it might be worth reviewing the JUnit positive test case to make sure there isn't a flaw in the test case logic ... to avoid missing a boundary case that needs correction for the 2.5.x branch.

yasserzamani:
I probed and discovered that in both 2.5.x and 2.6 branches, the `loadButAdd` can add but can't replace:

- passes
```java
loadButAdd(String.class, "testWW_5004", "true");
assertTrue(BooleanUtils.toBoolean(container.getInstance(String.class, "testWW_5004")));
````
- fails
```java
loadButAdd(String.class, "struts.ognl.allowStaticMethodAccess", "true");
assertTrue(BooleanUtils.toBoolean(container.getInstance(String.class, "struts.ognl.allowStaticMethodAccess")));
```

So above test case results aren't valid.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

Are you sure that the behaviour for `loadButAdd()` isn't working as expected ?

It seems that `loadButAdd()` depends upon `ContainerBuilder`, which doesn't permit "duplicates" by default (i.e. if a value is already defined, a new value cannot replace it).  Might that be to help protect against an unintentional (or malicious) attempt to change previously configured values for a container instance ?

The result for the test listed as _fails_ above might be expected due to the key _struts.ognl.allowStaticMethodAccess_ being previously set false by the test setup configuration.  If the test is modified so that the ContainerBuilder involved in the test is set to allow duplicates, then the 2nd test might work ?

yasserzamani:
Hi there,

Not sure if unit tests or method `loadButAdd` are intended to be different, but in a real run, users can  override beans and constant beans via `struts.xml` and also plugins can. Have you seen any piece of code that prohibits duplicates?

That day, I also tried to debug and comment here the reason, but Struts internal DI logics baffled me with somehow a strange behaviour! I saw a lot of cycling recursive duplicate `ContainerBuilder` and `ContainerImpl` beans creations and injections! Anyway, at end, `ContainerImpl` cannot find the new added constant in it's factories field, returns null and test fails :(

I think it's nice to prepare some diagram to documenting Struts internal DI cycles and injections :)

Regards!

JCgH4164838Gh792C124B5:
Hello.

It's good to hear that I'm not the only one who finds the Struts internal DI logic difficult to follow on occasion.  :).

I didn't realize in your test that the lookup returned null (instead of "false", which is what I was thinking).  What you described almost seems like the factory field got injected into the "wrong container" (somehow during the recursion).  

I went and re-checked manual setting of `allowStaticMethodAccess` in the config (using 2.5.21 and modification to one of the Showcase JSPs) and that still seems to work as expected (turning it on and off).

Maybe you've found a corner-case for the DI logic ?  A flow-diagram of the "normal" DI processing would definitely be nice.  :)

yasserzamani:
Yes it looks like a corner-case has emerged in unit tests, however, previously I also had seen similar behaviors in a real app debug. And I forgot to mention, that day I saw the method `loadButAdd` already has working usages. It seems the issue emerges when we pass a `name` or when we want to replace constant (current usages aren't constant and don't have `name`).

Regarding diagram, I wonder with which methodology we can extract one. Maybe we can use AOP and log all bean creations, injections and etc then analyze the log of a minimal app.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

Thanks for your hard work looking into these things.  :)  :)

Your idea for a diagram creation seems reasonable to me (I couldn't think of an alternative, other than maybe just a normal debug/trace log for a simple Struts app).

I had a chance to try replicating your tests above (in 2.5.21 only, adding the equivalent two tests to a new test in SetPropertiesTest).  I found the same pass/fail condition, but in my test attempt:
- container.getInstance(String.class, "struts.ognl.allowStaticMethodAccess")
was able to find a factory instance (not a null) using that key.  So same failure, but different cause.

Debugging/stepping through the code I think I found the reason for the behaviour within the test case.  The sequence of events was:
1) Test case calls `loadButAdd()` for the key "struts.ognl.allowStaticMethodAccess".
2) The container instance sets a non-constant factory instance for key "struts.ognl.allowStaticMethodAccess" with a String value of "true".  (in both factories and singletonFactories).
3) A little later on the call sequence to retrieve the container from the `ConfigurationManager` (via `XWorkTestCaseHelper`) results in a reload of the configuration.
4) The XWorkTestCaseHelper uses a `DefaultConfiguration` for its configuration, which has a whole series of constants defined for it, one of which is: "struts.ognl.allowStaticMethodAccess" with constant value "false".
5) When the constants are loaded, the constant value in step 4 overwrites the value set in step 2.
_Note_: It appears the internal logic ensures that any configuration constants override values manually set on the container (which makes sense, ensuring configured values always take precedence).  
6) When the test case attempts to retrieve the value from the container it gets back the value from steps 4-5, so it is "false".
7) Test case failed since "false" was not true when converted.

Assuming I didn't mess something up in my local configuration (running under Java 8u201), the unit test seems to be working properly/failing for a good reason, given that a pre-existing constant value for struts.ognl.allowStaticMethodAccess is defined in DefaultConfiguration.

Let me know if you see a flaw in the above logic, and if your unit test still behaves differently when debugging locally (maybe something strange happening with the local project/IDE/JDK ?),

yasserzamani:
With thanks to Sunday free times :) IntelliJ IDEA's vigor and "conditional breakpoints", after a few hours effort I found the villain, `DefaultConfiguration.java`. Search for `props.setConstants(` in it - if a provider don't use `props` for properties and uses `.factory` method directly (like `loadButAdd`), that `props.setConstants(` replaces that property if other providers have defined it using `props`. I fixed this at https://github.com/apache/struts/commit/a15c12a05109da5ad48efb77921f4178ceab16fa#diff-eb1418aa89ec4a08d5a290387976b520R82.

Then I also learnt we can have cleanups for your previous PRs like https://github.com/apache/struts/commit/a15c12a05109da5ad48efb77921f4178ceab16fa#diff-fb5059fc7e863b3fc70d933d7cbefad2R1236 and delete those xml overheads in tests.

Thanks a lot @JCgH4164838Gh792C124B5 for everything , thanks a lot @lukaszlenart for `loadButAdd` :)

yasserzamani:
Similar thing cherry picked then manually merged/completed then pushed into master.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

Thanks for your efforts in researching this and finding a cleaner solution for the unit tests.  :) :)

One thing that the "test xmls" did do (I think) was provide an indirect confirmation that the configuration flags worked as expected when loaded from an xml configuration (similar to a real-user config).

Do you think there would be any value in having some of those old tests added into `XmlConfigurationProviderTest` (or even back into `OgnlUtilTest`) ?  Maybe just the non-default flag states (_allowStaticMethodAccess_ true, _devMode_ true, _allowStaticFieldAccess_ true for 2.6) in a single .xml file and a check that the resulting container has the expected values after loading that config ?

My only thought is such a test might help catch an accidental impact to the "important flag" behaviours with respect to the xml configuration ...

PR #328, opened 2019-02-03
WW-5009 EmptyStackException in JSON plugin due to concurrency

aleksandr-m:
@yasserzamani I know that you've assigned that issue to yourself, but I think it is the most simple solution that we can merge in 2.5.x branch. Can you take a look.


JCgH4164838Gh792C124B5:
Hello @aleksandr-m .

Your solution appears to be cool/elegant for sure!  :)
Replicating it locally there were no failures when using the manual reproducer test (for me anyway).

Good work!

aleksandr-m:
@yasserzamani 

> By the way, do you have any objection to import same one into 2.6?

Sure. #329 

@JCgH4164838Gh792C124B5 

> Replicating it locally there were no failures when using the manual reproducer test (for me anyway).

Thanks for testing this.

PR #334, opened 2019-02-13
Fix for default (un-configured) allowStaticFieldAccess state, Showcase Application in 2.6.x

JCgH4164838Gh792C124B5:
Fix for default un-configured `allowStaticFieldAccess` state:
1) Ensure `OgnlValueStackFactory` returns true for allowStaticFieldAccess when not configured (consistency - same as `OgnlUtil` does).
2) Add test configuration file with no allowStaticXXX flags present (forces default values).
3) Add test to `OgnlValueStackTest` for the no allowStaticXXX flags present (verifies default values).

Minor config fixes for the Showcase Application in 2.6.x:
1) File upload examples.  Multiple file upload missing result page (list and array), added missing JSP.  Added missing input results in configuration.
2) File upload examples.  Fixed NPE for multiple file upload (array version) when no files selected.
3) Add manual test for public constant access via expression (to If tag JSP).


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

While attempting to apply some fixes for the 2.6.x Showcase application, I found that the manual public static field access test wasn't behaving properly.  
This led to discovering an inconsistency in the behaviour for allowStaticFieldAccess when the flag is not explicitly set in configuration (as is the case for the Showcase application).
This PR aims to fix that issue, provide a unit test to confirm the fix, and also add the Showcase application fixes (as they were all interrelated).

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -125,11 +125,17 @@ protected boolean containerAllowsStaticMethodAccess() {
 
     /**
      * Retrieve allowStaticFieldAccess state from the container (allows for lazy fetching)
+     * Note: Defaults to true when no state is defined in the container (same default behaviour as OgnlUtil)
      * 
      * @return
      */
     protected boolean containerAllowsStaticFieldAccess() {
-        return BooleanUtils.toBoolean(container.getInstance(String.class, StrutsConstants.STRUTS_ALLOW_STATIC_FIELD_ACCESS));
```
yasserzamani: I investigated why `containerAllowsStaticMethodAccess()` doesn't need such fix and found [this in default.properties](https://github.com/apache/struts/blob/9a14323a6a6414f845f240c8463503e0c940cec6/core/src/main/resources/org/apache/struts2/default.properties#L207)! I think we should keep same conduct as static methods for static fields.

lukaszlenart: Maybe we should drop that file and move everything into `struts-default.xml`?

apps/showcase/src/main/java/org/apache/struts2/showcase/fileupload/MultipleFileUploadUsingArrayAction.java
```diff
@@ -32,9 +32,9 @@
  */
 public class MultipleFileUploadUsingArrayAction extends ActionSupport {
 
-	private File[] uploads;
-	private String[] uploadFileNames;
-	private String[] uploadContentTypes;
+	private File[] uploads = new File[0];
```
aleksandr-m: This looks odd. Why new?

JCgH4164838Gh792C124B5: Hello @aleksandr-m .
The new initializers were the quickest fix (and equivalent to what is done in  `MultipleFileUploadUsingListAction.java`) for a boundary-condition bug in processing the action. 

Without allocated fields (via new in initializers), the following steps resulted in an HTTP 500 (due to NPE in `upload()`):
1) Choose Multiple File Upload (Array) from File menu.
2) Click on Submit (without adding any files to the upload).

With this fix both types of multiple file upload attempts can process empty payloads without failing.  The upload() methods of both classes could be changed as an alternative, but this was the least-invasive change.

yasserzamani:
I think, to answer this, we should firstly know the philosophy behind this file. Maybe it is designed as a container for primitive constants to make struts.xml shorter in length? - it contains a lot of consultants.

It seems it's not a bad design because log4j also supports both xml and properties files.

Wdyt?

lukaszlenart:
`default.properties` and `struts-default.xml` are basic configuration files that will set up an initial behaviour of the framework. User has no way to disable them or so, the only option is to define their own config files - `struts.xml` or `struts.properties` - to override settings provided by the prior.

From my point of view, having the whole configuration in one place is better and more maintainable.  BTW. we can extract some parts from `struts-default.xml` and put in other XML files and use `include` to tie them together.

And I think this file is somehow a leftover from WebWork days, I found only this reference to it
https://issues.apache.org/jira/browse/WW-1666

Having that said, I don't have a strong opinion that merging `default.properties` into `struts-default.xml` is a good idea :)

yasserzamani:
James, with thanks! Now you can revert following change to previous state after the commit :)

> * * @return */ protected boolean containerAllowsStaticFieldAccess() { - return BooleanUtils.toBoolean(container.getInstance(String.class, StrutsConstants.STRUTS_ALLOW_STATIC_FIELD_ACCESS));

JCgH4164838Gh792C124B5:
Hello Again.

The PR has been updated as per @yasserzamani 's request (with adjustment to the unit tests to compensate for it).  Interactively (and with the unit tests) everything seems to still work as expected.

As previously noted, in any situation where a container/factory doesn't have the static field access defined (e.g. one that uses `DefaultConfiguration`), the container will still return false for `containerAllowsStaticFieldAccess()`.  This might not happen in real processing, though.  :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -125,6 +125,7 @@ protected boolean containerAllowsStaticMethodAccess() {
 
     /**
      * Retrieve allowStaticFieldAccess state from the container (allows for lazy fetching)
+     * Note: Defaults to true when no state is defined in the container (same default behaviour as OgnlUtil)
```
yasserzamani: Now this should be vanished  as well :)

yasserzamani:
@JCgH4164838Gh792C124B5 I understand your concern but if we want this for `allowStaticField` then we should do it for all configurations if they're not defined! I don't think it easy as well as don't see it in high priority for Struts and even not sure if it's needed for better design at all :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
JCgH4164838Gh792C124B5: Sorry ... I missed that.  The latest commit removes the (now) unwanted comment line.  Thanks for catching that.  :)

JCgH4164838Gh792C124B5:
Interesting ... the Java 8 coveralls job failed this time even though it was a one-line comment change:
"Report submission to Coveralls API failed with HTTP status 422: Unprocessable Entity (Couldn't find a repository matching this job.".
The actual Java 8 build succeeded (according to the raw report), so maybe something intermittent with the CI environment or job ?

yasserzamani:
Never mind...looks like a temporary conflict. We will dive in if persisted with next others.

lukaszlenart:
Is this ready?

yasserzamani:
Looks no :)

Hi @JCgH4164838Gh792C124B5 , your recent merge with master has dropped your tests which can be added again via a new commit.

Also, I think the redundant file `xwork-test-allowstatic-notset.xml` now can be dropped and avoided like below:

```java
        loadConfigurationProviders(new StubConfigurationProvider() {
            @Override
            public void register(ContainerBuilder builder,
                                 LocatableProperties props) throws ConfigurationException {
                if(props.contanis(StrutsConstants.STRUTS_ALLOW_STATIC_FIELD_ACCESS)) {
                    props.remove(StrutsConstants.STRUTS_ALLOW_STATIC_FIELD_ACCESS);
                }
            }
        });
```

JCgH4164838Gh792C124B5:
Hmm ... GitHub indicated there was a merge conflict when I checked about 4 days ago so I used the builtin conflict resolution tool (via the button) ... that might have been a mistake.  :)

Changes to master between 18 days ago and 4 days ago resulted in the conflict, and although the conflict got resolved I didn't realize it dropped something we wanted to keep.  ;)

I'll try to resynch locally and untangle where it went wrong ... and deal with xwork-test-allowstatic-notset.xml.  Give me a day or so and I'll hopefully have it sorted-out and an updated commit with it fixed.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

The latest commit has recovered the lost tests, as well as added a few more to cover more state combinations.  The reload logic has been replaced by your improved logic (from master, as suggested above) and the now unnecessary .xml file was dropped as well.

Hopefully I didn't miss anything.  :)

Let me know what you think, and if any more changes are needed for the PR.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -980,6 +1001,272 @@ public void testWarnAboutInvalidProperties() {
...
 
+   /**
+     * Test a default OgnlValueStackFactory and OgnlValueStack generated by it
+     * when a default configuration is used.
+     */
+    public void testOgnlValueStackFromOgnlValueStackFactoryDefaultConfig() {
+        OgnlValueStackFactory ognlValueStackFactory = (OgnlValueStackFactory)
+                container.getInstance(ValueStackFactory.class, Container.DEFAULT_NAME);
+
+        OgnlValueStack ognlValueStack = (OgnlValueStack) ognlValueStackFactory.createValueStack();
```
yasserzamani: I see there is already `OgnlValueStackTest.createValueStack` available method which is used in other tests. I think it's a good practice to keep current existing codes conduct and use it here instead.

P.S. It seems `OgnlValueStackTest.createValueStack` needs a fix about not used `allowStaticFieldAccess` parameter. Please feel free also if any improvement to it was needed to support your newly added tests :) thanks in advance!

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -980,6 +1001,272 @@ public void testWarnAboutInvalidProperties() {
...
+            reloadTestContainerConfigurationNoStaticAccessFlagsSet();
+            ognlValueStackFactory.setContainer(container);
+            ognlValueStackFactory.setTextProvider(dummyTextProvider);
+            ognlValueStackFactory.setXWorkConverter(xWorkConverter);
+        }
+        catch (Exception ex) {
+            fail("Unable to reload container configuration and set valuestackfactory - exception: " + ex);
+        }
+
+        OgnlValueStack ognlValueStack = (OgnlValueStack) ognlValueStackFactory.createValueStack();
```
yasserzamani: Same here. Additionally I feel above object creations (`new ...`s) and `set...`s shouldn't be needed. Calling `reload...` then `OgnlValueStackTest.createValueStack` should be enough.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -980,6 +1001,272 @@ public void testWarnAboutInvalidProperties() {
...
+            reloadTestContainerConfiguration(false, false, false);
+            ognlValueStackFactory.setContainer(container);
+            ognlValueStackFactory.setTextProvider(dummyTextProvider);
+            ognlValueStackFactory.setXWorkConverter(xWorkConverter);
+        }
+        catch (Exception ex) {
+            fail("Unable to reload container configuration and set valuestackfactory - exception: " + ex);
+        }
+
+        OgnlValueStack ognlValueStack = (OgnlValueStack) ognlValueStackFactory.createValueStack();
```
yasserzamani: (same here as previous similar comment)

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -980,6 +1001,272 @@ public void testWarnAboutInvalidProperties() {
...
+            reloadTestContainerConfiguration(false, true, true);
+            ognlValueStackFactory.setContainer(container);
+            ognlValueStackFactory.setTextProvider(dummyTextProvider);
+            ognlValueStackFactory.setXWorkConverter(xWorkConverter);
+        }
+        catch (Exception ex) {
+            fail("Unable to reload container configuration and set valuestackfactory - exception: " + ex);
+        }
+
+        OgnlValueStack ognlValueStack = (OgnlValueStack) ognlValueStackFactory.createValueStack();
```
yasserzamani: (same here as previous similar comment)

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -980,6 +1001,272 @@ public void testWarnAboutInvalidProperties() {
...
+            reloadTestContainerConfiguration(false, true, false);
+            ognlValueStackFactory.setContainer(container);
+            ognlValueStackFactory.setTextProvider(dummyTextProvider);
+            ognlValueStackFactory.setXWorkConverter(xWorkConverter);
+        }
+        catch (Exception ex) {
+            fail("Unable to reload container configuration and set valuestackfactory - exception: " + ex);
+        }
+
+        OgnlValueStack ognlValueStack = (OgnlValueStack) ognlValueStackFactory.createValueStack();
```
yasserzamani: (same here as previous similar comment)

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
JCgH4164838Gh792C124B5: Hello @yasserzamani .

I fixed the issue with `createValueStack()` not setting the `allowStaticFieldAccess` parameter.

However since the other tests were intentionally using an (_intentionally_) different creation strategy using `OgnlValueStackFactory` directly to generate the `OgnlValueStack` instance for testing I couldn't see a good way to modify the existing `createValueStack()` method.

Instead I refactored some of the repetitive code into two separate generator methods for `OgnlValueStackFactory` instances:
- `createDefaultValueStackFactory()`, which generates based on the container configuration (from `setUp()`)
- `createRawValueStackFactory()`, which generates from raw factories (independent of the configuration from `setUp()` to better isolate the state testing of the flags) with boolean control flags to set the state (including the "clear-static-flags-from-properties").

It makes the overall code cleaner and promotes re-use (one of which was to add one additional test with flags set to match the default-from-container-state, to verify both behave the same).

Hopefully the changes look better and make sense ?

yasserzamani:
Thanks, yes you're right. Current codes have different creation strategy
which cannot be applied on what you're trying to test.

By the way, I myself also pushed what I was trying to say. Please feel
free to inform me if they break something in your logic if I failed to
understand something. Elsewhere, it's LGTM to squash and merge.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

I read your changes, synchronized my branch and tried things out ... everything still seems to run/test as expected.  Using the `Boolean` vs. `boolean` and null checks was clever ... thanks for that.  :)

The only minor reservation I had was that the `OgnlValueStackFactory` instance is now coming from the `Container` (configured by `setUp()` and the environment) ... but what I was trying before was _probably "overkill"_ to create and use an isolated/raw instance.  The reloaded/overridden properties mechanism should be sufficient in the "real world" anyway, and the code is much more compact.

So ... you changes LGTM.  Thanks for your time, consideration and effort with the code.  :) :)

yasserzamani:
Cool, you're welcome! yes I also removed `devMode`. IntelliJ IDEA is very clever and warned it's always false so I removed as it's not needed for now (it's always a good practice to consider the IDE warnings on new code lines). I also renamed `create...` to `get...` and updated corresponding javadocs as it's not further similar to `createValueStack` because it doesn't `new...` but gets from container. (all of these are at 4340d50 which you may want to see if similar are applicable to this PR's peer on 2.5)

At bottom thanks again for tests. I'm merging this :)

P.S.  I am just curious to know why you double tests with `final` variation. Despite more tests mean more good, I personally think it wasn't needed to bother yourself on this because `final` is not about access logic and you're not going to test if `final`s are settable via OGNL so you could reduce complexity to half :)

JCgH4164838Gh792C124B5:
Hello @yasserzamani .

I'll try to take a look and see what of [https://github.com/apache/struts/commit/4340d50d222888cc0450838001dd247cf7c84b65](https://github.com/apache/struts/commit/4340d50d222888cc0450838001dd247cf7c84b65) might apply to the changes from the peer 2.5 PR, when I have a chance.

In terms of the `final` variations in the tests, there had been discussion about possibly limiting access to "constant" fields at one point.  So ... it seemed prudent to supply tests that confirm the current behaviour (ability to access public fields, both "constant" and "non-constant", but ready for the future).
Also the tests were intended as an additional sanity check to guard against a future OGNL behaviour change (_unlikely but possible_) where the `final` field access behaviour was impacted, even for reads.  I figured those two reasons combined were a good enough reason for the extra tests/complexity.  :)

yasserzamani:
Cool! you're right. Now I understand they put OGNL behavior about constant fields under monitor. Thanks for clarification.

PR #337, opened 2019-02-21
Upgrade SLF4J to 1.7.26 (WW-5023)

JCgH4164838Gh792C124B5:
Update SLF4J in Struts 2.6 to the latest 1.7.x version.
- update library version in pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1069,13 +1069,11 @@
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
                 <version>1.7.26</version>
-                <optional>true</optional>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-simple</artifactId>
                 <version>1.7.26</version>
-                <optional>true</optional>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -1069,11 +1069,13 @@
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
                 <version>1.7.26</version>
+                <optional>true</optional>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-simple</artifactId>
                 <version>1.7.26</version>
+                <optional>true</optional>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -1068,12 +1068,12 @@
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
-                <version>1.7.25</version>
+                <version>1.7.26</version>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-simple</artifactId>
-                <version>1.7.25</version>
+                <version>1.7.26</version>
             </dependency>
 
             <dependency>


aleksandr-m:
Why do we need slf4j at all?

JCgH4164838Gh792C124B5:
Seems to be an optional dependency, first appeared with WW-3959 back in Struts 2.3.12.

My best guess it's present for convenience/flexibility in "out-of-the-box" logging for the Struts release (provides some additional flexibility to use SLF4J and/or commons-logging ) ?

lukaszlenart:
We do not use it directly but many other dependencies depend on it. In 2.6 we will use Log4j2 as the only logging layer which can be bridged with other logging facades. If this is really an optional dependency we should mark it as such.

JCgH4164838Gh792C124B5:
Updated the commit to mark SLF4J as optional in the project pom.xml.  It is already listed as an optional dependency in the Core pom.xml (the only other pom that I could find referencing it).

lukaszlenart:
I always have a dilemma if we should mark dependencies as optional here or rather leave that to a particular POM as here we define just a set of available dependencies that can be used across the project.

wdyt?

rgielen:
My take on this: if there is any code referencing this optional dependency, leave it with scope optional. If there are no code references, throw it out since it does not help anyone (only documentation does).

Looking on master, there seems to be no real ref to slf4j, thus IMO there is no need to keep it in the core pom and in consequence also remove is from top level pom DM section.

That said, it **does** make sense to add it to the **BOM** as it is an optional feature for which we then provide sensible defaults, once used by an application developer

JCgH4164838Gh792C124B5:
From the discussion there appear to be 3 options:
1) Keep the dependency as optional (top-level and core, as per this PR).
2) Remove the optional dependency flag from the top-level (revert project pom to previous state).
3) Completely remove the SLF4J dependency from 2.6.  For this option it would probably be best to close this PR and open a new JIRA for 2.6 to remove the SLF4J dependency.

Let me know what you think/prefer.

lukaszlenart:
Less dependencies is better, but it would be good to provide the latest version if any of our first level dependency depends on it.

In other worlds, if `mvn dependency:tree` shows that the project is using it, it may sense to have this dependency in the BOM.

JCgH4164838Gh792C124B5:
I tried _commenting out the SLF4J dependencies_ (_artifactIds_: `slf4j-api`, `slf4j-simple`) from the parent and core pom.xmls and see what happened with both a maven build attempt and mvn dependency:tree execution.  The results were:
1) Maven build failed at Tiles Plugin.
2) Dependency tree indicated dependencies on SLF4J:
- Tiles-Plugin
- Showcase (due to tiles plugin)
- Rest Showcase (due to jwebunit)
- CDI-Plugin (due to weld)
- Tiles Portlet-Plugin (due to tiles-plugin)

It seems that the overall Struts 2.6 project still has dependencies on SLF4J.  Someone with more Maven experience might want to confirm the interpretation above.

I tried putting the SLF4J dependencies in the BOM's pom.xml but it didn't seem to help (entirely possible with my Maven experience that I did it incorrectly...).

Thoughts ?

lukaszlenart:
Ah... right, Tiles is the main reason why we specify SL4J in the pom, it means it cannot be marked as `optional`. Thanks a lot for digging into to this :)

JCgH4164838Gh792C124B5:
No problem.  :)
The commit has been updated to remove the optional flag state from the project pom.xml, as per above discussion and comments.  It should now be back to the original state, just with the latest SLF4J version number in place.

lukaszlenart:
osm! LGTM 👍

PR #339, opened 2019-03-01
Taking action invocation out of synchronized block on TokenSessionStoreInterceptor to reduce contention on the session id

lunalisk:
We recently ran into an issue with the TokenSessionStoreInterceptor due to it's synchronized block holding until the action invocation returns.

If a different action requires to check something related with the same session key, then this action must wait until the Interceptor releases the lock.

We believe the invocation should not be part of the synchronized block

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/TokenSessionStoreInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/TokenSessionStoreInterceptor.java
@@ -108,8 +108,13 @@ public class TokenSessionStoreInterceptor extends TokenInterceptor {
             if (!TokenHelper.validToken()) {
                 return handleInvalidToken(invocation);
             }
-            return handleValidToken(invocation);
+            // we know the token name and token must be there
+            String key = TokenHelper.getTokenName();
+            String token = TokenHelper.getToken(key);
+            String sessionTokenName = TokenHelper.buildTokenSessionAttributeName(key);
+            InvocationSessionStore.storeInvocation(sessionTokenName, token, invocation);
         }
+        return handleValidToken(invocation);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


JCgH4164838Gh792C124B5:
Hello @lunalisk .

It might not be safe to move the invocation outside of the synchronized block within the interceptor.  The change in this PR appears to allow for _potential concurrent/duplicate submits_ from the form with the valid token component.

The API docs say the token-mechanism ensures one-request-per-token, but someone from the Apache Struts Team will have to evaluate this PR and comment further.

A more complex solution separating the session-related synchronization and the token-related synchronization might be needed to achieve your desired result.  We'll see what they say.    :)

yasserzamani:
These are historical backgrounds to study for who would cracking at this PR:

- The closed issue [WW-3582 - Token Interceptor is holding HttpSession lock which can trigger deadlocks](https://issues.apache.org/jira/browse/WW-3582)
- The closed issue [WW-3865 - TokenSesion double submit sends a blank page to ie and stacktrace on server](https://issues.apache.org/jira/browse/WW-3865)
- The discussion [TokenInterceptor locking HTTP session for entire action invocation?](http://markmail.org/thread/ogal3uaekesb6wby)
- The open issue [WW-4409 - Modify the TokenInterceptor not to lock the session object while handling and invalid token](https://issues.apache.org/jira/browse/WW-4409)
- The commit [Solves WW-3582 - moves in call to handleValidToken() method outside of lock on session object](https://github.com/apache/struts/commit/f713cfc8372d63c284f84362f6806798f1abdbeb)
- The commit [WW-3865 rollbacks changes to TokenInterceptor applied with WW-3582 and adds new method handleToken() to allow in different way lock session object](https://github.com/apache/struts/commit/b83b8586e0aa7eda6a4175ee449f9a926431d831)

PR #340, opened 2019-03-06
WW-5024 ActionMappingParametersInterceptor double wrapping

rhollencamp:
objects are wrapped in a Parameter twice when injected into the action
context parameters

Refs [WW-5024](https://issues.apache.org/jira/browse/WW-5024)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
@@ -100,12 +100,12 @@ public class ActionMappingParametersInterceptor extends ParametersInterceptor {
      *                  </p>
      */
     @Override
-    protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
+    protected void addParametersToContext(ActionContext ac, Map<String, Parameter> newParams) {
         HttpParameters previousParams = ac.getParameters();
 
-        // TODO can we just change the method signature to Map<String, Parameter>?
+        // have to unwrap from Parameter to Object since HttpParameters.Builder will re-wrap
         Map<String, Object> unwrappedParams = newParams.entrySet().stream().collect(Collectors.toMap(
-            Map.Entry::getKey, e -> ((Parameter) e.getValue()).getObject()));
+            Map.Entry::getKey, e -> e.getValue().getObject()));
 
         HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(unwrappedParams);
 
--- a/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
@@ -22,11 +22,9 @@ import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.interceptor.ParametersInterceptor;
 import org.apache.struts2.ServletActionContext;
 import org.apache.struts2.dispatcher.HttpParameters;
-import org.apache.struts2.dispatcher.Parameter;
 import org.apache.struts2.dispatcher.mapper.ActionMapping;
 
 import java.util.Map;
-import java.util.stream.Collectors;
 
 /**
  * <!-- START SNIPPET: description -->
@@ -100,14 +98,9 @@ public class ActionMappingParametersInterceptor extends ParametersInterceptor {
      *                  </p>
      */
     @Override
-    protected void addParametersToContext(ActionContext ac, Map<String, Parameter> newParams) {
+    protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
         HttpParameters previousParams = ac.getParameters();
-
-        // have to unwrap from Parameter to Object since HttpParameters.Builder will re-wrap
-        Map<String, Object> unwrappedParams = newParams.entrySet().stream().collect(Collectors.toMap(
-            Map.Entry::getKey, e -> e.getValue().getObject()));
-
-        HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(unwrappedParams);
+        HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(newParams);
 
         ac.setParameters(combinedParams.build());
     }
--- a/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
@@ -22,9 +22,11 @@ import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.interceptor.ParametersInterceptor;
 import org.apache.struts2.ServletActionContext;
 import org.apache.struts2.dispatcher.HttpParameters;
+import org.apache.struts2.dispatcher.Parameter;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
```diff
@@ -100,7 +102,12 @@ protected HttpParameters retrieveParameters(ActionContext ac) {
     @Override
     protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
         HttpParameters previousParams = ac.getParameters();
-        HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(newParams);
+
+        // TODO can we just change the method signature to Map<String, Parameter>?
```
lukaszlenart: I'm ok with changing the signature here

JCgH4164838Gh792C124B5: The original code looks like it would survive a _null_ `newParams`, but the updated code could throw a NullPointerException where the old code wouldn't.  Might need a guard for that condition ...

Does newParams _always_ contain double-wrapped `Parameter` instances ?  If not then the new code might fail (I'm not the best with lambda expressions).

If the incoming `Map` always contains double-wrapped `Parameter` values is there a reason why they are double-wrapped ?  If not then maybe the "upstream" double-wrapping should be corrected somewhere further up in the call sequence ...

JCgH4164838Gh792C124B5:
Hello @rhollencamp .

Do you have a scenario or reproducer app which shows where/how the double-wrapping occurs ?

I couldn't seem to reproduce the behaviour using the Showcase application under debug, but there could be specific conditions under which it happens (custom code, chaining, etc.).

Not being able to reproduce the condition, the only location in the code that I could find which might introduce a double-wrapping of `Parameter` was here: [`HttpParameters.build()` code ](https://github.com/apache/struts/blob/ce4f192676aeb433f8aa020977b62cda0fedf27b/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java#L191-L192).

If somehow the `HttpParameters.build()` can introduce a double-wrapping then it might be easier/safer to guard against the double-wrapping in the first place (rather than have to try and unwrap in the interceptor).  Something like adding a wrapping-guard to HttpParameters.build():
```java
parameters.put(name, 
    value instanceof Parameter ? (Parameter) value : new Parameter.Request(name, value));
```
If it's happening due to code elsewhere, then the above change won't necessarily help.  If the builder _is_ involved then it might work (and require less-overhead) to prevent the double-wrapping.

core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
```diff
@@ -98,9 +100,14 @@ protected HttpParameters retrieveParameters(ActionContext ac) {
...
-    protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
+    protected void addParametersToContext(ActionContext ac, Map<String, Parameter> newParams) {
         HttpParameters previousParams = ac.getParameters();
-        HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(newParams);
+
+        // have to unwrap from Parameter to Object since HttpParameters.Builder will re-wrap
+        Map<String, Object> unwrappedParams = newParams.entrySet().stream().collect(Collectors.toMap(
+            Map.Entry::getKey, e -> e.getValue().getObject()));
+
+        HttpParameters.Builder combinedParams = HttpParameters.create().withParent(previousParams).withExtraParams(unwrappedParams);
```
lukaszlenart: I would rather go with extending the builder and adding a new method `withParameters` that accepts `Map<String, Parameter>` - this will allow to reuse it in the future.

yasserzamani:
Could you please elaborate what issue you encounter with double-wrapped? If any then similar to what @JCgH4164838Gh792C124B5 already mentioned, it seems fixing the wrapper itself (`HttpParameters` class) to not wrap already wrapped params looks a better idea.

rproserpio:
> Could you please elaborate what issue you encounter with double-wrapped? If any then similar to what @JCgH4164838Gh792C124B5 already mentioned, it seems fixing the wrapper itself (`HttpParameters` class) to not wrap already wrapped params looks a better idea.

I've encountered problems with double wrapped params while using the rest plugin (2.5.20).
If I'm not mistaken the `id` url param is extracted by the `ActionMappingParametersInterceptor`, which also call the appropriate setter on the action, and then re-added to the normal params double-wrapped.
The following `ParametersInterceptor` calls the setter for each param, so the `setId` is called again.
This time the `id` is doubled wrapped and cannot be converted (for custom classes) or encoded correctly (when the id is a String with url encoded characters).

rhollencamp:
@JCgH4164838Gh792C124B5 here is an example app that shows the issue https://github.com/rhollencamp/struts-issue-poc/commit/d309051940215ffccedfc3cbb1fbaf6a5e0f4435

core/src/main/java/org/apache/struts2/interceptor/ActionMappingParametersInterceptor.java
rhollencamp: the incoming map is single-wrapped; this is the spot that wrapping the already-wrapped objects

lukaszlenart:
@rhollencamp would you mind creating a ticket in [JIRA](https://issues.apache.org/jira/projects/WW/)? This helps us prepare Version Notes, thanks in advance :)

JCgH4164838Gh792C124B5:
Hello @rhollencamp .

Thanks for the reproducer and identifying this issue.  I'll try the reproducer out later in the week when I have more time, and advise if I uncover anything relevant to this PR. 

By-the-way since you've encountered this issue with the 2.5.20 series you might want to ask @lukaszlenart about a possible backport of a final fix to 2.5.x.

With what you and @rproserpio have encountered with the double-wrapping, do you see it only happening at the point when `Builder.withExtraParams()` is processing ?   I'm curious if there might be other calling sequences where it could also happen and if HttpParameters.build() might still play a factor.

lukaszlenart:
I'm fine with backporting this to 2.5.x

aleksandr-m:
Maybe it is better to add a new method to `HttpParameters.Builder` instead of modifying `withExtraParams` and use it in `ActionMappingParametersInterceptor`.

Overall it is seems like very localized specific issue which should be resolved with minimum impact on existing functionality.

Also what about `Parameter` wrapped in `Parameter` wrapped in `Parameter` wrapped in...

yasserzamani:
@JCgH4164838Gh792C124B5 sorry I couldn't found enough time to experience with this issue and fully understand it but it's in my stack :)

lukaszlenart:
I would rather go with an easy fix now (in 2.5.x) and then consider large refactor in 2.6 of `HttpParameters` - we can pass in a `HttpServletRequest` and build a new fresh parameters based on top of that and assuming any other params need to be rewrapped.

Also I opt for a dedicated method in `Builder` instead of extending the existing one as mentioned by @aleksandr-m

JCgH4164838Gh792C124B5:
If @rhollencamp is OK with the idea, I could try to put together a minimum-impact PR fix for 2.5.x line based on the above discussion ?

rhollencamp:
Any of the approaches discussed will accomplish the goal, and in my eyes at least are more or less the same (minor pros/cons for each but nothing big). IMO just modify `ActionMappingParametersInterceptor` -- that is the class that is passing objects it knows are already wrapped into `Builder`, and this doesn't seem to be a widespread enough problem to bother augmenting `Builder`

lukaszlenart:
Closed by merging #343

PR #344, opened 2019-03-30
[WW-4958] Uses less restrictive RegEx to check if it's a multipart request

lukaszlenart:
Fixes [WW-4958](https://issues.apache.org/jira/browse/WW-4958)


lukaszlenart:
@apache/struts-committers can someone hit the button?

yasserzamani:
After merge I discovered some issues via reading WW-4958 comments references. I tried GitHub's revert on this but wasn't what I expected so I opened the new one #350 to address those.

lukaszlenart:
I wouldn't revert changes using Github functionality, please be aware that this repo is just a mirror, bi-directional but mirror. Also we should keep history consistent, if some changes went in, even wrong one, we shouldn't fix them with another commit and not to revert history of changes.

PR #347, opened 2019-04-06
Proposed fix for WW-5029 for the 2.5.x branch

JCgH4164838Gh792C124B5:
Proposed fix for WW-5029 for the 2.5.x branch:
- NOTE: If the PR is accepted **please credit Maxime Clement for this change** as they found
        the issue, identified the probable cause/related details and opened the JIRA.
- Updated XWorkConfigurationProvider buildAllowedMethods(), loadGlobalAllowedMethods() so that
  they now handle situations when a SAX parser produces multiple elements to represent the tag
  body value.
- No changes to unit tests.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
@@ -874,13 +874,16 @@ public class XmlConfigurationProvider implements ConfigurationProvider {
         if (allowedMethodsEls.getLength() > 0) {
             // user defined 'allowed-methods' so used them whatever Strict DMI was enabled or not
             allowedMethods = new HashSet<>(packageContext.getGlobalAllowedMethods());
-
-            if (allowedMethodsEls.getLength() > 0) {
-                Node n = allowedMethodsEls.item(0).getFirstChild();
-                if (n != null) {
-                    String s = n.getNodeValue().trim();
-                    if (s.length() > 0) {
-                        allowedMethods.addAll(TextParseUtil.commaDelimitedStringToSet(s));
+            // Fix for WW-5029 (concatenate all possible text node children)
+            Node allowedMethodsNode = allowedMethodsEls.item(0);
+            if (allowedMethodsNode != null) {
+                NodeList allowedMethodsChildren = allowedMethodsNode.getChildNodes();
+                for (int i = 0; i < allowedMethodsChildren.getLength(); i++) {
+                    Node allowedMethodsChildNode = allowedMethodsChildren.item(i);
+                    String childNodeValue = (allowedMethodsChildNode != null ? allowedMethodsChildNode.getNodeValue() : "");
+                    childNodeValue = (childNodeValue != null ? childNodeValue.trim() : "");
+                    if (childNodeValue.length() > 0) {
+                        allowedMethods.addAll(TextParseUtil.commaDelimitedStringToSet(childNodeValue));
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

It wasn't possible to re-create the issue that _Maxime Clement_ identified in the JIRA, but the analysis that was provided in the JIRA makes sense given the code.

Please review the PR and advise what you think.  It should be a pretty safe change, but something may have been overlooked. 

The regression passed and debug output of the Showcase application (with some artificial `global-allowed-methods` and `allowed-methods` added in for testing) showed expected results for the "Collected allowed methods:" log output.

JCgH4164838Gh792C124B5:
Oops ... _please ignore the intermediate commit above_.  Accidentally copied changes meant for another branch into this one and had to unwind the changes with a force-push.  :blush:

Unfortunately it upset coveralls ... but otherwise I think I restored the PR to its proper state ...

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
```diff
@@ -874,13 +874,16 @@ protected String guessResultType(String type) {
...
+            // Fix for WW-5029 (concatenate all possible text node children)
+            Node allowedMethodsNode = allowedMethodsEls.item(0);
+            if (allowedMethodsNode != null) {
+                NodeList allowedMethodsChildren = allowedMethodsNode.getChildNodes();
+                for (int i = 0; i < allowedMethodsChildren.getLength(); i++) {
+                    Node allowedMethodsChildNode = allowedMethodsChildren.item(i);
+                    String childNodeValue = (allowedMethodsChildNode != null ? allowedMethodsChildNode.getNodeValue() : "");
+                    childNodeValue = (childNodeValue != null ? childNodeValue.trim() : "");
+                    if (childNodeValue.length() > 0) {
+                        allowedMethods.addAll(TextParseUtil.commaDelimitedStringToSet(childNodeValue));
```
mclem: If I understand the code correctly, this change will not really fix the problem as for each text node of the `allowed-methods` element, you are independently parsing the comma delimited list of values and adding it to the list of allowed methods.
e.g. if the SAX parser reads the `allowed-methods` tag as `[ "method1,me", "thod2" ]`, this implementation will retrieve the following allowed methods : `[ "method1", "me", "thod2" ]`.

I think you should instead construct a single string with the content of all nodes and call `TextParseUtil.commaDelimitedStringToSet` only once, something like this (not tested):
```
if (allowedMethodsEls.getLength() > 0) {
	Node allowedMethodsNode = allowedMethodsEls.item(0);
	NodeList allowedMethodsChildren = allowedMethodsNode.getChildNodes();
	StringBuilder nodeValue = new StringBuilder();
	for (int i = 0; i < allowedMethodsChildren.getLength(); i++) {
		if (allowedMethodsChildren.item(i).getNodeType() == Node.TEXT_NODE) {
			String val = allowedMethodsChildren.item(i).getNodeValue();
			if (val != null) {
				nodeValue.append(val);
			}
		}
	}
	String val = value.toString().trim();
	if (val.length() > 0) {
		allowedMethods.addAll(TextParseUtil.commaDelimitedStringToSet(val));
	}
}
```

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
```diff
@@ -937,11 +940,17 @@ protected void loadGlobalAllowedMethods(PackageConfig.Builder packageContext, El
...
+            // Fix for WW-5029 (concatenate all possible text node children)
+            Node globaAllowedMethodsNode = globalAllowedMethodsElms.item(0);
+            if (globaAllowedMethodsNode != null) {
+                NodeList globaAllowedMethodsChildren = globaAllowedMethodsNode.getChildNodes();
+                for (int i = 0; i < globaAllowedMethodsChildren.getLength(); i++) {
+                    Node globalAllowedMethodsChildNode = globaAllowedMethodsChildren.item(i);
+                    String childNodeValue = (globalAllowedMethodsChildNode != null ? globalAllowedMethodsChildNode.getNodeValue() : "");
+                    childNodeValue = (childNodeValue != null ? childNodeValue.trim() : "");
+                    if (childNodeValue.length() > 0) {
+                        globalAllowedMethods.addAll(TextParseUtil.commaDelimitedStringToSet(childNodeValue));
```
mclem: Same comment as above, `TextParseUtil.commaDelimitedStringToSet` should be called only once with the complete value of the `allowed-methods` node.

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
JCgH4164838Gh792C124B5: Hello @mclem.
Thanks for pointing out that weakness ... looks like I was focusing too much on the other elements and missed that.  :blush:
Wasn't able to come up with an effective unit test for this scenario, and the 1st attempt at a fix still worked when the SAX parser provided a single node of content.
The latest commit provides updated logic similar to what you suggested above, which should work in theory ...

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
JCgH4164838Gh792C124B5: Both methods were updated,  See above response.

JCgH4164838Gh792C124B5:
Thanks to @mclem the PR has been updated with (hopefully) corrected logic to guard against the scenario he identified.
If anyone can confirm the fix appears to work (or appears to still fail) when running under WebSphere please advise.

JCgH4164838Gh792C124B5:
Hello Again.

Some unit tests for this PR's intended fix have been added.  In addition two new unit tests for testing Result types were also added.  The mock objects being used are a little rough, but appear to be sufficient for test purposes.

If anyone notices issues with the tests, please advise.

JCgH4164838Gh792C124B5:
Hello @mclem .

Please advise if the updated changes look OK now.  If the intermittent SAX parsing issue can happen with WebSphere 8.5 it might affect other application servers (intermittently) as well.

If @lukaszlenart , @aleksandr-m and @yasserzamani have any recommendations for additional changes or improvements to this PR they can advise.

lukaszlenart:
Nope, astonish work! 👏

yasserzamani:
LGTM too  👏

PR #348, opened 2019-04-07
Proposed fix for WW-5011 for the 2.5.x branch

JCgH4164838Gh792C124B5:
Proposed fixes for WW-5011 and WW-5028 for the 2.5.x branch:
- NOTE: If the PR is accepted please credit Jason Pyeron for the WW-5011 portion of this change
        as he found the issue, proposed a fix and opened the WW-5011 JIRA.
- For WW-5028:
  - Disable printing stacktrace on exceptions by the Dispatcher by default.
  - It can be re-enabled by configuration.
- For WW-5011:
  - Provide an "opt-in" alternate file resource handler for the Tiles plugin.
  - It can be enabled via "context-param" in web.xml

### DIFF
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
@@ -30,11 +30,21 @@ import javax.servlet.ServletContext;
 
 public class StrutsTilesInitializer extends AbstractTilesInitializer {
 
+    public static final String STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING = "struts.tiles.alternatefileresourcehandling";
+
     private static final Logger LOG = LogManager.getLogger(StrutsTilesInitializer.class);
 
     @Override
     protected ApplicationContext createTilesApplicationContext(ApplicationContext preliminaryContext) {
         ServletContext servletContext = (ServletContext) preliminaryContext.getContext();
+
+        // Opt-in alternate file resource handling for WW-5011
+        if (Boolean.parseBoolean(servletContext.getInitParameter(STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING))) {
+            StrutsWildcardServletApplicationContext.setUseAlternateFileResourceHandling(true);
+        } else {
+            StrutsWildcardServletApplicationContext.setUseAlternateFileResourceHandling(false);
+        }
+        LOG.trace("Tiles alternate file resourcehandling: " + StrutsWildcardServletApplicationContext.isUseAlternateFileResourceHandling());
 
         if (servletContext.getInitParameter(DefinitionsFactory.DEFINITIONS_CONFIG) != null) {
             LOG.trace("Found definitions config in web.xml, using standard Servlet support ....");
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
@@ -30,7 +30,7 @@ import javax.servlet.ServletContext;
 
 public class StrutsTilesInitializer extends AbstractTilesInitializer {
 
-    public static final String STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING = "struts.tiles.alternatefileresourcehandling";
+    public static final String STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING = "struts.tiles.alternateFileResourceHandling";
 
     private static final Logger LOG = LogManager.getLogger(StrutsTilesInitializer.class);
 
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsApplicationResourceAlternate.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsApplicationResourceAlternate.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.tiles;
+
+import org.apache.tiles.request.locale.PostfixedApplicationResource;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+
+public class StrutsApplicationResourceAlternate extends PostfixedApplicationResource {
+
+    private static final String URL_FILE_PROTOCOL = "file";
+
+    private final URL url;
+
+    /**
+     * Generate an alternate localePath parameter string (external form) for any
+     * URL that uses a file protocol,  For any other protocol, it simply returns the
+     * url Path.
+     * 
+     * For a file url, the alternate file localePath supports a full path including
+     * a sub-context (e.g. Some_Path#SomeSubcontext).
+     * 
+     * @param url
+     * 
+     * @return 
+     */
+    public static String alternateFileLocalePath(URL url) {
+        final String result;
+        final String urlProtocol = url.getProtocol();
+
+        if (URL_FILE_PROTOCOL.equalsIgnoreCase(urlProtocol)) {
+            final String urlExternalForm = url.toExternalForm();
+            if (urlExternalForm != null && urlExternalForm.startsWith(urlProtocol + ':')) {
+                result = urlExternalForm.substring(urlProtocol.length() + 1);
+            } else {
+                result = url.getPath();
+            }
+        } else {
+            result = url.getPath();
+        }
+
+        return result;
+    }
+
+    /**
+     * Construct a StrutsApplicationResourceAlternate that uses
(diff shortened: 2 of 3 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This PR proposes solutions for the two JIRAs mentioned in the commit and initial comment.  Although they are not directly related to each other, both involve providing control parameters and were being worked on at the same time, so they got bundled together.

They are not very larger changes, so hopefully it's OK to bundle them (at least for now).

**Note**: _If the PR is accepted_ **please credit Jason Pyeron for the WW-5011 portion of this change**.

lukaszlenart:
I would not combine two different fixes in one PR. I know it's very hard to manage different branches and changes but the review process is way more harder especially if some changes would have been requested.

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -219,4 +219,7 @@ struts.ognl.enableExpressionCache=true
 ### Indicates if Dispatcher should handle unexpected exceptions by calling sendError()
 ### or simply rethrow it as a ServletException to allow future processing by other frameworks like Spring Security
 struts.handle.exception=true
+
+### Indicates if Dispatcher should print a stracktrace upon receiving unexpected exceptions
+struts.printstacktrace.onexception=false
```
lukaszlenart: This whole part look very good, I'm down 👍

plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsApplicationResource.java
```diff
@@ -27,13 +27,55 @@
...
     }
 
+    /**
+     * Construct a StrutsApplicationResource that has the option of using 
+     * alternate file resource handling logic (for subcontexts).
+     * 
+     * @param url
+     * @param useAlternateFileResourceHandling 
+     */
+    public StrutsApplicationResource(URL url, boolean useAlternateFileResourceHandling) {
```
lukaszlenart: Instead of a flag I would introduce another version of `StrutsApplicationResource` and create a proper instance accordingly.

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -582,7 +597,9 @@ public void serviceAction(HttpServletRequest request, HttpServletResponse respon
             logConfigurationException(request, e);
             sendError(request, response, HttpServletResponse.SC_NOT_FOUND, e);
         } catch (Exception e) {
-            e.printStackTrace();
+            if (printStackTraceOnException) {
```
aleksandr-m: Why not just use logger?

aleksandr-m:
@JCgH4164838Gh792C124B5 Please use different commits and PR-s for different issues.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart  and @aleksandr-m .

Please advise if you want me to split this PR into separate PRs for each of the two separate issues (I know it isn't the best form ... I just ended up working on them together in the same build project).

The PR separation could be done before or after addressing the comments and feedback you've provided above.

lukaszlenart:
I would like to see two different PRs when we go further.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart and @aleksandr-m .

This PR has been updated (using an amended commit) so that it only deals with WW-5011.   I believe it addresses the change that @lukaszlenart requested as well.

I will open a new (separate) PR for the WW-5028 and look to address @aleksandr-m 's question in that PR.

plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
```diff
@@ -30,12 +30,22 @@
...
 
+    public static final String STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING = "struts.tiles.alternatefileresourcehandling";
+
     private static final Logger LOG = LogManager.getLogger(StrutsTilesInitializer.class);
 
     @Override
     protected ApplicationContext createTilesApplicationContext(ApplicationContext preliminaryContext) {
         ServletContext servletContext = (ServletContext) preliminaryContext.getContext();
 
+        // Opt-in alternate file resource handling for WW-5011
```
yasserzamani: Sorry but it doesn't look good to me at all! It's not user's business to switch Struts behavior to fix Struts issue! I would like to **add** such urls support to `StrutsApplicationResource` class privately - In this case, the fix should be transparent for outside of that class.

JCgH4164838Gh792C124B5: Hello @yasserzamani .

_I think I understand your concern_ here, but _changing the existing default behaviour could inadvertently break existing usage_.  For that reason it seemed an opt-in (via context param in this case) would be reasonable (not all that different from use cases where Struts configuration parameters are available to users to modify behaviour).

It would certainly be nicer/cleaner to change `StrutsApplicationResource` transparently, but due to the risk of breaking something, maybe that could be done in Struts 2.6.x instead (where some breaking behaviour can be expected) ?

yasserzamani: As far as I could understand the issue, I think the issue is `StrutsApplicationResource` cannot handle resources whose have `#` in their url, right? if so, that's why I bold "add" in my above comment :) I meant, now it should **also** support such urls, it doesn't break previous behavior but is union of the previous behavior and newly added one. Am I wrong? Or maybe you mean it isn't doable in any way without breaking backward-compatibility? if so why? it looks strange to me ! e.g. as a quick dirty solution I can think about this:

```java
String myGetPath(url) {
result=url.getPath();
nsIndex=url.indexOf('#');
if(nsIndex>0)
result += url.subString(nsIndex);
return result;
}
```

lukaszlenart:
I wonder if it's possible to automatically discover that special case and use a proper resource class instead of having a flag that must be managed by a user. Wouldn't be possible to somehow figure it out?

JCgH4164838Gh792C124B5:
My past experience with `Tiles` configuration has been that the resource look-ups can be "_finicky_" and Struts 2.5.x users who have things working (by combination of Tiles and Struts configuration) might be unintentionally impacted if the default file URL processing behaviour is changed in the `Tiles Plugin`.

For that reason only it seemed better to provide an "opt-in" solution that could address the issue that was reported for _Apache Roller_, but not pose any risk of breaking existing `Struts Tiles Plugin` usage for 2.5.x.  Also, from past posts, it seems most users have been able to deal with the sub-contesxt "#" type URLs by re-configuring their Struts/Tiles/application archive configurations (making this a "niche fix").

The "sub-context" character(s) for Tomcat seems to have changed over time "##" or "#", and might be different for other application servers too.  That could make "automatic discovery" error-prone too.

A transparent fix  _might_ be possible but it didn't seem worth the risk (even a small risk) for 2.5.x.  For the those reasons I figured an "opt-in" fix for 2.5.x was a reasonable compromise.  
For 2.6.x a fully transparent fix (_no control flag_) can replace this one as some "minor breakage" during migration is more acceptable with 2.6.

That's my reasoning anyway, but the PR can be adjusted to fit the final consensus for 2.5.x.  :)

yasserzamani:
Sorry @JCgH4164838Gh792C124B5 , yesterday I found some time to spend on issue and understood what you have done and trying to do 👍 I found actually it's not a Struts issue because I converted e.g. `/path/file###version.extension` to URL and I saw it misses information:

> URL data:
> path:/path/file
> ref:#version.extension

Missed information is count of number signs! I think firstly we should research if it's officially OK to have "#" in jar and war e.g. what about `/path/context#1/war#1/...`. Maybe we are not responsible to fix it at all and tomcat project is. I don't know, we should explore.

plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
```diff
@@ -30,12 +30,22 @@
 
 public class StrutsTilesInitializer extends AbstractTilesInitializer {
 
+    public static final String STRUTS_TILES_ALTERNATE_FILERESOURCEHANDLING = "struts.tiles.alternatefileresourcehandling";
```
lukaszlenart: I would use a camel case version here as more user friendly, eg.:
```
struts.tiles.alternateFileRsourceHandling
```

lukaszlenart:
I'm fine with current solution, we can always improve it in an another release

plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesInitializer.java
JCgH4164838Gh792C124B5: Thanks for pointing that out.  It has been changed to camel case in an updated commit (and the JIRA comment now references the new value as well).

JCgH4164838Gh792C124B5:
Hello Again.

As per Yasser's comments above this PR could be trying to address something that isn't necessarily Struts / Tiles "responsibility" to fix.  This PR fix could still be considered a "workaround fix" in that case.

Yasser also recently asked (on JIRA) the issue reporter (Jason Pyeron) to see if they can share their "working patch" in some fashion.  Their feedback could augment or supersede this PR.

yasserzamani:
With further explores I found I felt right and there is a clean solution exists :) I found it's a Struts issue and now, I would like to import #352 as a fix for this branch also instead. thanks!

yasserzamani:
Hi @JCgH4164838Gh792C124B5 , by transparent fix I meant something like #353 - according to your experience with Tiles wdyt about that and backward-compatibility? Does it look good?

JCgH4164838Gh792C124B5:
Hello @yasserzamani .
I've left some newer comments for PR#353 (add-ons to my original review).  
Please let me know your thoughts on those comments in PR#353 (I added comments in response to most of your review replies from Apr. 25th).
A few more tweaks (and hopefully opt-out mechanism) and PR#353 should be "good to go" (at which point I can close this PR).

yasserzamani:
Hi @JCgH4164838Gh792C124B5 , thanks for them! I improved my work there inspiring from Tiles URLApplicationResource which I think addresses your comments :) - IMHO an opt-out mechanism isn't essential and is a waste complexity -- these are just an improvement on path not change on path.

JCgH4164838Gh792C124B5:
The issue this PR attempted to resolve has now been dealt with in Y. Zamani's more recent PR (as per the comments above).

There's nothing further in this PR that is needed, so it will be closed.

PR #349, opened 2019-04-13
Proposed fix for WW-5028 for the 2.5.x branch:

JCgH4164838Gh792C124B5:
Proposed fix for WW-5028 for the 2.5.x branch:
- Disable printing stacktrace on exceptions by the Dispatcher by default.
- Printing stacktrace on exception can be re-enabled by configuration.
- Now prints stacktrace on exception using LOG, as suggested by A. Mashchenko and the the JIRA reporter.

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
@@ -582,8 +582,10 @@ public class Dispatcher {
             logConfigurationException(request, e);
             sendError(request, response, HttpServletResponse.SC_NOT_FOUND, e);
         } catch (Exception e) {
-            e.printStackTrace();
             if (handleException || devMode) {
+                if (devMode) {
+                    LOG.debug("Dispatcher serviceAction failed", e);
+                }
                 sendError(request, response, HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e);
             } else {
                 throw new ServletException(e);
--- a/core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
+++ b/core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
@@ -321,6 +321,38 @@ public class DispatcherTest extends StrutsInternalTestCase {
         assertTrue(du.isMultipartRequest(req));
     }
 
+    /**
+     * Verify proper default (true) handleExceptionState for Dispatcher and that
+     * it properly reflects a manually configured change to false.
+     * 
+     * @throws Exception 
+     */
+    public void testHandleException() throws Exception {
+        Dispatcher du = initDispatcher(new HashMap<String, String>());
+        assertTrue("Default Dispatcher handleException state not true ?", du.isHandleException());
+
+        Dispatcher du2 = initDispatcher(new HashMap<String, String>() {{
+            put(StrutsConstants.STRUTS_HANDLE_EXCEPTION, "false");
+        }});
+        assertFalse("Modified Dispatcher handleException state not false ?", du2.isHandleException());
+    }
+
+    /**
+     * Verify proper default (false) devMode for Dispatcher and that
+     * it properly reflects a manually configured change to true.
+     * 
+     * @throws Exception 
+     */
+    public void testDevMode() throws Exception {
+        Dispatcher du = initDispatcher(new HashMap<String, String>());
+        assertFalse("Default Dispatcher devMode state not false ?", du.isDevMode());
+
+        Dispatcher du2 = initDispatcher(new HashMap<String, String>() {{
+            put(StrutsConstants.STRUTS_DEVMODE, "true");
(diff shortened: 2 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello @lukaszlenart and @aleksandr-m .

This PR separates the fix for WW-5028 from PR#348 into its own PR.

It also addresses @aleksandr-m 's comment and moves the stacktrace print (when enabled) from stdout to the LOG instance for the dispatcher.

Let me know if additional changes are needed.

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -304,6 +310,20 @@ public boolean isHandleException() {
         return handleException;
     }
 
+    /**
+     * Modify state of StrutsConstants.STRUTS_PRINTSTACKTRACE_ONEXCEPTION setting.
+     * 
+     * @param printStackTraceOnException 
+     */
+    @Inject(value = StrutsConstants.STRUTS_PRINTSTACKTRACE_ONEXCEPTION, required = false)
```
yasserzamani: I think @aleksandr-m meant (or at least I would like) to just revert [accidentally added `e.print...`](https://issues.apache.org/jira/browse/WW-5028?focusedCommentId=16808380&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16808380) line. Or as a last resort we can just replace it with a `LOG.debug`. IMHO more complexity (e.g. enable user to change it via configuration) isn't essential here.

JCgH4164838Gh792C124B5: Hello @yasserzamani .

Thanks for the feedback.

Having a control flag might not be essential, but having control over Dispatcher producing stacktraces (whether to stdout or a log) is convenient and seems to provide a solution in line with what the JIRA opener was seeking.

I think it's fair to offer users the ability to decide, via a control flag, if they want to see/track these `Dispatcher` errors (sometime Log level-separation isn't flexible enough).

Once the others give their feedback the PR can be updated to go with whatever the consensus is.

yasserzamani: I try to represent my concern in another way :) I think anyone who agree a control flag should answer then why we don't have such logic for other classes and why only `Dispatcher` is specialized for such user control on exceptions via configuration?

I understand your concern and now I can have a better recommendation below instead of my previous recommendation `LOG.debug`:
```diff
Index: core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java	(revision 68044e741bf2566c96c876265ce99f9a851f96f7)
+++ core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java	(date 1551013035271)
@@ -101,7 +101,7 @@
     }
 
     protected void handleErrorInDevMode(HttpServletResponse response, int code, Exception e) {
-        LOG.debug("Exception occurred during processing request: {}", e.getMessage(), e);
+        LOG.warn("Exception occurred during processing request: {}", e.getMessage(), e);
         try {
             List<Throwable> chain = new ArrayList<>();
             Throwable cur = e;
```
i.e. warn in `devMode` which I think is enough.

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -582,7 +602,9 @@ public void serviceAction(HttpServletRequest request, HttpServletResponse respon
             logConfigurationException(request, e);
             sendError(request, response, HttpServletResponse.SC_NOT_FOUND, e);
         } catch (Exception e) {
-            e.printStackTrace();
+            if (printStackTraceOnException) {
```
lukaszlenart: Instead of using a dedicated flag I would reuse the existing one - `devMode` in such case. I think the my purpose of putting `e.printStackTrace()` here was to fix some issues with unit tests, not a production related one.

aleksandr-m: And there is already an `if` statement below `if (handleException || devMode) {` LOG can go there.

JCgH4164838Gh792C124B5:
A revised commit attempts to address the feedback - removed the separate flag, uses devMode only to control it, only uses debug level output for the stacktrace.

It uses the existing (handleException || devMode) block but still limits the stacktrace to devMode only (as a debug output).  If you'd prefer a change to DefaultDispatcherErrorHandler instead (as per Yasser's suggestion above) the PR can be updated to match that instead.

lukaszlenart:
I'm good with the solution, thank you for all your work :)

PR #350, opened 2019-04-14
fix multipart request RegEx (relates to WW-4958)

yasserzamani:
parameter      = token "=" ( token / **quoted-string** )
      Note: Unlike some similar constructs in other header fields, media
      type parameters **do not allow whitespace (even "bad" whitespace)
      around the "=" character.**

Reference: https://tools.ietf.org/html/rfc7231#section-3.1.1.1

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
@@ -85,7 +85,7 @@ public class Dispatcher {
      */
     public static final String REQUEST_POST_METHOD = "POST";
 
-    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart/form-data(?:\\s*;\\s*boundary\\s?=[0-9a-zA-Z'()+_,\\-./:=?]{1,70})?(?:\\s*;\\s*charset\\s?=\\s?[a-zA-Z\\-0-9]{3,14})?";
+    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart/form-data(?:\\s*;\\s*boundary=[0-9a-zA-Z'()+_,\\-./:=?]{1,70})?(?:\\s*;\\s*charset=[a-zA-Z\\-0-9]{3,14})?";
 
     /**
      * Provide a thread local instance.
--- a/core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
+++ b/core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
@@ -323,18 +323,6 @@ public class DispatcherTest extends StrutsInternalTestCase {
         req.setContentType("multipart/form-data;boundary=---------------------------207103069210263;charset=UTF-16LE");
         assertTrue(du.isMultipartRequest(req));
 
-        req.setContentType("multipart/form-data;boundary=---------------------------207103069210263 ; charset = UTF-16LE");
-        assertTrue(du.isMultipartRequest(req));
-
-        req.setContentType("multipart/form-data;boundary=---------------------------207103069210263;charset = UTF-16LE");
-        assertTrue(du.isMultipartRequest(req));
-
-        req.setContentType("multipart/form-data;boundary=---------------------------207103069210263;charset= UTF-16LE");
-        assertTrue(du.isMultipartRequest(req));
-
-        req.setContentType("multipart/form-data;boundary=---------------------------207103069210263;charset =UTF-16LE");
-        assertTrue(du.isMultipartRequest(req));
-
         req.setContentType("multipart/form-data;boundary=---------------------------207103069210263; charset=UTF-16LE");
         assertTrue(du.isMultipartRequest(req));
 
@@ -348,9 +336,6 @@ public class DispatcherTest extends StrutsInternalTestCase {
         assertTrue(du.isMultipartRequest(req));
 
         req.setContentType("multipart/form-data ; boundary=---------------------------207103069210263;charset=UTF-16LE");
-        assertTrue(du.isMultipartRequest(req));
-
-        req.setContentType("multipart/form-data;boundary =---------------------------207103069210263;charset=UTF-16LE");
         assertTrue(du.isMultipartRequest(req));
 
         req.setContentType("Multipart/Form-Data ; boundary=---------------------------207103069210263;charset=UTF-16LE");


core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -85,7 +85,7 @@
      */
     public static final String REQUEST_POST_METHOD = "POST";
 
-    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart/form-data(?:\\s*;\\s*boundary\\s?=[0-9a-zA-Z'()+_,\\-./:=?]{1,70})?(?:\\s*;\\s*charset\\s?=\\s?[a-zA-Z\\-0-9]{3,14})?";
+    public static final String MULTIPART_FORM_DATA_REGEX = "^multipart/form-data(?:\\s*;\\s*boundary=[0-9a-zA-Z'()+_,\\-./:=?\"]{1,70})?(?:\\s*;\\s*charset=[a-zA-Z\\-0-9\"]{3,14})?";
```
JCgH4164838Gh792C124B5: Hello @yasserzamani .

Good catch on the missing quoted string element from RFC7231 3.1.1.1.  👍 

I just noticed that 3.1.1.2 seems to indicate "A charset is identified by a case-insensitive token.  charset = token", 
In 3.1.1.1 it has a mixed-case example of charset as well: "Text/HTML;_Charset_="utf-8"".  So it looks like charset should be case-insensitive too.  If you agree with that assessment, then maybe the `MULTIPART_FORM_DATA_REGEX` expression should be updated so that charset is case-insensitive:

`"^multipart/form-data(?:\\s*;\\s*boundary=[0-9a-zA-Z'()+_,\\-./:=?\"]{1,70})?(?:\\s*;\\s*(?i:charset)=[a-zA-Z\\-0-9\"]{3,14})?";`

What do you think ?

yasserzamani: Nice catch 👍 However, I tried to test it but passed, then I investigated and saw Struts uses `contentType.toLower` before regex check so is case-insensitive and `A-Z` in regex aren't effective and can be removed!

Nevertheless I found more issues and found it not such simple to check. For example what about when `charset` come first then `boundary`, it's valid by RFC. @lukaszlenart I think `contentType.leftTrim().startsWithIgnoreCase("multipart")` looks enough. Could you recall the philosophy behind why honoring `boundary` in 4e9fa8423931417da8bc60ce220f46935b54c5de and `charset` in 10027dae424231aeb6b7a4110ddc3c308bd2038b? I think despite how much parameters it has or will have, it's multipart when starts with "multipart".

JCgH4164838Gh792C124B5: The two JIRAs only seem to refer to allowing customized regex for controlling multipart processing.  Could simply be an attempt to limit malformed submissions, but hopefully Lukasz can clarify.

Given the `contentType.toLowerCase()` before the regex is checked, and Yasser's comment about possible reverse ordering for `charset` and `boundary`, then this regex for `MULTIPART_FORM_DATA_REGEX` might be good enough as a default:

`"^multipart/form-data(?:(?:\\s*;\\s*boundary=[0-9a-z'()+_,\\-./:=?\"]{1,70})?(?:\\s*;\\s*charset=[a-z\\-0-9\"]{3,14})?||(?:\\s*;\\s*charset=[a-z\\-0-9\"]{3,14})?(?:\\s*;\\s*boundary=[0-9a-z'()+_,\\-./:=?\"]{1,70})?)";`

The above regex passed a slightly out-of-date Struts 2.5.21-SNAPSHOT version regression suite.

Since Struts users can always override it with something custom to their preference, maybe it's OK to just provide a reasonable default (to cover 99% of the cases) ?

lukaszlenart: Using simple `contentType.leftTrim().startsWithIgnoreCase("multipart")` is not a good idea, we were there and it was a disaster.

As @JCgH4164838Gh792C124B5 said, providing a reasonable defaults is fair enough, we don't have to satisfy everyone upfront. There is only one issue reported that the RegEx is too restrictive, I would relax it a bit and that's all.

Users can alway override the default and report an issue, then we can work on another solution, iteratively.

core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
```diff
@@ -355,6 +355,10 @@ public void testIsMultipartRequest() throws Exception {
 
         req.setContentType("Multipart/Form-Data ; boundary=---------------------------207103069210263;charset=UTF-16LE");
         assertTrue(du.isMultipartRequest(req));
+
+        //test order-insensitive, tabs and tspecials characters
```
sepe81: tspecials -> special

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
yasserzamani: OK I reviewed security list archive and now pushed an enough (no less no too) restrictive regex while keeping security same :)

core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
yasserzamani: I gave it from the reference (https://tools.ietf.org/html/rfc2045#section-5.1) - maybe it means token specials.

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
JCgH4164838Gh792C124B5: Good work on the research into this, @yasserzamani .  👍 

Would it be a little safer to _keep one of the slightly-more-restrictive regex_ (discussed earlier, in earlier revisions of the PR) in place as the default for `MULTIPART_FORM_DATA_REGEX` ?

The _more permissive/balanced regex_ you've researched and constructed _could be referenced in the Struts documentation_ (along with the full comment explanation). in the section explaining how to use the "struts.multipart.validationRegex" override mechanism (https://struts.apache.org/core-developers/file-upload.html).

That way the default remains somewhat more restrictive, but users have _a more permissive option visible/handy in the documentation_ to use if the default doesn't work for them.

What do you think ?

lukaszlenart: Good idea, 👍

yasserzamani: According to definitions, blocking valid requests also means poor security.

Nevertheless, IMHO it's not nice for Struts to bother user while it simply can be handled internally. Nowadays developers have enough personal concerns to handle so they love 0-configuration frameworks like `spring-boot` where they don't have to configure as much as is possible.

lukaszlenart: I still opt for more restrictive RegEx that resolves the reported issues. Till now there were no other reports that the current, very restive RegEx is a problem so I would assume relaxing it a bit is what we need.

yasserzamani: 👍 so I reset this PR to only remove white-space around "=" character which is more restrictive - and according to RFCs white-space around "=" character is invalid in media type parameters.

lukaszlenart:
LGTM 👍

lukaszlenart:
@yasserzamani would you mind preparing the same fix for 2.6?

PR #351, opened 2019-04-15
Add common .gitignore patterns from gitignore.io

sepe81:
Ignore .factorypath b/c it gets generated during maven compile.

Don't need to ignore every single target folder.
Instead use target/ pattern and add more common maven files.

Ignore .sonar folder for local sonar scanning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@
 .metadata/
 Servers/
 
+# Java annotation processor (APT)
+.factorypath
+
 #VSCode
 .vscode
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -24,42 +27,19 @@ Servers/
...
-plugins/tiles/target
-bundles/target
-plugins/target
-target
+
 plugins/testng/test-output
 test-output
+
+# Sonar
+/.sonar/
```
lukaszlenart: Do we need that leading slash?

lukaszlenart:
Nice work 👏

.gitignore
sepe81: > Do we need that leading slash?

Not necessarily, but according to https://git-scm.com/docs/gitignore "A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

Because there is only one `.sonar` dir in the root IMHO with the leading slash this is more specific.

lukaszlenart: 👍

lukaszlenart:
LGTM 👍

PR #352, opened 2019-04-20
WW-5011 use Tiles already provided URLApplicationResource

yasserzamani:
see also: https://github.com/joshlong/spring-framework/blob/910dd4146721a803cd9d246b91854e77ffe9272a/spring-webmvc-tiles3/src/main/java/org/springframework/web/servlet/view/tiles3/SpringWildcardServletTilesApplicationContext.java#L88


yasserzamani:
> I wonder if the recent JIRA the PRs seeking to mitigate WW-5001 might also find WW-4672 (unresolved) relevant ?

Wow! great catch, thanks! I'm not a Tiles expert but I guess yes it fixes that also because now we use Tiles itself provided impel which isn't anticipated to be incompatible with itself.

> Will dropping StrutsApplicationResource completely possibly 0break projects using custom resource handlers if they extended/referenced StrutsApplicationResource ? Maybe this would be safer to attempt with 2.6 instead ?
> 
> It is not really clear to me why PostfixedApplicationResource was extended instead of URLApplicationResource by StrutsApplicationResource in the first place, but maybe there was a specific reason ? Maybe it was to avoid allowing usage of the File or InputStream references available via URLApplicationResource ?
> 
> Attempting this PR's idea in the 2.6 branch might be less risky (there might be a reason why the original implementation went the direction it did that isn't absolutely clear).
> 
> What do you think ?

I don't think so because as far as I can see it is not intended to allow user `custom resource handlers` and I don't see it inject-able and documented so we don't expect user to use it. It firstly is introduced in #73 commit "[Adds support for I18N](https://github.com/apache/struts/commit/328da4e2258573acd049cdd3fd5bce2f37869adf)" to fix `StrutsWildcardServletApplicationContext.getResource(ApplicationResource base, Locale locale)` logic to honor `locale`. But I don't think adding `StrutsApplicationResource` is needed. As far as I can see, `new URLApplicationResource(localePath, localFile.toURI().toURL())` not only does what `new StrutsApplicationResource(localFile.toURI().toURL())` does but also is more cleaner and less buggy.

lukaszlenart:
I think it would be safer instead of dropping `StrutsApplicationResource`, changing it to inherit from `URLApplicationResource` (which also extends `PostfixedApplicationResource`).

yasserzamani:
I already checked safety and realized that `URLApplicationResource > StrutsApplicationResource` I meant it's methods at least do what ours do but also are more safe and robust that our i.e. our may fail in some special cases like WW-5011.

lukaszlenart:
This PR goes into the `master` branch and dropping the class is ok, so 👍

lukaszlenart:
LGTM 👍

lukaszlenart:
@yasserzamani do you plan to prepare a similar fix for 2.5.x?

lukaszlenart:
This PR broke support for Tiles, `StrutsApplicationResource` needs to be restored to properly support wildcard matching

PR #355, opened 2019-05-11
Minor enhancement/fix to AbstractLocalizedTextProvider:

JCgH4164838Gh792C124B5:
Minor enhancement/fix to AbstractLocalizedTextProvider:
- Made "constant" RELOADED static to save an initialization every time.
- Updated clearBundle(final String bundleName) method comment as it
  seemed inaccurate, added debug log output when called.
- Introduced protected clearBundle(final String bundleName, Locale locale)
  method for use by descendants, with debug log output when called.
- Introduced protected clearMissingBundlesCache() method for use by
  descendants, with debug log when called.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
@@ -53,7 +53,7 @@ abstract class AbstractLocalizedTextProvider implements LocalizedTextProvider {
     public static final String STRUTS_MESSAGES_BUNDLE = "org/apache/struts2/struts-messages";
 
     private static final String TOMCAT_RESOURCE_ENTRIES_FIELD = "resourceEntries";
-    private final String RELOADED = "com.opensymphony.xwork2.util.LocalizedTextProvider.reloaded";
+    private static final String RELOADED = "com.opensymphony.xwork2.util.LocalizedTextProvider.reloaded";
 
     protected final ConcurrentMap<String, ResourceBundle> bundlesMap = new ConcurrentHashMap<>();
     protected boolean devMode = false;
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This PR provides a minor fix/enhancement to AbstractLocalizedTextProvider while looking at the issue/condition described in WW-4943.

Although it doesn't address their issue directly the new utility methods might be of use for a custom (user-created) AbstractLocalizedTextProvider that can provide a scenario-specific-workaround.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart , @yasserzamani  and @aleksandr-m .

Please take a look at this PR when you have some "free cycles".  It didn't seem significant enough for a Jira, but if you think one is needed (assuming you think the PR is worth pursuing), I can create one for it as well.

There was consideration for making AbstractLocalizedTextProvider.clearBundle() a "null-op" since it seems to be ineffective (and confirmed by the new unit test) ... but someone might have extended AbstractLocalizedTextProvider and is relying on the current behaviour for some reason.

lukaszlenart:
> There was consideration for making AbstractLocalizedTextProvider.clearBundle() a "null-op" since it seems to be ineffective (and confirmed by the new unit test) ... but someone might have extended AbstractLocalizedTextProvider and is relying on the current behaviour for some reason.

We can change that in 2.6 and break backward compatibility

lukaszlenart:
LGTM 👍

JCgH4164838Gh792C124B5:
Jira WW-5034 has been opened for this enhancement for tracking purposes.

lukaszlenart:
Thanks!

When you first open a JIRA ticket and then reference it by key in PR's title or description, ASF bot will link the ticket to the PR.

lukaszlenart:
@JCgH4164838Gh792C124B5 do you plan to move these changes to 2.6 with option to break backward compatibility?

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .
Sure.  I will try to put together an equivalent PR for 2.6 sometime over the next week.  It will probably mark  the single-parameter clearBundle() as deprecated (while making it a no-op).

lukaszlenart:
Thanks!

PR #357, opened 2019-05-28
WW-4999 honor logMissingProperties on findValue as well

yasserzamani:
fixes WW-4999


lukaszlenart:
This is a wrong fix, see my comment
https://issues.apache.org/jira/browse/WW-4999?focusedCommentId=16747731&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16747731

There was a change in the OGNL that broke this functionality.

yasserzamani:
You're right, this is wrong and I close this PR.

FYI, despite OGNL, as far as I can see Struts OgnlValueStack and CompoundRootAccessor logic, `logMissingProperties` will never be reached and honored except with `true` for `throwExceptionOnFailure` - because `handleOgnlException` is only called in catch blocks and catch blocks will never be reached except on an exception and an exception never thrown except with `true` for `throwExceptionOnFailure`.

PR #358, opened 2019-05-30
fix logMissingProperties (WW-4999)

yasserzamani:
Moves checking OgnlValueStack.THROW_EXCEPTION_ON_FAILURE outside loop because it shouldn't throw exception on first failure while is trying all root objects.

Returns on first successful call because it's not rational and is confusing user to skip when user method successfully returns null as an actual result.

Fixes WW-4999 via honoring (devMode && logMissingProperties) for OgnlValueStack.THROW_EXCEPTION_ON_FAILURE and REPORT_ERRORS_ON_NO_PROP.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -340,7 +340,8 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     }
 
     protected boolean shouldLogMissingPropertyWarning(OgnlException e) {
-        return (e instanceof NoSuchPropertyException || e instanceof MethodFailedException)
+        return (e instanceof NoSuchPropertyException ||
+                (e instanceof MethodFailedException && e.getReason() instanceof NoSuchMethodException))
         		&& devMode && logMissingProperties;
     }
 
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -182,8 +182,7 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
 
     private void trySetValue(String expr, Object value, boolean throwExceptionOnFailure, Map<String, Object> context) throws OgnlException {
         context.put(XWorkConverter.CONVERSION_PROPERTY_FULLNAME, expr);
-        context.put(REPORT_ERRORS_ON_NO_PROP, throwExceptionOnFailure || (devMode && logMissingProperties)
-                ? Boolean.TRUE : Boolean.FALSE);
+        context.put(REPORT_ERRORS_ON_NO_PROP, throwExceptionOnFailure || logMissingProperties ? Boolean.TRUE : Boolean.FALSE);
         ognlUtil.setValue(expr, context, root, value);
     }
 
@@ -247,7 +246,7 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     }
 
     protected void setupExceptionOnFailure(boolean throwExceptionOnFailure) {
-        if (throwExceptionOnFailure || (devMode && logMissingProperties)) {
+        if (throwExceptionOnFailure || logMissingProperties) {
             context.put(THROW_EXCEPTION_ON_FAILURE, true);
         }
     }
@@ -342,7 +341,7 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     protected boolean shouldLogMissingPropertyWarning(OgnlException e) {
         return (e instanceof NoSuchPropertyException ||
                 (e instanceof MethodFailedException && e.getReason() instanceof NoSuchMethodException))
-        		&& devMode && logMissingProperties;
+        		&& logMissingProperties;
     }
 
     private Object tryFindValue(String expr, Class asType) throws OgnlException {
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -182,7 +182,8 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
 
     private void trySetValue(String expr, Object value, boolean throwExceptionOnFailure, Map<String, Object> context) throws OgnlException {
         context.put(XWorkConverter.CONVERSION_PROPERTY_FULLNAME, expr);
-        context.put(REPORT_ERRORS_ON_NO_PROP, (throwExceptionOnFailure) ? Boolean.TRUE : Boolean.FALSE);
+        context.put(REPORT_ERRORS_ON_NO_PROP, throwExceptionOnFailure || (devMode && logMissingProperties)
+                ? Boolean.TRUE : Boolean.FALSE);
         ognlUtil.setValue(expr, context, root, value);
     }
 
@@ -246,7 +247,7 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     }
 
     protected void setupExceptionOnFailure(boolean throwExceptionOnFailure) {
-        if (throwExceptionOnFailure) {
+        if (throwExceptionOnFailure || (devMode && logMissingProperties)) {
             context.put(THROW_EXCEPTION_ON_FAILURE, true);
         }
     }
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
@@ -217,6 +217,7 @@ public class CompoundRootAccessor implements PropertyAccessor, MethodAccessor, C
             return null;
         }
 
+        Throwable reason = null;
         for (Object o : root) {
             if (o == null) {
                 continue;
@@ -233,22 +234,20 @@ public class CompoundRootAccessor implements PropertyAccessor, MethodAccessor, C
 
             if ((argTypes == null) || !invalidMethods.containsKey(mc)) {
                 try {
-                    Object value = OgnlRuntime.callMethod((OgnlContext) context, o, name, objects);
-
-                    if (value != null) {
-                        return value;
-                    }
+                    return OgnlRuntime.callMethod((OgnlContext) context, o, name, objects);
                 } catch (OgnlException e) {
                     // try the next one
-                    Throwable reason = e.getReason();
+                    reason = e.getReason();
 
-                    if (!context.containsKey(OgnlValueStack.THROW_EXCEPTION_ON_FAILURE) && (mc != null) && (reason != null) && (reason.getClass() == NoSuchMethodException.class)) {
+                    if ((mc != null) && (reason != null) && (reason.getClass() == NoSuchMethodException.class)) {
                         invalidMethods.put(mc, Boolean.TRUE);
-                    } else if (reason != null) {
-                        throw new MethodFailedException(o, name, e.getReason());
                     }
                 }
             }
+        }
+
+        if (context.containsKey(OgnlValueStack.THROW_EXCEPTION_ON_FAILURE)) {
+            throw new MethodFailedException(target, name, reason);
         }
 
         return null;
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
@@ -218,13 +218,13 @@ public class CompoundRootAccessor implements PropertyAccessor, MethodAccessor, C
         }
 
         Throwable reason = null;
+        Class[] argTypes = getArgTypes(objects);
         for (Object o : root) {
             if (o == null) {
                 continue;
             }
 
             Class clazz = o.getClass();
-            Class[] argTypes = getArgTypes(objects);
 
             MethodCall mc = null;
 
@@ -236,12 +236,17 @@ public class CompoundRootAccessor implements PropertyAccessor, MethodAccessor, C
                 try {
                     return OgnlRuntime.callMethod((OgnlContext) context, o, name, objects);
                 } catch (OgnlException e) {
-                    // try the next one
                     reason = e.getReason();
 
-                    if ((mc != null) && (reason != null) && (reason.getClass() == NoSuchMethodException.class)) {
+                    if (reason != null && !(reason instanceof NoSuchMethodException)) {
+                        // method has found but thrown an exception
+                        break;
+                    }
+
+                    if ((mc != null) && (reason != null)) {
                         invalidMethods.put(mc, Boolean.TRUE);
                     }
+                    // continue and try the next one
                 }
             }
         }
--- a/core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
@@ -954,6 +954,33 @@ public class OgnlValueStackTest extends XWorkTestCase {
         assertNull(vs.findValue("@com.nothing.here.Nothing@BLAH"));
     }
 
+    /**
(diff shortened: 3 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
```diff
@@ -233,24 +234,22 @@ public Object callMethod(Map context, Object target, String name, Object[] objec
...
                         invalidMethods.put(mc, Boolean.TRUE);
-                    } else if (reason != null) {
-                        throw new MethodFailedException(o, name, e.getReason());
                     }
                 }
             }
         }
 
+        if (context.containsKey(OgnlValueStack.THROW_EXCEPTION_ON_FAILURE)) {
+            throw new MethodFailedException(target, name, reason);
```
qqu0127: Seems like previously this `MethodFailedException` is thrown when the method is found but failed on execution. Now we are mixing up both cases "no such method" and "method failed" by throwing this exception outside the loop. Is this the expected behavior?
Also, shall we consider break the loop at the point we threw `MethodFailedException`?

yasserzamani:
> Is there a reason we have to set both `devMode` and `logMissingProperties` true in order to enable logging? Could we just use `logMissingProperties` for clarity?

@apache/struts-committers, I agree with @qqu0127 , any objection?

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
yasserzamani: No, previously it was mixed up both cases also so it is same as previous in this behavior. The only difference for it is: Now it is outside of the loop. Being inside of the loop is a bug. I discovered it in #357 - For example `OgnlValueStack.findValue(expr: "someProp", throwExceptionOnFailure: true)` always fail when top of stack doesn't have `someProp` property while it must check also other objects in stack and fail if none of them had this property.

> Also, shall we consider break the loop at the point we threw MethodFailedException?

I think no. AFAIK it must try on all of objects in stack.

qqu0127: Looks ok to me, it makes sense to keep the mixed up behavior. 
Regarding the exception previously at line 248, doesn't that break the loop already? I don't think that would be an issue even `throwExceptionOnFailure` is set true. Because that is only thrown when the property is found, but an exception is thrown in application code (that's why `reason != null`).
Do you propose to search down the stack anyway even we've already "hit"?
The only use case I can think of: we have more than one "hit" on stack, while first one throws runtime exception and second one execute smoothly and return -- yes, in this case we don't want to break the loop.
I'm not super familiar with the coding style and design pattern of `ognl`, is that really what we want? 
Hiding exceptions doesn't look quite right to me...

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
```diff
@@ -233,24 +234,22 @@ public Object callMethod(Map context, Object target, String name, Object[] objec
...
+                    return OgnlRuntime.callMethod((OgnlContext) context, o, name, objects);
                 } catch (OgnlException e) {
                     // try the next one
-                    Throwable reason = e.getReason();
+                    reason = e.getReason();
 
-                    if (!context.containsKey(OgnlValueStack.THROW_EXCEPTION_ON_FAILURE) && (mc != null) && (reason != null) && (reason.getClass() == NoSuchMethodException.class)) {
+                    if ((mc != null) && (reason != null) && (reason.getClass() == NoSuchMethodException.class)) {
                         invalidMethods.put(mc, Boolean.TRUE);
-                    } else if (reason != null) {
```
JCgH4164838Gh792C124B5: Is it certain that dropping the "if (reason != null)" check as part of the decision whether to throw the exception or not won't cause an issue ?

It looks like the old code would "skip" throwing an Exception in that case, no matter what `THROW_EXCEPTION_ON_FAILURE` is set to ?

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
```diff
@@ -233,24 +234,22 @@ public Object callMethod(Map context, Object target, String name, Object[] objec
...
-                        throw new MethodFailedException(o, name, e.getReason());
```
JCgH4164838Gh792C124B5: Was the old logic intentionally limiting the loop to 1 iteration (unless the exception reason was `null` ) ?
Previously it would seem things would _only succeed_ if the method was defined on the object at the top of the stack. 

Will "walking the whole valuestack" potentially introduce a different issue (unexpectedly) ?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -246,7 +247,7 @@ public Object findValue(String expr, boolean throwExceptionOnFailure) {
     }
 
     protected void setupExceptionOnFailure(boolean throwExceptionOnFailure) {
-        if (throwExceptionOnFailure) {
+        if (throwExceptionOnFailure || (devMode && logMissingProperties)) {
```
JCgH4164838Gh792C124B5: I think this is the main reason why `devMode` and `logMissingProperties` should continue to be "linked" (at least in 2.5.x).  Most users probably only want to check for missing properties in devMode (development) anyway.

If `logMissingProperties` becomes a "standalone flag", it could turn on `THROW_EXCEPTION_ON_FAILURE` in a production configuration - and that could have greater repercussions than intended.  Maybe it could be decoupled in 2.6 in a "safe way" ?

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
```diff
@@ -217,6 +217,7 @@ public Object callMethod(Map context, Object target, String name, Object[] objec
             return null;
         }
 
+        Throwable reason = null;
```
JCgH4164838Gh792C124B5: (Oops ... forgot to include this comment in the review).

Assuming the loop logic will continue to process more than 1 iteration, I suggest a small optimization:
 - Move Line **227**:  `Class[] argTypes = getArgTypes(objects);`  -- to Line **219/220** (outside of the loop).
_Should_ be safe as the argument types don't change across the loop iterations.

lukaszlenart:
Would you mind adding a test case where an existing property throws a runtime exception? Such case should not be reported as _missing property_. Basically something has changed in struts <-> ognl flow and `ognl.NoSuchPropertyException` isn't thrown/handled any more. I was trying to figure out what was wrong but failed :(

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
yasserzamani: Now, it doesn't break the loop but previously it did, and I thought it was an issue when `throwExceptionOnFailure` is set true because of casual failed tests #357 - for example it accidentally uncovered that `XWorkConverter.getConversionErrorMessage` seems fails when user set `throwExceptionOnFailure` to true, I think.  right?

I just found all usages of `OgnlValueStack.THROW_EXCEPTION_ON_FAILURE` and found all of them outside loop except here. Thanks for heads up for `reason != null`, it makes sense! But now I'm surprised if `reason != null` is a good decision to distinguish between `property not found` and `user app exception` then why #357 tests fail?! (see prev para) In #357 everything is same as before except `throwExceptionOnFailure` which is hard-coded to true.

yasserzamani:
Hi all,

Thanks all for comments! I tried to fix all of them in a new commit above. The only weakness is when user method body itself throws `NoSuchMethodException` which cannot be fixed nicely without changing ognl itself.

@lukaszlenart , I realized these issues are old. I checked out Struts 2.5 (5/5/2016) and tested it against two issues. 1. `devMode=true&&logMissingProperties=true` doesn't work their either. 2. `throwExceptionOnFailure=true` causes failure on `XWorkConverter.getConversionErrorMessage`.

@qqu0127 , I think in this new commit I was able to address your concerns. Could you check please? and thanks a tone for heads up!

@JCgH4164838Gh792C124B5 , I'm working on your concerns :)

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
yasserzamani: As @qqu0127 mentioned, it seems `reason != null` tries to distinguish between `property not found` and `user app` exception but actually it doesn't work. As far as I could see and debug, OGNL always fill it here
 
https://github.com/jkuhnert/ognl/blob/b6e58c4db776beb20e20d5792227d1da7b4355e0/src/main/java/ognl/OgnlRuntime.java#L1529

in both `NoSuchMethodException` (i.e. property not found) and `InvocationTargetException` (i.e. user app exception).

I realized that this is an old issue, please see this test : https://github.com/apache/struts/blob/d4dd3386cc02833fa7113a496092b2e09a53c9e4/core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java#L958

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
yasserzamani: > Was the old logic intentionally limiting the loop to 1 iteration (unless the exception reason was `null` ) ?
> Previously it would seem things would _only succeed_ if the method was defined on the object at the top of the stack.

In case of `throwExceptionOnFailure=true`, Yes, but it is a wrong behavior. Please see my test at https://github.com/apache/struts/blob/d4dd3386cc02833fa7113a496092b2e09a53c9e4/core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java#L958 which fails on Struts 2.5 (5/5/2016) on L#975.

> Will "walking the whole valuestack" potentially introduce a different issue (unexpectedly) ?

No, it is already current behavior in case of `throwExceptionOnFailure=false`.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
yasserzamani: `throwExceptionOnFailure || logMissingProperties` also works even better. As far as I can see in codes, Struts will check them separately later to check throw any exception and log any message.

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
yasserzamani: 👍 fixed

yasserzamani:
I tested and finally found it more clear and nice in both test and production codes to decouple logMissingProperties from devMode so did it.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -238,6 +245,66 @@ public void testFailOnMissingProperty() {
         }
     }
 
+    public void testLogMissingProperties() {
```
JCgH4164838Gh792C124B5: Since `testNotThrowExceptionOnTopMissingProperty()` added `@since 2.5.21`, do you want the same annotation here ?

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -238,6 +245,66 @@ public void testFailOnMissingProperty() {
...
+                    testAppender.logEvents.get(1).getMessage().getFormattedMessage());
+            assertEquals("Could not find property [missingProp3]!",
+                    testAppender.logEvents.get(2).getMessage().getFormattedMessage());
+        } finally {
+            testAppender.stop();
+            logger.removeAppender(testAppender);
+        }
+    }
+
+    public void testNotLogUserExceptionsAsMissingProperties() {
```
JCgH4164838Gh792C124B5: And `@since 2.5.21` here ?

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java
```diff
@@ -238,6 +245,66 @@ public void testFailOnMissingProperty() {
...
+            vs.findValue("getBite()", false);
+            vs.findValue("getBite()", void.class, false);
+
+            assertEquals(0, testAppender.logEvents.size());
+        } finally {
+            testAppender.stop();
+            logger.removeAppender(testAppender);
+        }
+    }
+
```
JCgH4164838Gh792C124B5: @yasserzamani , do you think adding equivalent tests  `testLogMissingPropertiesFalse()` and `testNotLogUserExceptionsAsMissingPropertiesFalse()` would be a good idea ?

They could set `vs.setLogMissingProperties("false");` and then confirm that nothing was logged.  Might be a good additional check ?

As it stands the changes LGTM, but I think those two additional tests might be useful to catch a future change that accidentally causes the opposite behaviour (i.e. something gets logged when `logMissingProperties` is `false`, but should not).

JCgH4164838Gh792C124B5: Hello @yasserzamani .  Thanks for responding to the feedback and your excellent work on this PR.  👍 .  
Thanks to @qqu0127 for opening the Jira issue and reviewing the PR's changes too.

core/src/main/java/com/opensymphony/xwork2/ognl/accessor/CompoundRootAccessor.java
qqu0127: Oh, you are right. `reason != null` itself is not a complete check. Actually, I fixed this in my workplace integrated with our logging system internally, based on 2.5.17.  And it classifies the exceptions correctly. FYI, we modified `ValueStack`. (we also changed `CompoundRootAccessor`, but differently, so not sure if it works here)
```
  @Override
  public Object handleOgnlException(String expr, boolean throwExceptionOnFailure, OgnlException e) {
    Object ret = findInContext(expr);
    if (ret != null) {
      return ret;
    }
    String errMsg = null;
    if (shouldLogWarning()) {
      if (e.getCause() == null || e.getCause() instanceof NoSuchMethodException) {
        errMsg = String.format("Could not evaluate expression [%s]", expr);
      }
      else {
        // exception in invoked method
        errMsg = String.format("Caught an exception in evaluating expression [%s]", expr);
      }

      logStackTrace(errMsg, e);
    }
    if (throwExceptionOnFailure) {
      throw new XWorkException(e, errMsg);
    }
    return null;
  }
```
And 
```
@Override
  protected Object handleOtherException(String expr, boolean throwExceptionOnFailure, Exception e) {
    if (shouldLogWarning()) {
      if (e.getCause() == null) {
        logStackTrace(String.format("Could not evaluate expression [%s]", expr), e);
      }
      else {
        logStackTrace(String.format("Caught an exception in evaluating expression [%s]", expr), e);
      }
    }
    if (throwExceptionOnFailure) {
      throw new XWorkException(e);
    }
    return findInContext(expr);
  }
```

qqu0127: I'm still not sure if we should remove `elseif (reason != null)`. In my view it changes the flow too much (validity and performance), and it's irrelevant to the issue.

yasserzamani: We should because with that following test fails. I tested and even Struts 2.5 (5/5/2016) with no change also fails on this test: https://github.com/apache/struts/blob/a50af87644320a3ca85283ccab7bc72822cb230b/core/src/test/java/com/opensymphony/xwork2/ognl/OgnlValueStackTest.java#L982
but this is and was a wrong behavior.

lukaszlenart:
LGTM 👍

lukaszlenart:
@yasserzamani do you plan cherry-pick those changes into 2.6?

yasserzamani:
Thanks for thanks and 👍s :)

@lukaszlenart , cherry picked into 2.6 and travis is green :)

lukaszlenart:
💯

PR #364, opened 2019-08-31
Defines a Docker build

lukaszlenart:
This PR switches from a normal Maven Pipeline build to Docker based, plus it fixes build on TravisCI


cnenning:
lgtom 👍

PR #366, opened 2019-09-16
Improved logging for DefaultDispatcherErrorHandler, DefaultStaticContentLoader

JCgH4164838Gh792C124B5:
Improved logging for DefaultDispatcherErrorHandler and DefaultStaticContentLoader.

Provide informational logging for the processing of the two dispatcher classes when an http sendError fails for the two known failure types (IOException, IllegalStateException).

In both circumstances it is beneficial to have the developer aware of the failures via the logs.  These are not events that should be happening on a regular basis, so there is little risk of a log flood.

DefaultStaticContentLoader didn't catch either exception type previously, so changing it to make handling the same as DefaultDispatcherErrorHandler.

For DefaultDispatcherErrorHandler the IOException was caught with no notice of failure previously.  Now there will be an info level log output.
Previously the IllegalStateException was not caught, which resulted in the unrecoverable exception being thrown up to the calling thread, usually resulting in an ugly stacktrace to stdout/stderr.  Now there will be an info level log output instead.

If devMode is true, the info log outputs will include the exception parameter to the log, so a stacktrace can be viewed for more details.  If devMode is false (production mode), then only the exception's tostring() output will be produced.

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java
@@ -97,10 +97,10 @@ public class DefaultDispatcherErrorHandler implements DispatcherErrorHandler {
             response.sendError(code, e.getMessage());
         } catch (IOException e1) {
             // we're already sending an error, not much else we can do if more stuff breaks
-            LOG.info("Unable to send error response, code: " + code + "! (IOException): " + e1);
+            LOG.warn("Unable to send error response, code: {}! (IOException): {}", code, e1.toString());
         } catch (IllegalStateException ise) {
             // Log illegalstate instead of passing unrecoverable exception to calling thread
-            LOG.info("Unable to send error response, code: " + code + "! isCommited: " + response.isCommitted() + " (IllegalStateException): " + ise);
+            LOG.warn("Unable to send error response, code: {}! isCommited: {}. (IllegalStateException): {}", code, response.isCommitted(), ise.toString());
         }
     }
 
@@ -126,10 +126,10 @@ public class DefaultDispatcherErrorHandler implements DispatcherErrorHandler {
                 response.sendError(code, "Unable to show problem report:\n" + exp + "\n\n" + LocationUtils.getLocation(exp));
             } catch (IOException ex) {
                 // we're already sending an error, not much else we can do if more stuff breaks
-                LOG.info("Unable to send error response, code: " + code + "! (IOException): ", ex);  // Stacktrace with DevMode
+                LOG.warn("Unable to send error response, code: {}! (IOException): {}", code, ex);  // Stacktrace with DevMode
             } catch (IllegalStateException ise) {
                 // Log illegalstate instead of passing unrecoverable exception to calling thread
-                LOG.info("Unable to send error response, code: " + code + "! isCommited: " + response.isCommitted() + " (IllegalStateException): ", ise);  // Stacktrace with DevMode
+                LOG.warn("Unable to send error response, code: {}! isCommited: {}. (IllegalStateException): {}", code, response.isCommitted(), ise);  // Stacktrace with DevMode
             }
         }
     }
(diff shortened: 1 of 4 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This potential change is essentially to improve logging around the two default dispatcher classes that process http `sendError()` calls. 

If an `IllegalStateException` arises in 2.5.x during `sendError()` you typically end up with a stacktrace in the application server logs (usually of limited help in non-development contexts).  The current implementation of `DefaultDispatcherErrorHandler` also silently eats any `IOException` during `sendError()` (nothing in the logs).

This PR seeks to provide a more consistent handling for the two conditions when processing `sendError()` in the dispatcher classes (for the two that do so).

No JIRA has been created for this idea yet, as it seemed like it should be reviewed first.  Please take a look and advise what you think (and if it does seem reasonable and should have a JIRA, one can be created and refer back to this PR).

JCgH4164838Gh792C124B5:
The coveralls coverage warning could be addressed by using mock objects ... but might be excessive for what's involved with this PR.  If people think some unit tests should be considered to do so, please advise.

aleksandr-m:
@JCgH4164838Gh792C124B5 Use `{}` notation in logs.

lukaszlenart:
As @aleksandr-m already mentioned please use [`{ }`](https://logging.apache.org/log4j/2.x/manual/messages.html) and `warn` level is better than `info` :)

JCgH4164838Gh792C124B5:
Updated the PR as requested by A. Mashchenko and L. Lenart.

Now uses the preferred `{}` notation for log4j2 log calls and upgrades the outputs to `warn` level.
Originally thought that the `info` level might be easier for developers to mask out if needed, but warnings are probably more appropriate given the subject.

core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java
```diff
@@ -97,6 +97,10 @@ protected void sendErrorResponse(HttpServletRequest request, HttpServletResponse
             response.sendError(code, e.getMessage());
         } catch (IOException e1) {
             // we're already sending an error, not much else we can do if more stuff breaks
+            LOG.warn("Unable to send error response, code: {}! (IOException): {}", code, e1.toString());
+        } catch (IllegalStateException ise) {
+            // Log illegalstate instead of passing unrecoverable exception to calling thread
+            LOG.warn("Unable to send error response, code: {}! isCommited: {}. (IllegalStateException): {}", code, response.isCommitted(), ise.toString());
```
lukaszlenart: You don't have to use `.toString()`, Log4j2 will do it for you, also this is in opposite to lines below :)

aleksandr-m: Don't put `{}` for the exception.
```suggestion
LOG.warn("Unable to send error response, code: {}; isCommited: {}", code, response.isCommitted(), ise);
```

JCgH4164838Gh792C124B5: Hi.

Performing the `toString()` calls was an intentional attempt to avoid Log4j2 converting it into a stacktrace (as opposed to the DevMode flow with the comment "  // Stacktrace with DevMode").  I wasn't sure if the stacktrace at this point would provide a lot of benefit (as opposed to when DevMode is true), so opted to force the short-form.

If both of you prefer just outputting the whole stacktrace all the time, I can do that instead, and make the logging symmetric in both methods.  I suppose since the `DefaultDispatchErrorHandler` can be replaced by a developer who wants custom behaviour, the default implementation can just dump the full stacktrace (since they can change it if they want or need to).

There are configuration options in Log4j2 to configure `throwable` output formatting but it only seems to be possible at the appender level (which would require some configuration gymnastics to achieve what forcing the `.toSring()` attempts here ... if it's even possible).  If anyone knows a Log4j2 configuration that could modify the output only for a specific class/method combination please feel free to share.  :smile:

JCgH4164838Gh792C124B5:
Hello again.

The PR's been updated again with changes that should address the concerns that were raised.  Some basic unit tests were added too (in an attempt to prove the new code functions as intended).  I'll look at adding a similar unit test for the other PR when I clean it up (in a day or too I hope).

core/src/main/java/org/apache/struts2/dispatcher/DefaultDispatcherErrorHandler.java
lukaszlenart: Oh, nice, I didn't know this :)

lukaszlenart:
I think this is ready, I'm ok with all the changes and the explanation :) LGTM 👍

PR #367, opened 2019-09-18
Minor improvement proposed for ServletRedirectResult sendRedirect()

JCgH4164838Gh792C124B5:
Minor improvement proposed for ServletRedirectResult sendRedirect()

Supply log warning when an IOException or IllegalStateException occurs to better allow developers to track the failed redirect location and status details.  The exceptions are re-thrown to ensure existing flow-control behaviour is preserved.

When getWriter() is called, utilize a finally block to ensure close is called.

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/ServletRedirectResult.java
+++ b/core/src/main/java/org/apache/struts2/result/ServletRedirectResult.java
@@ -263,11 +263,11 @@ public class ServletRedirectResult extends StrutsResultSupport implements Reflec
                 }
             }
         } catch (IOException ioe) {
-            LOG.warn("Unable to redirect to: {}, code: {}! (IOException): {}", finalLocation, statusCode, ioe.toString());
-            throw ioe;  // Re-throw required to preserve existing default behaviour
+            LOG.warn("Unable to redirect to: {}, code: {}; {}", finalLocation, statusCode, ioe);
+            throw ioe;  // Re-throw required to preserve existing default behaviour (no stacktrace in above warn for this reason)
         } catch (IllegalStateException ise) {
-            LOG.warn("Unable to redirect to: {}, code: {}! isCommited: {}. (IllegalStateException): {}", finalLocation, statusCode, response.isCommitted(), ise.toString());
-            throw ise;  // Re-throw required to preserve existing default behaviour
+            LOG.warn("Unable to redirect to: {}, code: {}; isCommited: {}; {}", finalLocation, statusCode, response.isCommitted(), ise);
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This potential change is essentially to improve logging around the code that processes http `sendRedirect()` calls (`ServletRedirectResult` is the only class that does so within the core classes).

It shares some similarities to #366, but to preserve existing behaviour it must re-throw the exceptions it catches.  The additional logging should be helpful without too much risk of log flooding.  If it proves an issue in some specific scenarios it could be controlled by logging configuration (or by overriding `sendRedirect()` with a custom `ServletRedirectResult` implementation.

Please take a look and advise what you think (and if worth considering, if it should have a JIRA created for it).

core/src/main/java/org/apache/struts2/result/ServletRedirectResult.java
```diff
@@ -250,13 +250,24 @@ protected void doExecute(String finalLocation, ActionInvocation invocation) thro
...
+                    response.getWriter().write(finalLocation);
+                } finally {
+                    response.getWriter().close();
+                }
+            }
+        } catch (IOException ioe) {
+            LOG.warn("Unable to redirect to: {}, code: {}! (IOException): {}", finalLocation, statusCode, ioe.toString());
+            throw ioe;  // Re-throw required to preserve existing default behaviour
+        } catch (IllegalStateException ise) {
+            LOG.warn("Unable to redirect to: {}, code: {}! isCommited: {}. (IllegalStateException): {}", finalLocation, statusCode, response.isCommitted(), ise.toString());
```
lukaszlenart: Please do  not use `.toString`, Log4j2 will do it for you :)

aleksandr-m: Don't put `{}` for the exception.

JCgH4164838Gh792C124B5: Hi Again.

The reasoning for the `toString()` here was pretty much the same as https://github.com/apache/struts/pull/366#discussion_r326865745.

The same general logic holds here, so we could just output the full stacktrace and then users who want or need different behaviour can override the `ServletRedirectResult` if needed.

JCgH4164838Gh792C124B5:
Hi.

This PR was updated with changes that better align with the feedback.  The new unit tests also help reassure the new code works as advertised (should make coveralls happier too).

lukaszlenart:
Superb, LGTM 👍

PR #371, opened 2019-10-01
WW-5041 Upgrade to OGNL 3.1.26 and adapt to its new features

yasserzamani:
It is green on my local builds. When OGNL 3.1.26 released, this one would be green either.

If merged, Struts would be less brittle when user applies an OgnlSecurityManager. Also Struts would block OGNL expressions longer than 200 characters (albeit it's unit-tested that it's still able to set long values e.g. a long email text).

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -184,6 +184,19 @@ public class OgnlUtil {
         this.disallowProxyMemberAccess = Boolean.parseBoolean(disallowProxyMemberAccess);
     }
 
+    /**
+     * @param maxLength Injects the Struts OGNL maximum expression length.
+     */
+    @Inject(value = StrutsConstants.STRUTS_OGNL_EXPRESSION_MAX_LENGTH, required = false)
+    protected void applyExpressionMaxLength(String maxLength) {
+        if (maxLength == null || maxLength.isEmpty()) {
+            // user is going to disable this functionality
+            Ognl.applyExpressionMaxLength(null);
+        } else {
+            Ognl.applyExpressionMaxLength(Integer.parseInt(maxLength));
+        }
+    }
+
     public boolean isDisallowProxyMemberAccess() {
         return disallowProxyMemberAccess;
     }
@@ -755,6 +768,9 @@ public class OgnlUtil {
             setValue(name, context, o, value);
         } catch (OgnlException e) {
             Throwable reason = e.getReason();
+            if (reason instanceof SecurityException) {
+                LOG.warn("Could not evaluate this expression due to security constraints: [{}]", name, e);
+            }
             String msg = "Caught OgnlException while setting property '" + name + "' on type '" + o.getClass().getName() + "'.";
             Throwable exception = (reason == null) ? e : reason;
 
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
@@ -204,6 +204,9 @@ public class OgnlValueStack implements Serializable, ValueStack, ClearableValueS
     }
 
     protected void handleOgnlException(String expr, Object value, boolean throwExceptionOnFailure, OgnlException e) {
+        if (e.getReason() instanceof SecurityException) {
+            LOG.warn("Could not evaluate this expression due to security constraints: [{}]", expr, e);
+        }
     	boolean shouldLog = shouldLogMissingPropertyWarning(e);
     	String msg = null;
     	if (throwExceptionOnFailure || shouldLog) {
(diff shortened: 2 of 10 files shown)


JCgH4164838Gh792C124B5:
Hi Yasser.  I've only had time to take a quick glance at the PR, but I'll try to take a closer look in a few days.  Not sure whether the default block-length should be a bit longer (e.g. 256) or a bit shorter (e.g. 128, 160) ... hard to say what the "average" application might need.

yasserzamani:
Thank you so much!

Just FYI, my analyze of so far received security mail list reports has
led to consequent 200 :)

p.s. I always try to practice my new vocabulary/grammar in my
communications because it has always been my best method to persist them
in memory! That was my first try to use `consequent` as an adjective :)
excuse me if that's or anything is/was/will be wrong :(

lukaszlenart:
I didn't release OGNL 3.1.26 yet

yasserzamani:
Yes I wanted to share the work sooner to catch comments and ideas earlier. If you released that then this would build again and being green (like my local one).

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -184,6 +184,19 @@ protected void setDisallowProxyMemberAccess(String disallowProxyMemberAccess) {
         this.disallowProxyMemberAccess = Boolean.parseBoolean(disallowProxyMemberAccess);
     }
 
+    /**
+     * @param maxLength Injects the Struts OGNL maximum expression length.
+     */
+    @Inject(value = StrutsConstants.STRUTS_OGNL_EXPRESSION_MAX_LENGTH, required = false)
+    protected void applyExpressionMaxLength(String maxLength) {
```
JCgH4164838Gh792C124B5: Since `applyExpressionMaxLength()` may produce an exception, maybe the set attempts should be guarded:
```
try {
...
} catch (Exception ex) {
    LOG.warn("Unable to set OGNL Expression Max Length {}.", maxLength, ex);
    // Decide whether to only log exception or also re-throw with: "throw ex;"
}
```

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -204,6 +204,9 @@ protected void handleRuntimeException(String expr, Object value, boolean throwEx
     }
 
     protected void handleOgnlException(String expr, Object value, boolean throwExceptionOnFailure, OgnlException e) {
```
JCgH4164838Gh792C124B5: Null-guard:
` if (e != null && e.getReason() instanceof SecurityException) {`
so that this change doesn't risk creating an NPE where one couldn't happen before if the passed `e` parameter is null.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -326,7 +329,12 @@ private Object tryFindValueWhenExpressionIsNotNull(String expr, Class asType) th
     }
 
     protected Object handleOgnlException(String expr, boolean throwExceptionOnFailure, OgnlException e) {
-        Object ret = findInContext(expr);
+        Object ret = null;
```
JCgH4164838Gh792C124B5: Null-guard (same reasoning as above):
` if (e != null && e.getReason() instanceof SecurityException) {`

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -219,4 +219,8 @@ struts.ognl.enableExpressionCache=true
 ### Indicates if Dispatcher should handle unexpected exceptions by calling sendError()
 ### or simply rethrow it as a ServletException to allow future processing by other frameworks like Spring Security
 struts.handle.exception=true
+
+### applies maximum length allowed on OGNL expressions for security enhancement
```
JCgH4164838Gh792C124B5: The value `200` still seems a bit short, and likely to be exceeded by a fair number of legitimate expressions.
Maybe `256` characters ( 0.25 KB ) would be a little more flexible for a starting point (but still short enough to have some benefit) ?

core/src/main/resources/struts-default.xml
```diff
@@ -55,13 +55,18 @@
     <constant name="struts.excludedPackageNames"
               value="
                 ognl.,
+                java.io.,
```
JCgH4164838Gh792C124B5: If you're considering including `java.io`, maybe `java.net` and `java.nio` should also be included in the default exclusions ?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
yasserzamani: AFAIK Struts didn't use to operate like that. It fails in such conditions with fatal BadConfException and compels user to fix it.

JCgH4164838Gh792C124B5: Hi Yasser.

I don't think Struts configuration normally operates like that, but OGNL behaviour around the applyExpressionMaxLength could change in the future.  The log warning would make it clearer what failed either way (from a Struts perspective, including what maximum length was being requested when it failed) ... and the exception can still be re-thrown to compel the user to fix it.

The "controlled warning" seemed like it might be of some use, even if the exception is still re-thrown.  :smile:

yasserzamani:
Hi @JCgH4164838Gh792C124B5, sorry for delayed response, I have been out
of PC since last week, I'll add your proposed fixes tomorrow :)

yasserzamani:
Hi @lukaszlenart , Is OGNL 3.1.26 going to release before Struts 2.5.21?

lukaszlenart:
I can release it in minutes but I thought it would be good to include [this change](https://github.com/jkuhnert/ognl/pull/87)

yasserzamani:
Cool! They LGTM :+1 please.

lukaszlenart:
This PR will also fix [WW-5031](https://issues.apache.org/jira/browse/WW-5031)

lukaszlenart:
LGTM 👍

PR #373, opened 2019-10-14
Proposed fix for WW-5022 (escape html tag body control flag)

JCgH4164838Gh792C124B5:
Proposed fix for WW-5022 (escape html tag body control flag)
- Added escapeHtmlBody parameter to s:a and s:submit tags.
- No other tags appear to require this feature (but can be added to any component).
- Added new unit tests for escapeHtmlBody (and usesBody for component).
- Fixed broken s:a tags in ShowCase app.

### DIFF
--- a/apps/showcase/src/main/webapp/WEB-INF/decorators/main.jsp
+++ b/apps/showcase/src/main/webapp/WEB-INF/decorators/main.jsp
@@ -107,7 +107,7 @@
 
         <div id="navbar" class="navbar-collapse collapse">
                 <ul class="nav navbar-nav">
-                    <li><s:a value="%{home}"><i class="glyphicon glyphicon-home"></i> Home</s:a></li>
+                    <li><s:a value="%{home}" escapeHtmlBody="false"><i class="glyphicon glyphicon-home"></i> Home</s:a></li>
                     <li class="dropdown">
                         <a href="#" class="dropdown-toggle" data-toggle="dropdown">
                             <i class="glyphicon glyphicon-cog"></i> Configuration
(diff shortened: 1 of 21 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This PR is a starting point for a fix to WW-5022 (automatic html escaping of some tag bodies breaks content display in 2.6.x).

Previous commentary in the JIRA by L. Lenart and Y. Zamani identified `s:a` and `s:submit` tags as being impacted and I could not find any others during this PR's construction.  If anyone can spot other tags that may need the new parameter please advise with a comment.

The FreeMarker template changes seem to be OK, but please feel free to suggest a more efficient alternative.

Thanks.

lukaszlenart:
Nice work! My only concern is maybe we should set the defaults to `false` to keep backward compatibility, wdyt?

JCgH4164838Gh792C124B5:
Hello @lukaszlenart.

When reading the comments in the JIRA it seemed that both you and Yasser were more in favour of HTML-escaping by default, with an "opt-out" for the tags that might need it.  The PR was configured based on that interpretation.

The defaults could be set to `false`, but maybe auto-escaping by default (defaults set to `true` as in the current PR state) is a little safer for content-handling in 2.6 ?

If 2.6.0 has the defaults set to `true` but developer feedback requests it be changed it could be flipped `false` in a later 2.6.x release (without impacting anyone who explicitly set a desired override value in their tags).  If there's a 2.5.x to 2.6.x migration guide this change could be highlighted there.

The default can be left as-is (`true`) or changed (`false`) can be set either way to start with in the PR, just let me know what you prefer.

lukaszlenart:
@JCgH4164838Gh792C124B5 thanks a lot for this in deep explanation, I think you've pointed it out and I agree that leaving `true` as a default value makes perfect sense :)

lukaszlenart:
LGTM 👍

PR #379, opened 2019-11-08
WW-5045 Update jasperreports to 6.10.0

sepe81:
and switch exclude from jackson-annotation to jackson-annotations
b/c the artifact name changed


lukaszlenart:
Would you mind opening a task in JIRA (type `Dependency`) and link it here?

lukaszlenart:
Nice, LGTM 👍

PR #382, opened 2019-11-18
WW-5048 Update various dependencies to newest version

sepe81:


lukaszlenart:
Nice, would you mind creating a jira ticket and specify all these versions in there? That helps when we are preparing a release notes.

lukaszlenart:
LGTM 👍

PR #385, opened 2019-11-30
Fix incorrect encoding strings in FileUploadInterceptorTest

JCgH4164838Gh792C124B5:
Fix incorrect encoding strings provided in FileUploadInterceptorTest
- Tests were using "text/html" which is not a valid character encoding. No impact currently, but the tests could fail in the future if this is not corrected.  If the code is re-used elsewhere for a test of an action a
valid encoding will be needed.
- Switch to a predefined constant Charset to get the string instead.

### DIFF
--- a/core/src/test/java/org/apache/struts2/interceptor/FileUploadInterceptorTest.java
+++ b/core/src/test/java/org/apache/struts2/interceptor/FileUploadInterceptorTest.java
@@ -40,6 +40,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


sepe81:
I wonder if it would be enough to integrate those kind of fixes into the master branch. This would minimize the number of changes for maintenance branches.

JCgH4164838Gh792C124B5:
Hi.  This is a minor correction for future-proofing the `FileUploadInterceptorTest`.
The incorrect encoding was discovered when trying to build a unit test of an action with a mock file upload based on this test  (and `UnsupportedEncodingException` arose).
It will be easy to cherry-pick into 2.6.x as well.

JCgH4164838Gh792C124B5:
Hi @sepe81.  I think I see your point concerning master vs. maintenance branches, but since it was found within 2.5.x first I figured it should be fixed there first.  A PR for the master branch alone could be done, but a fix in 2.5.x and cherry-pick into 2.6 should be pretty simple for this one.
I'm fine either way ... we can wait and see if anyone else prefers one over the other.

sepe81:
Thanks for the quick reply @JCgH4164838Gh792C124B5 -- I am OK either way, too. Just wanted to express my thoughts on that AND hope to see version 2.6 one day. ;-)

lukaszlenart:
@sepe81 is right, please focus on Struts 2.6 :) I rather do not expect to release one more 2.5.x version.

PR #391, opened 2019-12-26
[WW-5003] Drops XWorkException and uses StrutsException instead

lukaszlenart:
Implements [WW-5003](https://issues.apache.org/jira/browse/WW-5003)


yasserzamani:
LGTM 👍 Thanks!

PR #390, opened 2019-12-21
WW-5050 cherry pick #371 and related commits

yasserzamani:
Upgrade to OGNL 3.2.12 and adapt to its new features.

PR #392, opened 2020-01-11
Minor follow-up to WW-5050 changes to ContainerImpl

JCgH4164838Gh792C124B5:
Minor follow-up to WW-5050 changes to ContainerImpl.
- Made isStatic() utility method static, added javadoc comment.
- Refactored isNotPublic() into isPublicForReflection() with negation moved
to calling locations.  Also added javadoc comment.
- Added final declaration to several variables used in methods.
- Added @Override annotation to overridden methods.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/inject/ContainerImpl.java
+++ b/core/src/main/java/com/opensymphony/xwork2/inject/ContainerImpl.java
@@ -37,7 +37,7 @@ class ContainerImpl implements Container {
 
     ContainerImpl(Map<Key<?>, InternalFactory<?>> factories) {
         this.factories = factories;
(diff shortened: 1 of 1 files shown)


JCgH4164838Gh792C124B5:
Hello Struts Team.

Taking a second look at the changes to `ContainerImpl` made in PR #390 the meaning of `isNotPublic()` was not completely clear based on the method name.  This PR aims to make the intent of the method clearer and make it a "positive state method" like `isStatic()`.

The other changes were just minor ones (finality, missing override annotations and making `isStatic()` itself static).

The changes are small enough that it seemed reasonable to attach this PR to the existing WW JIRA rather than creating a new one just for it (especially if people don't think the changes are reasonable).

PR #393, opened 2020-01-12
[WW-5054] debug=browser

lukaszlenart:
To test this enable `devMode` and add `?debug=browser` to a url, eg.: `http://localhost:8080/struts2-showcase/?debug=browser`

Fixes [WW-5054](https://issues.apache.org/jira/browse/WW-5054)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -65,18 +65,23 @@ public class OgnlUtil {
     private Set<Pattern> excludedPackageNamePatterns;
     private Set<String> excludedPackageNames;
 
+    private Set<Class<?>> devModeExcludedClasses;
+    private Set<Pattern> devModeExcludedPackageNamePatterns;
+    private Set<String> devModeExcludedPackageNames;
+
     private Container container;
     private boolean allowStaticFieldAccess = true;
     private boolean allowStaticMethodAccess;
     private boolean disallowProxyMemberAccess;
 
     public OgnlUtil() {
-        excludedClasses = new HashSet<>();
-        excludedPackageNamePatterns = new HashSet<>();
-        excludedPackageNames = new HashSet<>();
-        excludedClasses = Collections.unmodifiableSet(excludedClasses);
-        excludedPackageNamePatterns = Collections.unmodifiableSet(excludedPackageNamePatterns);
-        excludedPackageNames = Collections.unmodifiableSet(excludedPackageNames);
+        excludedClasses = Collections.unmodifiableSet(new HashSet<>());
+        excludedPackageNamePatterns = Collections.unmodifiableSet(new HashSet<>());
+        excludedPackageNames = Collections.unmodifiableSet(new HashSet<>());
+
+        devModeExcludedClasses = Collections.unmodifiableSet(new HashSet<>());
+        devModeExcludedPackageNamePatterns = Collections.unmodifiableSet(new HashSet<>());
+        devModeExcludedPackageNames = Collections.unmodifiableSet(new HashSet<>());
     }
 
     @Inject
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -818,21 +787,25 @@ void internalSetProperty(String name, Object value, Object o, Map<String, Object
...
         if (resolver == null) {
             resolver = container.getInstance(CompoundRootAccessor.class);
         }
 
         SecurityMemberAccess memberAccess = new SecurityMemberAccess(allowStaticMethodAccess, allowStaticFieldAccess);
-        memberAccess.setExcludedClasses(excludedClasses);
-        memberAccess.setExcludedPackageNamePatterns(excludedPackageNamePatterns);
-        memberAccess.setExcludedPackageNames(excludedPackageNames);
-        memberAccess.setDisallowProxyMemberAccess(disallowProxyMemberAccess);
+        if (devMode) {
```
JCgH4164838Gh792C124B5: Hello @lukaszlenart.  Allowing `devMode` to disable the exclusion processing makes me a little nervous.

For one thing, _it could produce confusion_ when something works fine locally with `devMode` `true` and then fails due to default exclusions when `devMode` is `false` for a deployed version.  It would also make running `devMode` potentially _riskier_ than it currently is with the exclusion mechanism set/active.

If a developer wants to access classes blocked by the exclusion mechanism (even in `devMode`), they can already modify their local development `struts.xml` configuration to override/set "`struts.excludedClasses`, `struts.excludedPackageNames`, `struts.excludedPackageNamePatterns`" as necessary for that purpose.  Once they are done their testing they can remove the override.

Maybe there could be a compromise in introducing a concept of "`devMode.excludedClasses`, `devMode.excludedPackageNames`, `devMode.excludedPackageNamePatterns`" ?  When in `devMode` those exclusion lists could be used instead of the standard `struts.excludedXXX` ones, with defaults that are slightly more relaxed than the standard ones but still restrictive ?  That would allow some flexibility between modes, still allow for explicit restrictions in devMode (as restrictive or permissive as the developer chooses) and clearly show what is restricted.

What do you think ?

lukaszlenart: Yeah... disabling all the checks isn't the best idea, but the suggestion with `devMode.excludedClasses` and so looks promising, let me add such change, thanks!

core/src/main/java/org/apache/struts2/interceptor/debugging/DebuggingInterceptor.java
```diff
@@ -210,24 +215,24 @@ public void beforeResult(ActionInvocation inv, String actionResult) {
                         public void beforeResult(ActionInvocation inv, String actionResult) {
                             String rootObjectExpression = getParameter(OBJECT_PARAM);
                             if (rootObjectExpression == null)
-                                rootObjectExpression = "#context";
+                                rootObjectExpression = "top";
```
JCgH4164838Gh792C124B5: Hi.  The change to `top` seems OK, but is there a risk it might break anything (vs `#context`) ?

Also, would adding `{ }` to the if block here as well as https://github.com/apache/struts/blob/1a0ef45160eaf3e66d828dfa363a314e690b0249/core/src/main/java/org/apache/struts2/interceptor/debugging/DebuggingInterceptor.java#L233 be something to consider as well ?

core/src/main/java/org/apache/struts2/config/entities/ConstantConfig.java
```diff
@@ -257,6 +260,9 @@ private String classesToString(Set<Class<?>> classes) {
         map.put(StrutsConstants.STRUTS_EXCLUDED_CLASSES, classesToString(excludedClasses));
         map.put(StrutsConstants.STRUTS_EXCLUDED_PACKAGE_NAME_PATTERNS, StringUtils.join(excludedPackageNamePatterns, ','));
         map.put(StrutsConstants.STRUTS_EXCLUDED_PACKAGE_NAMES, StringUtils.join(excludedPackageNames, ','));
+        map.put(StrutsConstants.STRUTS_DEV_MODE_EXCLUDED_CLASSES, classesToString(devModeExcludedClasses));
```
JCgH4164838Gh792C124B5: LGTM. 👍 

Your implementation gives developers the flexibility to change exclusions in a clear and explicit manner for `devMode` true.

core/src/main/java/org/apache/struts2/interceptor/debugging/DebuggingInterceptor.java
lukaszlenart: Access to `#context` was removed some times ago, I switched to `action` which is a better choice. And added `{ }` :)

lukaszlenart:
🙇

PR #394, opened 2020-02-01
[WW-5047] Upgrades to VelocityEngine 2.1 and VelocityTools 3.0

lukaszlenart:
Implements [WW-5047](https://issues.apache.org/jira/browse/WW-5047)

### DIFF
--- a/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
+++ b/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
@@ -22,21 +22,23 @@ import org.apache.velocity.VelocityContext;
 
 import com.opensymphony.xwork2.util.ValueStack;
 
+import java.util.List;
+
 public class StrutsVelocityContext extends VelocityContext {
 
-    ValueStack stack;
-    VelocityContext[] chainedContexts;
+    private ValueStack stack;
+    private List<VelocityContext> chainedContexts;
 
     public StrutsVelocityContext(ValueStack stack) {
         this(null, stack);
     }
 
-    public StrutsVelocityContext(VelocityContext[] chainedContexts, ValueStack stack) {
+    public StrutsVelocityContext(List<VelocityContext> chainedContexts, ValueStack stack) {
         this.chainedContexts = chainedContexts;
         this.stack = stack;
     }
 
-    public boolean internalContainsKey(Object key) {
+    public boolean internalContainsKey(String key) {
         boolean contains = super.internalContainsKey(key);
 
         // first let's check to see if we contain the requested key
@@ -46,13 +48,13 @@ public class StrutsVelocityContext extends VelocityContext {
 
         // if not, let's search for the key in the ognl value stack
         if (stack != null) {
-            Object o = stack.findValue(key.toString());
+            Object o = stack.findValue(key);
 
             if (o != null) {
                 return true;
             }
 
-            o = stack.getContext().get(key.toString());
+            o = stack.getContext().get(key);
             if (o != null) {
                 return true;
             }
--- a/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
+++ b/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
@@ -22,6 +22,8 @@ import org.apache.velocity.VelocityContext;
 
 import com.opensymphony.xwork2.util.ValueStack;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 public class StrutsVelocityContext extends VelocityContext {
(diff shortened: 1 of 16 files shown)


plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -161,32 +164,24 @@ public Context createContext(ValueStack stack, HttpServletRequest req, HttpServl
...
-        }
-        List contextList = new ArrayList();
-        for (int i = 0; i < chainedContextNames.length; i++) {
-            String className = chainedContextNames[i];
-            try {
-                VelocityContext velocityContext = (VelocityContext) objectFactory.buildBean(className, null);
-                contextList.add(velocityContext);
-            } catch (Exception e) {
-                LOG.warn("Warning. {} caught while attempting to instantiate a chained VelocityContext, {} -- skipping", e.getClass().getName(), className);
+    protected List<VelocityContext> prepareChainedContexts(HttpServletRequest servletRequest, HttpServletResponse servletResponse, Map<String, Object> extraContext) {
```
JCgH4164838Gh792C124B5: Similar to the constructor involving `VelocityContext[]`, since this method is `protected`, changing the return type might introduce a backwards-compatibility issue for someone (but less likely than the constructor).

Looks like there's no way to avoid this unless the new method uses a different parameter signature than the old one does.  One (ugly) way this might work would be to do something like  this for a backwards-compatible version:
```
// Reorder parameters in new method (different signature from old)
protected List<VelocityContext> prepareChainedContexts(Map<String, Object> extraContext, HttpServletRequest servletRequest, HttpServletResponse servletResponse) {
    ... // The rest of the new method
}

protected VelocityContext[] prepareChainedContexts(HttpServletRequest servletRequest, HttpServletResponse servletResponse, Map<String, Object> extraContext) {
    return prepareChainedContexts(extraContext, servletRequest, servletResponse).toArray();
}
// And change the other internal calls to use the new parameter order, just like the backwards-compatible call.
```

It is _probably not worth the hassle_, but I figured it was worth mentioning.  :smile:

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -161,32 +164,24 @@ public Context createContext(ValueStack stack, HttpServletRequest req, HttpServl
...
             }
         }
-        if (contextList.size() > 0) {
-            VelocityContext[] extraContexts = new VelocityContext[contextList.size()];
-            contextList.toArray(extraContexts);
-            return extraContexts;
-        } else {
-            return null;
-        }
+        return contextList;
```
JCgH4164838Gh792C124B5: Possible (but _unlikely_) that someone extended the class and was expecting the null to detect empty `contextList` or `this.chainedContextNames`.

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -161,32 +164,24 @@ public Context createContext(ValueStack stack, HttpServletRequest req, HttpServl
      * perform any initialization of the contexts.  All that must be done in the
      * context itself.
      *
-     * @param servletRequest the servlet request object
+     * @param servletRequest  the servlet request object
```
JCgH4164838Gh792C124B5: Was this an unintentional whitespace change ?

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -161,32 +164,24 @@ public Context createContext(ValueStack stack, HttpServletRequest req, HttpServl
...
      * @param servletResponse the servlet response object
-     * @param extraContext map with extra context
+     * @param extraContext    map with extra context
```
JCgH4164838Gh792C124B5: Was this an unintentional whitespace change ?

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
```diff
@@ -22,21 +22,23 @@
...
-    VelocityContext[] chainedContexts;
+    private ValueStack stack;
+    private List<VelocityContext> chainedContexts;
 
     public StrutsVelocityContext(ValueStack stack) {
         this(null, stack);
     }
 
-    public StrutsVelocityContext(VelocityContext[] chainedContexts, ValueStack stack) {
+    public StrutsVelocityContext(List<VelocityContext> chainedContexts, ValueStack stack) {
```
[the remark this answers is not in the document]
lukaszlenart: Good point, I will add the new constructor and mark the old one as `@Depreacted`

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -120,22 +125,19 @@ public VelocityEngine getVelocityEngine() {
      * </ul>
      *
      * @param stack the current {@link ValueStack}
-     * @param req the current HttpServletRequest
-     * @param res the current HttpServletResponse
+     * @param req   the current HttpServletRequest
```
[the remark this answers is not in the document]
lukaszlenart: This is IDEA's auto-formatting defaults where all params' descriptions are aligned to a one margin.

```
...
     * @param stack the current {@link ValueStack}
     * @param req   the current HttpServletRequest
     * @param res   the current HttpServletResponse
     * @return a new StrutsVelocityContext
```

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -120,22 +125,19 @@ public VelocityEngine getVelocityEngine() {
...
+     * @param res   the current HttpServletResponse
```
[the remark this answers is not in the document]
lukaszlenart: Same as above

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
lukaszlenart: The same case with auto-format

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
lukaszlenart: Yeah... it's hard to keep backward-compatibility all the time, that's why those changes were introduced in a major Struts version 2.6.

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
lukaszlenart: I changed JavaDoc to reflect the new behaviour

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -447,21 +440,18 @@ protected VelocityEngine newVelocityEngine(ServletContext context) {
      * <li>we need to define the various Struts custom user directives such as #param, #tag, and #bodytag</li>
      * </ul>
      *
-     * @param context the servlet context
+     * @param context    the servlet context
```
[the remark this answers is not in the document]
lukaszlenart: Yes, the same as others

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
JCgH4164838Gh792C124B5: OK.  If it really creates an issue for someone in 2.6 they can always propose a solution.  :smile:
Interesting to note the IDEA IDE formatting behaviour as well.

PR #398, opened 2020-03-22
Proposed implementation to address WW-5030.

JCgH4164838Gh792C124B5:
Proposed implementation to address WW-5030.
Introduce new junit-portlet plugin.
- Utilizes a copy of org.springframework.mock.web.portlet package from Spring 4.3.x, refactored to org.apache.struts2.mock.web.portlet.  Also copies one interface from org.apache.struts2.mock.web.portlet.  The package-info.java files identify the package renaming.
- Provides test cases for the mock objects.
- Modifies the portlet plugin to use the junit-portlet plugin as a dependency.  Adds a small boundary case fix for PortletRequestMap.


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

Initially considered adding the mock objects to the junit-plugin itself, but that increases the size of its jar for a set of mocks that are only really intended for portlet unit tests.  So ... this change proposes to use a new plugin (junit-portlet) to try and limit the dependencies to a separate plugin.

Please note that the mock objects in this PR are directly from Spring 4.3.x (Apache-2.0 license) and retain their original form (the only change is a package name refactoring).  The package-info files explains package renaming and the reasoning as a comment within them (to make the change clear).
There also is one minor change (noted by a comment) in MockActionResponse for a constructor that always failed.

Beyond the above, the only new code introduced is a unit test module to test the mock objects themselves.

If anyone notes any issues with the above (compliance or otherwise), please advise.

lukaszlenart:
Astonish! 👏 💯

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

I think what you say above with respect to refactoring this PR to make it a `portlet-mocks-plugin` instead makes sense (and may address a concern raised via comment in the Jira).

I will attempt to make that change in the next few days and if successful, I will push the changes in a new commit.  Thanks for reviewing and providing the suggestions for refactoring.  :smile:

lukaszlenart:
We can merge this PR as is, and then you can refactor it if needed - what works better :)

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

A refactoring attempt was already underway when I read the comment, so I continued with it anyway.  :wink:

The PR has been updated and now reflects a `portlet-mocks-plugin` that hopefully has minimized the dependencies.  Tried a couple of strategies to deal with cycles in project but the current one was the only one that worked-out.

The copy-back/update for my local repository had a few mistakes, so the commit had to be amended a few times ... :sweat:

Also confirmed that both the original version and the refactored version of the plugin worked with the test reproducer project provided with the Jira using Spring 4.3.x and 5.2.x. (after modifying its pom to use the new plugin).

The latest revision seems cleaner than the original, but let me know what you think.

PR #400, opened 2020-04-08
[WW-5065] Removing unnecessary part of AbstractMatcher#replaceParameters

atkaiser:
For [WW-5065] fixing the bug in the AbstractMatcher#replaceParameters method and adding a test case for it.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
@@ -158,14 +158,6 @@ public abstract class AbstractMatcher<E> implements Serializable {
             map.put(entry.getKey(), convertParam(entry.getValue(), vars));
         }
         
-        //the values map will contain entries like name->"Lex Luthor" and 1->"Lex Luthor"
-        //now add the non-numeric values
-        for (Map.Entry<String,String> entry: vars.entrySet()) {
-            if (!NumberUtils.isCreatable(entry.getKey())) {
-                map.put(entry.getKey(), entry.getValue());
-            }
-        }
-        
         return map;
     }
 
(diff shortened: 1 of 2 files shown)


lukaszlenart:
@atkaiser this looks good to me, thanks for a unittest 👏 Did you have a chance this change in ypur project?

@aleksandr-m do you have any other concerns?

JCgH4164838Gh792C124B5:
Hello.

This PR's change seems to make sense based on the discussion in the JIRA as well the name/comments of the `replaceParameters()` method of `AbstractMatcher`.

One concern could be that some existing Struts 2.5.x applications may rely (directly or indirectly) on the previous behaviour that places (some) wildcard parameters into the result.  If that is a possibility, maybe for 2.5.x this change should be made the default, but followed-up by another PR with an opt-in mechanism to re-enable the old behaviour ?

For 2.6 the new behaviour from this PR should probably be the only option (but an opt-in for the old behaviour could be added later if people really demanded it).

Thoughts ?

lukaszlenart:
@JCgH4164838Gh792C124B5 that would be nice, @atkaiser if that's ok I can introduce a flag to allow disable this "feature"

atkaiser:
@lukaszlenart That sounds fine to me.

Unfortunately I haven't gotten a chance to try it out on my project yet and probably won't for a couple of weeks.

lukaszlenart:
Ok, nice. I wonder how to name this new constant, eg.: `struts.matcher.expandParameters=true`?

JCgH4164838Gh792C124B5:
Hi.  Here's a couple of alternate suggestions for a new constant name to enable the old behaviour:

`struts.matcher.addParametersToMapDuringReplace=true`
`struts.matcher.replaceAndAddParameters=true`

The first one above is pretty clear (if a bit "wordy"), but whatever constant name Lukasz comes up with will probably be fine.  :smile:

atkaiser:
Not sure if more suggestions are helpful, but I might suggest something that more describes the external affect rather than what is happening internally, such as:

`struts.matcher.setNamedParametersOnResultConfig=true`

But it the end I don't really care, they all sound fine to me.

lukaszlenart:
Right, this should express its external effect

```
struts.matcher.appendNamedParameters=true
```

If not objections I'm going to merge this PR and add the flag.

atkaiser:
Sounds good to me

lukaszlenart:
LGTM 👍

lukaszlenart:
@atkaiser @JCgH4164838Gh792C124B5 PR is ready https://github.com/apache/struts/pull/405

JCgH4164838Gh792C124B5:
Hi @atkaiser.  Thanks for discovering this issue, reporting it and providing a PR for a fix.  👍

PR #402, opened 2020-04-18
WW-5067-Fix1 (correct accidental artifactId change)

JCgH4164838Gh792C124B5:
WW-5067-Fix1
- Accidental change of artifactId when groupId was changed for weld-se in original PR.  This commit restores the originally intended artifactId.

### DIFF
--- a/plugins/cdi/pom.xml
+++ b/plugins/cdi/pom.xml
@@ -48,7 +48,7 @@
 
         <dependency>
             <groupId>org.jboss.weld.se</groupId>
-            <artifactId>weld-se-core</artifactId>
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

While looking at equivalent changes for 2.6 it was discovered the weld-se artifactId was accidentally changed (from weld-se tp weld-se-core) in the pom(s).  It did not produce any obvious failure, but it is safer to retain the originally intended artifactId (only the groupId was intended to change).

lukaszlenart:
LGTM 👍

PR #404, opened 2020-04-20
Initial fix for WW-5069 (improve build behaviour JDK9+)

JCgH4164838Gh792C124B5:
Initial fix for WW-5069:

- Force US Locale for two tests that use date strings of the form MM/DD/CCYY.
- When the tests/build are run in an environment with a non-US Locale and newer JDKs the tests may fail due to inability to parse the US date format.

Note: Was noticed on Windows 10 with JDK11, but could affect other environments when run in a non-US Locale.

### DIFF
--- a/core/src/test/java/com/opensymphony/xwork2/validator/DateRangeValidatorTest.java
+++ b/core/src/test/java/com/opensymphony/xwork2/validator/DateRangeValidatorTest.java
@@ -48,6 +48,7 @@ public class DateRangeValidatorTest extends XWorkTestCase {
         HashMap<String, Object> params = new HashMap<>();
         params.put("date", date.getTime());
         context.put(ActionContext.PARAMETERS, HttpParameters.create(params).build());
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This is a very basic PR (it was just a bit-of-a-pain to isolate the issue).

Basically, the default locale behaviour for some date validation tests differs for JDK7/8 compared to newer JDKs (such as JDK 11).  When running the tests in a non-US locale some date validation unit tests can fail, causing the Struts 2 Core build to fail.

The proposed improvement is to ensure the affected unit tests utilize the expected US locale to process the MM/DD/CCYY formats as expected.  Now they should work as intended on JDK7+, even if the base system's locale is not US.

**Note**: Windows 10 with JDK 11 (OpenJDK and Oracle) still appear to fail during a full Struts 2 build due to an apache-rat-plugin failure.  I was forced to locally use "-Drat.skip=true" to verify the full build with JDK11, but the Core build by itself passed fine).

lukaszlenart:
LGTM 👍

PR #405, opened 2020-04-22
[WW-5065] Defines a new flag to control appending params

lukaszlenart:
Follow up on https://github.com/apache/struts/pull/400
Refs [WW-5065](https://issues.apache.org/jira/browse/WW-5065)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
@@ -36,6 +36,9 @@ import java.util.*;
  * @since 2.1
  */
 public abstract class AbstractMatcher<E> implements Serializable {
+
+    private static final Logger LOG = LogManager.getLogger(AbstractMatcher.class);
+
     /**
      * <p> The logging instance </p>
      */
@@ -50,10 +53,23 @@ public abstract class AbstractMatcher<E> implements Serializable {
      * <p> The compiled patterns and their associated target objects </p>
      */
     List<Mapping<E>> compiledPatterns = new ArrayList<>();
-    ;
+
+    /**
+     * This flag controls if passed named params should be appended
+     * to the map in {@link #replaceParameters(Map, Map)}
+     * and will be accessible in {@link com.opensymphony.xwork2.config.entities.ResultConfig}.
+     * If set to false, the named parameters won't be appended.
+     *
+     * This behaviour is controlled by {@link org.apache.struts2.StrutsConstants#STRUTS_MATCHER_APPEND_NAMED_PARAMETERS}
+     *
+     * @since 2.5.23
+     * See WW-5065
+     */
+    private final boolean appendNamedParameters;
     
-    public AbstractMatcher(PatternMatcher<?> helper) {
+    public AbstractMatcher(PatternMatcher<?> helper, boolean appendNamedParameters) {
         this.wildcard = (PatternMatcher<Object>) helper;
+        this.appendNamedParameters = appendNamedParameters;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


atkaiser:
LGTM 👍

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
```diff
@@ -36,6 +36,9 @@
  * @since 2.1
  */
 public abstract class AbstractMatcher<E> implements Serializable {
+
+    private static final Logger LOG = LogManager.getLogger(AbstractMatcher.class);
```
JCgH4164838Gh792C124B5, line 40: This looks like it introduces a second private `Logger` reference ?  There is already:
```
 private static final Logger log = LogManager.getLogger(AbstractMatcher.class); 
```
Maybe the intention was to rename `log` to `LOG` and update all the logger calls ?

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
```diff
@@ -50,10 +53,23 @@
...
+     *
+     * This behaviour is controlled by {@link org.apache.struts2.StrutsConstants#STRUTS_MATCHER_APPEND_NAMED_PARAMETERS}
+     *
+     * @since 2.5.23
+     * See WW-5065
+     */
+    private final boolean appendNamedParameters;
     
-    public AbstractMatcher(PatternMatcher<?> helper) {
+    public AbstractMatcher(PatternMatcher<?> helper, boolean appendNamedParameters) {
```
JCgH4164838Gh792C124B5, line 70: Because the original method 
```
public AbstractMatcher(PatternMatcher<?> helper)
```
is public, people may have extended this class, which would then break their code unexpectedly with this change.  Maybe it would be a good idea to add keep the old method signature present with something like:
```
public AbstractMatcher(PatternMatcher<?> helper) {
    this(helper, true);  // Set boolean to whatever default setting is decided on
}
```
That should preserve backward-compatbility.

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -341,4 +341,7 @@
     public static final String STRUTS_DISALLOW_PROXY_MEMBER_ACCESS = "struts.disallowProxyMemberAccess";
 
     public static final String STRUTS_OGNL_AUTO_GROWTH_COLLECTION_LIMIT = "struts.ognl.autoGrowthCollectionLimit";
+
+    /** See {@link com.opensymphony.xwork2.config.impl.AbstractMatcher#appendNamedParameters */
+    public static final String STRUTS_MATCHER_APPEND_NAMED_PARAMETERS = "";
```
JCgH4164838Gh792C124B5, line 346: Looks like the constant is currently `""` (empty string).  Was that intentional, or should it be changed to:
```
public static final String STRUTS_MATCHER_APPEND_NAMED_PARAMETERS = "struts.matcher.appendNamedParameters";
```
which was the name Lukasz favoured in PR #400.

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
lukaszlenart: Right, done!

core/src/main/java/org/apache/struts2/StrutsConstants.java
lukaszlenart: Good catch, fixed!

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
lukaszlenart: Right, fixed :)

PR #406, opened 2020-04-25
[WW-5070] Adds more sophisticated logic to search for the Root

lukaszlenart:
Resolves [WW-5070](https://issues.apache.org/jira/browse/WW-5070)

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
@@ -27,6 +27,7 @@ import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.opensymphony.xwork2.ModelDriven;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
@@ -67,8 +68,6 @@ import com.opensymphony.xwork2.util.WildcardUtil;
  * </pre>
  */
 public class JSONResult implements Result {
-
-    private static final long serialVersionUID = 8624350183189931165L;
 
     private static final Logger LOG = LogManager.getLogger(JSONResult.class);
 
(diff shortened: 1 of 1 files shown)


plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -212,12 +211,22 @@ protected Object readRootObject(ActionInvocation invocation) {
     }
 
     protected Object findRootObject(ActionInvocation invocation) {
+        ValueStack stack = invocation.getStack();
```
JCgH4164838Gh792C124B5, line 214: The new logic seems to make sense, and the updated comment should make it clear what the behaviour will be in 2.6.

There might be some users with existing implementations depending on the current 2.5.x behaviour (always `peek()` if root is `null`), but they could implement a custom `JSONResult` with the old behaviour if they really needed it.

plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -68,8 +69,6 @@
  */
 public class JSONResult implements Result {
 
-    private static final long serialVersionUID = 8624350183189931165L;
```
JCgH4164838Gh792C124B5, line 71: Since `JSONResult` still extends `Serializable`, maybe the `serialVersionUID` should be kept ?

By removing it, anyone who Java-serialized instances in 2.5.x (possible) and then tries to Java-deserialize them in 2.6 could experience deserialization failures.

If this change is intentional because of the logic change for 2.6, maybe a brand new `serialVersionUID` value should be generated instead of just dropping it ?

lukaszlenart: Yeah, the idea was as this is a larger change to the class the ID should be changed, then I thought _why do we need the ID at all here, nobody should ever serialize a instance of `Result`_. So probably the main case is to stop extending `Serializable` by `Result`. I will put back a new ID.

plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -212,12 +211,22 @@ protected Object readRootObject(ActionInvocation invocation) {
...
             rootObject = stack.findValue(root);
         } else {
-            rootObject = invocation.getStack().peek(); // model overrides action
+            LOG.debug("Root was not defined, searching for #action");
+            rootObject = stack.findValue("#action");
+            if (rootObject instanceof ModelDriven) {
+                LOG.debug("Action is an instance of ModelDriven, assuming model is on the top of the stack and using it");
+                rootObject = stack.peek();
+            }
+            if (rootObject == null) {
```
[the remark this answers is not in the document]
lukaszlenart: Right, good point :)

plugins/json/src/main/java/org/apache/struts2/json/JSONResult.java
```diff
@@ -286,7 +294,9 @@ public String getRoot() {
     }
 
     /**
-     * Sets the root object to be serialized, defaults to the Action
+     * Sets the root object to be serialized, defaults to the Action.
+     * If the Action implements {@link ModelDriven}, model will be used instead
+     * and assumptions is the Model was pushed on the top of the stack
```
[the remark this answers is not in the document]
lukaszlenart: 👍

PR #408, opened 2020-04-26
[WW-4043] Moves TestUtils into junit-plugin

lukaszlenart:
Resolves [WW-4043](https://issues.apache.org/jira/browse/WW-4043)


plugins/junit/src/main/java/org/apache/struts2/util/TestUtils.java
```diff
@@ -85,13 +79,13 @@ public static boolean compare(URL url, String text) throws Exception {
...
-        Assert.assertEquals(bufferString,writerString);
+        Assert.assertEquals(bufferString, writerString);
     }
 
     public static String readContent(URL url) throws Exception {
         if (url == null)
             throw new Exception("unable to verify a null URL");
 
-        return IOUtils.toString(url.openStream());
+        return IOUtils.toString(url.openStream(), StandardCharsets.UTF_8);
```
JCgH4164838Gh792C124B5, line 89: It appears the previous implementation would have implicitly used `Charset.defaultCharset()` for processing the `URL`.  Some users might want (or need) a way to specify the `Charset` for processing the `URL`, so maybe a small refactoring into:

```
    public static String readContent(final URL url) throws Exception {
        return readContent(url, StandardCharsets.UTF_8);
    }

    public static String readContent(final URL url, final Charset encoding) throws Exception {
        if (url == null) {
            throw new IllegalArgumentException("Unable to verify a null URL");
        }
        if (encoding == null) {
            throw new IllegalArgumentException("Unable to verify the URL using a null Charset");
        }

        return IOUtils.toString(url.openStream(), encoding);
    }
```
might be useful ?

lukaszlenart: Done :)

lukaszlenart:
LGTM 👍

PR #410, opened 2020-04-27
[WW-5065] Defines a new flag to control appending params - cherry-pick

lukaszlenart:
Refs https://github.com/apache/struts/pull/405
Refs [WW-5065](https://issues.apache.org/jira/browse/WW-5065)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
@@ -40,11 +40,6 @@ public abstract class AbstractMatcher<E> implements Serializable {
     private static final Logger LOG = LogManager.getLogger(AbstractMatcher.class);
 
     /**
-     * <p> The logging instance </p>
-     */
-    private static final Logger log = LogManager.getLogger(AbstractMatcher.class);
-
-    /**
      * <p> Handles all wildcard pattern matching. </p>
      */
     PatternMatcher<Object> wildcard;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
```diff
@@ -107,6 +107,7 @@
 import com.opensymphony.xwork2.validator.DefaultValidatorFileParser;
 import com.opensymphony.xwork2.validator.ValidatorFactory;
 import com.opensymphony.xwork2.validator.ValidatorFileParser;
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
davelnewton, line 110: Seems spurious

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
```diff
@@ -192,7 +217,7 @@ protected String convertParam(String val, Map<String, String> vars) {
             c = val.charAt(x);
             if (x < len - 2 && 
                     c == '{' && '}' == val.charAt(x+2)) {
-                varVal = (String)vars.get(String.valueOf(val.charAt(x + 1)));
+                varVal = vars.get(String.valueOf(val.charAt(x + 1)));
```
davelnewton, line 220: (NON-BLOCKER) Never thought about it or ran in to it, but do we throw a startup message or error if it's something *other* than 0-9 or has a spurious `{42}` or `{42` or is that just considered part of a valid, if silly, path?

core/src/main/java/com/opensymphony/xwork2/config/providers/XWorkConfigurationProvider.java
lukaszlenart: 👍

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
lukaszlenart: Hm... good point, I think we can extract `OgnlTextParser` logic and reuse it here

lukaszlenart: https://issues.apache.org/jira/browse/WW-5073

PR #413, opened 2020-05-03
Proposed WW-5074 Fix

JCgH4164838Gh792C124B5:
Proposed WW-5074 Fix:
- Exclude ASM 3.3.1 from inclusion as a dependency for commons-digester3 (also mark commons-digester3 as optional dependency since it is only listed to allow the exclusion).
- Provide explicit test dependency of ASM 3.3.1 for portlet-plugin (otherwise its jmock tests are unable to execute).

### DIFF
--- a/plugins/portlet/pom.xml
+++ b/plugins/portlet/pom.xml
@@ -102,6 +102,15 @@
         </dependency>
 
         <dependency>
+            <!-- Requires ASM 3.3.1 for testing, as jmock-cglib apparently overrides a method made final in later ASM versions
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This PR attempts to fix the ASM version conflict issue (two jars with different versions, groupId changed after ASM 3.x) found after deploying the Struts Showcase applications.

This seemed to be the only combination of changes that both avoided the extra jar and resulted in a successful build.  More experienced Maven users might have a better strategy - please let me know if you do.  :smile:

PR #416, opened 2020-05-16
Partial backport of Tiny improvements PR#414 to 2.5.x:

JCgH4164838Gh792C124B5:
Partial backport of L. Lenart's PR#414 to 2.5.x:
- Improve `Dispatcher` `getLocale()` handling if running in AppEngine.
- Implemented the same improvement to the `defaultLocale` logic path within `getLocale()`.


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

Looking at #414 it seems like the Dispatcher change there would make sense for Struts 2.5.x as well.

So ... this PR is a partial backport for that change, with a small addition to the new logic @lukaszlenart applied to 2.6 for the other path in the if block.

If this change looks OK, and makes sense for 2.5.x, we can probably cherry-pick the small addition back into 2.6 as well.

Let me know what the team thinks.

PR #417, opened 2020-05-17
Provide unit test for partial backport PR#416 for 2.5.x

JCgH4164838Gh792C124B5:
Provide unit test for partial backport PR#416 for 2.5.x (from L. Lenart's PR#414) :
- DispatcherTest provides code coverage tests of all getLocale() paths.


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

The PR #416 was already accepted and merged, but Coveralls pointed out that code-coverage dropped.  Due to the potentially bad behaviour that could arise if Dispatcher Locale handling does not work as expected, a new set of unit tests for the Dispatcher getLocale() method has been created to confirm the expected behaviour.

If this PR looks OK to everyone, then it can be cherry-picked and modified as necessary to bring the tests into the 2.6.x stream.

PR #419, opened 2020-05-21
[WW-5077] Better logs

lukaszlenart:
Refs [WW-5077](https://issues.apache.org/jira/browse/WW-5077)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
@@ -25,13 +25,16 @@ import com.opensymphony.xwork2.XWorkConstants;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.security.AcceptedPatternsChecker;
 import com.opensymphony.xwork2.security.ExcludedPatternsChecker;
-import com.opensymphony.xwork2.util.*;
+import com.opensymphony.xwork2.util.ClearableValueStack;
+import com.opensymphony.xwork2.util.MemberAccessValueStack;
+import com.opensymphony.xwork2.util.ValueStack;
+import com.opensymphony.xwork2.util.ValueStackFactory;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -101,7 +104,7 @@ static private int countOGNLCharacters(String s) {
     static final Comparator<String> rbCollator = new Comparator<String>() {
         public int compare(String s1, String s2) {
             int l1 = countOGNLCharacters(s1),
-                l2 = countOGNLCharacters(s2);
+                    l2 = countOGNLCharacters(s2);
```
sepe81, line 107:
```suggestion
            int l1 = countOGNLCharacters(s1);
            int l2 = countOGNLCharacters(s2);
```

Maybe it's better (more clear) to split this into two separate assignments?

lukaszlenart: 👍

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -18,20 +18,20 @@
  */
 package com.opensymphony.xwork2.interceptor;
 
-import com.opensymphony.xwork2.ActionContext;
-import com.opensymphony.xwork2.ActionInvocation;
-import com.opensymphony.xwork2.TextProvider;
-import com.opensymphony.xwork2.XWorkConstants;
+import com.opensymphony.xwork2.*;
```
sepe81, line 21: better import explicit classes like in L25-29?

lukaszlenart: Strange... IDEA should use `*` in case of importing more than 5 classes from the same package, fixed!

lukaszlenart:
Cherry-pick is ready https://github.com/apache/struts/pull/420

PR #422, opened 2020-06-29
[WW-5080] Defines a new result type plain to use directly with Java code

lukaszlenart:
Implements [WW-5080](https://issues.apache.org/jira/browse/WW-5080)

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/PlainResult.java
+++ b/core/src/main/java/org/apache/struts2/result/PlainResult.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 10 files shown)


yasserzamani:
Shouldn't we do something in configs, allowing developer to define this result via struts.xml as well?
Already there is a PlainTextResult, shouldn't we amend it to extend this result?

lukaszlenart:
> Shouldn't we do something in configs, allowing developer to define this result via struts.xml as well?

No, this result is a plain code-based-result, you cannot define it in `struts.xml` and it can only be returned directly from an action. `DefaultActionInvocation` [supports](https://github.com/apache/struts/blob/master/core/src/main/java/com/opensymphony/xwork2/DefaultActionInvocation.java#L472-L478) returning results directly from actions for a long time.

> Already there is a PlainTextResult, shouldn't we amend it to extend this result?

No, the `PlainTextResult` result returns content of a file as a plain text. This new result allows output any type of content (text, JSON, html) but you must build it by a hand. That's why I decided to name it a `PlainResult` and with Java 8 you can use lambdas to construct the output:

```java
    public PlainResult execute() {
        return response -> response.write("ok");
    }
```

yasserzamani:
LGTM thanks!

PR #425, opened 2020-07-11
[WW-5077] DMI aware pattern

lukaszlenart:
Refs [WW-5077](https://issues.apache.org/jira/browse/WW-5077)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/security/DefaultAcceptedPatternsChecker.java
+++ b/core/src/main/java/com/opensymphony/xwork2/security/DefaultAcceptedPatternsChecker.java
@@ -21,8 +21,10 @@ package com.opensymphony.xwork2.security;
 import com.opensymphony.xwork2.XWorkConstants;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.util.TextParseUtil;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.struts2.StrutsConstants;
 
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/security/DefaultAcceptedPatternsChecker.java
```diff
@@ -35,7 +37,11 @@
     private static final Logger LOG = LogManager.getLogger(DefaultAcceptedPatternsChecker.class);
 
     public static final String[] ACCEPTED_PATTERNS = {
-            "\\w+((\\.\\w+)|(\\[\\d+\\])|(\\(\\d+\\))|(\\['(\\w|[\\u4e00-\\u9fa5])+'\\])|(\\('(\\w|[\\u4e00-\\u9fa5])+'\\)))*"
+            "\\w+((\\.\\w+)|(\\[\\d+])|(\\(\\d+\\))|(\\['(\\w|[\\u4e00-\\u9fa5])+'])|(\\('(\\w|[\\u4e00-\\u9fa5])+'\\)))*"
```
yasserzamani, line 40: Looks strange for me. Closing escaped [ brackets with not-escaped ] brackets?

lukaszlenart: It was suggested by IDEA, restored the previous version :)

yasserzamani: Ah sorry [it was right](https://stackoverflow.com/questions/43044918/why-closing-square-bracket-doesnt-require-escaping-in-java-regex).

And I spotted it's nice I think to refactor `((\\.\\w+)|(\\[\\d+\\])|(\\(\\d+\\))|(\\['(\\w|[\\u4e00-\\u9fa5])+'\\])|(\\('(\\w|[\\u4e00-\\u9fa5])+'\\)))*` part e.g. named `EXTENSION_PATTERN` or so because it's same in both.

lukaszlenart: Yes, but I would like to keep them as is, it's easier to understand what's going on. Also future changes to the patterns want affect each other.

PR #426, opened 2020-07-14
WW-5083: Adds support for Fetch Metadata in Struts2.

salcho:
Hello Struts devs,

This PR builds Fetch Metadata support for Struts2, namely:

- If a request has `Sec-Fetch-*` headers (i.e. comes from a modern browser), the Fetch Metadata Interceptor will reject the request if it is requested cross-site (a potential CSRF attack).
- One default Resource Isolation Policy is provided based on https://web.dev/fetch-metadata/, which prevents all major cross-site request forgery attacks.
- This Interceptor gives the ability to add exemptions to this security mitigation, that is: URLs that are meant to be accessed cross-site.
- The Fetch Metadata Interceptor has been added to the default interceptor stack.
- The `Vary` header has been added to responses to ensure that any cached responses include Fetch Metadata headers in their key. This is an added layer of security against cache poisoning.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
@@ -1,0 +1,49 @@
+package org.apache.struts2.interceptor;
+
+import org.apache.logging.log4j.util.Strings;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ *
+ * Default resource isolation policy used in {@link FetchMetadataInterceptor} that
+ * implements the {@link ResourceIsolationPolicy} interface. This default policy is based on
+ * <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>.
+ *
+ * @see <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>
+ *
+ * @author Santiago Diaz - saldiaz@google.com
+ * @author Giannis Chatziveroglou - giannic@google.com
+ **/
+
+public final class DefaultResourceIsolationPolicy implements ResourceIsolationPolicy {
+
+    @Override
+    public boolean isRequestAllowed(HttpServletRequest request) {
+        String site = request.getHeader(SEC_FETCH_SITE_HEADER);
+
+        // Allow requests from browsers which don't send Fetch Metadata
+        if (Strings.isEmpty((site))){
+            return true;
+        }
+
+        // Allow same-site and browser-initiated requests
+        if (SAME_ORIGIN.equals(site) || SAME_SITE.equals(site) || NONE.equals(site)) {
+            return true;
+        }
+
+        // Allow simple top-level navigations except <object> and <embed>
+        return isAllowedTopLevelNavigation(request);
+    }
+
+    private boolean isAllowedTopLevelNavigation(HttpServletRequest request)
+    {
+        String mode = request.getHeader(SEC_FETCH_MODE_HEADER);
+        String dest = request.getHeader(SEC_FETCH_DEST_HEADER);
+
+        boolean isSimpleTopLevelNavigation = MODE_NAVIGATE.equals(mode) || "GET".equals(request.getMethod());
+        boolean isNotObjectOrEmbedRequest = !DEST_EMBED.equals(dest) && !DEST_OBJECT.equals(dest);
+
+        return isSimpleTopLevelNavigation && isNotObjectOrEmbedRequest;
+    }
+}
--- a/core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.apache.struts2.interceptor;
 
 import org.apache.logging.log4j.util.Strings;
--- a/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
@@ -43,7 +43,7 @@ import org.apache.logging.log4j.Logger;
  * @see <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>
  **/
 
-public class FetchMetadataInterceptor extends AbstractInterceptor implements PreResultListener {
+public class FetchMetadataInterceptor extends AbstractInterceptor {
     private static final Logger logger = LogManager.getLogger(FetchMetadataInterceptor.class);
     private static final String VARY_HEADER_VALUE = String.format("%s,%s,%s", SEC_FETCH_DEST_HEADER, SEC_FETCH_SITE_HEADER, SEC_FETCH_MODE_HEADER);
     private static final String SC_FORBIDDEN = String.valueOf(HttpServletResponse.SC_FORBIDDEN);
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
```diff
@@ -0,0 +1,49 @@
+package org.apache.struts2.interceptor;
```
lukaszlenart, line 19: You are missing a header with license, see other files for an example

core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ *
+ * Default resource isolation policy used in {@link FetchMetadataInterceptor} that
+ * implements the {@link ResourceIsolationPolicy} interface. This default policy is based on
+ * <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>.
+ *
+ * @see <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>
+ *
+ * @author Santiago Diaz - saldiaz@google.com
+ * @author Giannis Chatziveroglou - giannic@google.com
```
lukaszlenart, line 34: Please remove these tags and emails, after merging this code it will be owned by ASF and you want to be bothered by users to help them resolve their problems, it's community duty :)

http://tinyurl.com/mw7t6

core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
```diff
@@ -0,0 +1,49 @@
...
+ **/
+
+public final class DefaultResourceIsolationPolicy implements ResourceIsolationPolicy {
```
lukaszlenart, line 37: I prefer to use `Struts` prefix to indicate that this a default `Struts` implementation :)

lukaszlenart, line 37:
```suggestion
public final class StrutsResourceIsolationPolicy implements ResourceIsolationPolicy {
```

core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
```diff
@@ -0,0 +1,49 @@
...
+        if (SAME_ORIGIN.equals(site) || SAME_SITE.equals(site) || NONE.equals(site)) {
+            return true;
+        }
+
+        // Allow simple top-level navigations except <object> and <embed>
+        return isAllowedTopLevelNavigation(request);
+    }
+
+    private boolean isAllowedTopLevelNavigation(HttpServletRequest request)
+    {
```
lukaszlenart, line 58: Inconsistent formatting, please move this bracket to the end of the previous line, thanks!

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -0,0 +1,96 @@
+package org.apache.struts2.interceptor;
```
lukaszlenart, line 19: Header with a license is missing

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+import com.opensymphony.xwork2.interceptor.PreResultListener;
+import com.opensymphony.xwork2.util.TextParseUtil;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * <!-- START SNIPPET: description -->
```
lukaszlenart, line 33: It's  not needed anymore, we don't use Confluence any longer, instead you can create a PR with description about the new interceptor [here](https://github.com/apache/struts-site/blob/master/source/core-developers/interceptors.md)

core/src/main/java/org/apache/struts2/interceptor/ResourceIsolationPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+ * Resource isolation policies are designed to protect against cross origin attacks and use the
+ * {@code sec-fetch-*} request headers to decide whether to accept or reject a request. Read more
+ * about <a href="https://web.dev/fetch-metadata/">Fetch Metadata.</a>
+ *
+ * See {@link DefaultResourceIsolationPolicy} for the default implementation used.
+ *
+ * @see <a href="https://web.dev/fetch-metadata/">https://web.dev/fetch-metadata/</a>
+ *
+ * @author Santiago Diaz - saldiaz@google.com
+ * @author Giannis Chatziveroglou - giannic@google.com
```
lukaszlenart, line 35: Please remove these tags as explained earlier

core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
salcho: just pushed another commit adding the missing licenses

core/src/main/java/org/apache/struts2/interceptor/DefaultResourceIsolationPolicy.java
salcho: haha that's fair! :)

aaronshim:
Great job, @salcho !

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+        if (exemptedPaths.contains(request.getContextPath())) {
+            return invocation.invoke();
+        }
+
+        // Check if request is allowed
+        if (resourceIsolationPolicy.isRequestAllowed(request)) {
+            return invocation.invoke();
+        }
+
+        beforeResult(invocation, SC_FORBIDDEN);
```
yasserzamani, line 81: Doesn't work without this line? -- it's already registered as a pre-result.

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+        return SC_FORBIDDEN;
```
yasserzamani, line 82: Maybe an info level log before this line will help why request is forbidden.

core/src/main/java/org/apache/struts2/interceptor/StrutsResourceIsolationPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+ **/
+
+public final class StrutsResourceIsolationPolicy implements ResourceIsolationPolicy {
+
+    @Override
+    public boolean isRequestAllowed(HttpServletRequest request) {
+        String site = request.getHeader(SEC_FETCH_SITE_HEADER);
+
+        // Allow requests from browsers which don't send Fetch Metadata
+        if (Strings.isEmpty((site))){
```
yasserzamani, line 41: nitpick: waste parenteses. Additionally you might use `org.apache.commons.lang3.StringUtils.isEmpty/isBlank` instead.

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
salcho: Unfortunately it doesn't, I was surprised too! I believe this is because the interceptor chain isn't propagated due to returning early?

salcho:
Thanks for your review Yasser! The log statement will be useful for debugging endpoints that should have been added to exempted paths.

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
yasserzamani: Yes likely. I'll take a look if it can be fixed. thanks!

lukaszlenart: Oh, good point. You must register your `PreResultListener` first instead of calling it by a hand. In such case you will execute your logic related to result as there are still some other interceptors in the execution stack.

Just do `invocation.addPreResultListener(this)`, see example in [DebuggingInterceptor](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/interceptor/debugging/DebuggingInterceptor.java#L156-L161)

salcho:
Hi @lukaszlenart , sorry, I couldn't reply to your comment so here's my reply:

I think this case is slightly different than usual in that there are two outcomes for Fetch Metadata: we **always** register ourselves as a `PreResultListener` (line 69) and then we have two possible outcomes: either the request is accepted as legitimate (in which case we let the interceptor chain continue and eventually our `beforeResult` method called) or we reject the request and by doing so interrupt the chain of interceptors. In the latter, the `beforeResult` callback is never called, but because we still want to add the `Vary` headers, we call it manually to reuse that logic.

If this code is somewhat harder to read, we could extract the logic for adding the `Vary` headers into a separate method and call that method explicitly in the rejection branch, so that it doesn't look like we're trying to do something related to the responsibilities of `PreResultListener`.

What do you think?

lukaszlenart:
Yeah... I think it would be better to stop using `PreResultListener` and just add the header on the beginning of the `intercept()` method.

lukaszlenart:
🎉  great 🎉 

LGTM 👍

PR #427, opened 2020-07-20
WW-5075 Fixes for OSGi plugin in 2.5.x

JCgH4164838Gh792C124B5:
Initial attempt to fix OSGi Plugin issues discovered during examination of WW-5075.
- Indication is that the OSGi Plugin has been failing since Struts 2.3.4.
- Applied fix to ResourceFinder to restore 2.3.x logic that was broken in 2.5.x.
- Implemented changes to restore OSGi Plugin functionality for 2.5.x.
- Applied changes to the OSGi Admin Bundle to function under 2.5.x.
- Updated the OSGi Admin Bundle JS libraries and related CSS to newer versions.
- Added new unit tests for the OSGi plugin.
- The OSGi plugin will only function properly with exploded/expanded WAR files.


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This activity took a lot longer than originally planned, but finally it is at a state where a PR for 2.5.x seemed reasonable.  Originally the plan was to focus only on a 2.6.x revision, but due to all the issues found along the way (in 2.3.x, and 2.5.x), I decided it was better to stabilize the OSGi plugin for 2.5.x first.

If after a review this PR is accepted for 2.5.x, I will proceed to put together a separate PR for 2.6.x (probably not a cherry-pick, but we will see).

The Demo Bundle remains non-functional even after this PR (in part due to Spring dropping its OSGi integration a while back and the Demo Bundle depends on it).  A 2.6.x version will probably drop the Demo Bundle.

JCgH4164838Gh792C124B5:
Hmmm.  The OSGi plugin builds are all failing for the same reason, unit test failures because Felix cannot create its default cache directory ("/tmp.felix-cache").

I had no issues with the unit tests failing on my local builds (using various JDKs), so it could be a Linux/Windows difference, or due to permissions in the CI environment.  Not sure if there is a workaround for either the test configuration or the CI environment for this condition.

plugins/osgi/src/main/java/org/apache/struts2/osgi/BundlePackageLoader.java
```diff
@@ -45,24 +47,96 @@
...
+        }
+
+        defaultFileManagerFactory.setContainer(container);
+
+        if (fileManagerFactory == null || fileManagerFactory.getFileManager() == null) {
+             LOG.warn("LoadPackages - FileManagerFactory parameter is null or produces a null FileManager, replacing with a new DefaultFileManagerFactory instance");
+             fileManagerFactory = defaultFileManagerFactory;
+        }
+
         prov.setFileManagerFactory(fileManagerFactory);
```
lukaszlenart, line 89: I would change this a bit because right now you are re-assigning to the method's parameter, also you are creating `DefaultFileManagerFactory` upfront even if it could be not needed. What about such change:

```java
if (fileManagerFactory == null || fileManagerFactory.getFileManager() == null) {
  LOG.warn("LoadPackages - FileManagerFactory parameter is null or produces a null FileManager, replacing with a new DefaultFileManagerFactory instance");

  final DefaultFileManagerFactory defaultFileManagerFactory = new DefaultFileManagerFactory();
  final Container container = getContextContainer();

  if (container == null) {
    LOG.warn("LoadPackages - Config Container is null.  May cause a NPE to be thrown");
  } else {
    container.inject(defaultFileManagerFactory);
  }

  prov.setFileManagerFactory(defaultFileManagerFactory);
} else {
  prov.setFileManagerFactory(fileManagerFactory);
}
```

plugins/osgi/src/main/java/org/apache/struts2/osgi/BundlePackageLoader.java
```diff
@@ -45,24 +47,96 @@
 
     private static final Logger LOG = LogManager.getLogger(BundlePackageLoader.class);
 
+    private Container contextContainer = null;
+
+    @Override
     public List<PackageConfig> loadPackages(Bundle bundle, BundleContext bundleContext, ObjectFactory objectFactory,
```
lukaszlenart, line 53: Maybe mark this `@Deprecated`?

yasserzamani:
Wow, thanks a lot!

Regarding felix cache permission issue, it seems you or user can use [org.osgi.framework.storage](https://felix.apache.org/documentation/subprojects/apache-felix-framework/apache-felix-framework-bundle-cache.html) configuration property e.g. `System.getProperty("java.io.tmpdir")`?

plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiUtil.java
```diff
@@ -92,4 +92,83 @@ public static boolean containsBean(Object beanFactory, String beanId) {
 
         return false;
     }
+
```
yasserzamani, line 95: What about using ready to use java version resolvers like apache commons or etc via googling?

plugins/osgi/src/main/java/org/apache/struts2/osgi/BundlePackageLoader.java
JCgH4164838Gh792C124B5: I cannot remember why I re-used the `fileManagerFactory` parameter (probably trying to limit the number of lines that I was changing at the time).  Anyway, your recommended changes above make more sense, so we will go with that logic instead (in an updated commit).

plugins/osgi/src/main/java/org/apache/struts2/osgi/BundlePackageLoader.java
JCgH4164838Gh792C124B5: Good point, it will be updated as well.

plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiUtil.java
JCgH4164838Gh792C124B5: Hi.  Thanks for the comments above, it helped point out an issue with the cache directory creation logic.  :)

In terms of using pre-existing Java version resolvers, I decided to "stay custom" within the plugin.  The main reasoning behind that was:
  1) To limit additional dependencies for the plugin at runtime (sometimes makes it more difficult to figure out why the OSGi container cannot start or load bundles), and 
  2) During testing I found that we need control over how the version was discovered, in order to be consistent with the OSGi container (with external java version resolvers, they may return a different version value than the one we need for the lookups to succeed).

So, for now at least, it seems safer to stick with the custom local Java version resolvers.

JCgH4164838Gh792C124B5:
Hi.  It looks like the latest commit fixed the test-failure issue involving the cache directory, and the builds for all four JDKs completed.

Let me know if there are any additional changes that should be made.  :)

lukaszlenart:
LGTM 👍

plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiUtil.java
yasserzamani: :+1:

PR #428, opened 2020-07-26
WW-5083 PR#426 follow-up.

JCgH4164838Gh792C124B5:
WW-5083 PR#426 follow-up.
- Updated ResourceIsolationPolicy Sec-Fetch* header cases to match spec.
- Added the Sec-Fetch-User header, plus additional dest/site/mode values  from the spec.
- Renamed ResourceIsolationPolicy interface constants to follow the naming convention that was already present.
- Made StrutsResourceIsolationPolicy checks case-insensitive (even if specification says things should be case-sensitive) to better handle client bugs that will likely occur in the future.
- Updated FetchMetaDataInterceport to use more standard LOG reference name, parameterization and call forms seen in other Struts 2 Interceptors.
- Including the Sec-Fetch-User in the Vary resonse header.
- Make setExemptedPaths an injectable method (but not required).
- Updated unit test to use more of the constants, added test confirming the Vary header replacement.
- A few whitespace changes and JavaDoc additions, including reference to the W3C specification site.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
@@ -76,7 +76,7 @@ public class FetchMetadataInterceptor extends AbstractInterceptor {
             return invocation.invoke();
         }
 
-        LOG.debug("Fetch metadata rejected cross-origin request to [{}]", contextPath);
+        LOG.info("Fetch metadata rejected cross-origin request to [{}]", contextPath);
         return SC_FORBIDDEN;
     }
 
--- a/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
@@ -21,12 +21,13 @@ package org.apache.struts2.interceptor;
 import static org.apache.struts2.interceptor.ResourceIsolationPolicy.SEC_FETCH_DEST_HEADER;
 import static org.apache.struts2.interceptor.ResourceIsolationPolicy.SEC_FETCH_MODE_HEADER;
 import static org.apache.struts2.interceptor.ResourceIsolationPolicy.SEC_FETCH_SITE_HEADER;
+import static org.apache.struts2.interceptor.ResourceIsolationPolicy.SEC_FETCH_USER_HEADER;
 import static org.apache.struts2.interceptor.ResourceIsolationPolicy.VARY_HEADER;
 
 import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.ActionInvocation;
+import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.interceptor.AbstractInterceptor;
-import com.opensymphony.xwork2.interceptor.PreResultListener;
 import com.opensymphony.xwork2.util.TextParseUtil;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


JCgH4164838Gh792C124B5:
Hello Struts Developers.

This PR is intended to be some minor cleanup to the new `FetchMetadataInterceptor`, recently added to the 2.6.x series via PR #426.

Hello @salcho.  Could you please take a look at these changes and let me know if they seem OK to you ?  Good work on the original PR, by the way :smile:).

Regards,
James.

salcho:
@JCgH4164838Gh792C124B5 

Hi James,

Thanks so much for the follow up! We're getting ready to submit a second PR soon and going through your changes is very useful. We discussed whether string comparisons should've been case insensitive and voted against it but being defensive, as this PR is, really can't do any harm. 

All these look good to me, in particular the fact that there's now an implicit empty set of exemptions in the interceptor, which I think is reasonable 👍 

Have a great start of the week! 
:)

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -41,17 +42,19 @@
...
-    private static final String VARY_HEADER_VALUE = String.format("%s,%s,%s", SEC_FETCH_DEST_HEADER, SEC_FETCH_SITE_HEADER, SEC_FETCH_MODE_HEADER);
+    private static final Logger LOG = LogManager.getLogger(FetchMetadataInterceptor.class);
+    private static final String VARY_HEADER_VALUE = String.format("%s,%s,%s,%s", SEC_FETCH_DEST_HEADER, SEC_FETCH_MODE_HEADER, SEC_FETCH_SITE_HEADER, SEC_FETCH_USER_HEADER);
     private static final String SC_FORBIDDEN = String.valueOf(HttpServletResponse.SC_FORBIDDEN);
 
     private final Set<String> exemptedPaths = new HashSet<>();
     private final ResourceIsolationPolicy resourceIsolationPolicy = new StrutsResourceIsolationPolicy();
 
-    public void setExemptedPaths(String paths){
+    @Inject (required=false)
```
yasserzamani, line 56: will be nice to have a test for injection

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
```diff
@@ -73,15 +76,31 @@ public String intercept(ActionInvocation invocation) throws Exception {
             return invocation.invoke();
         }
 
-        logger.atDebug().log(
-            "Fetch metadata rejected cross-origin request to %s",
-            contextPath
-        );
+        LOG.debug("Fetch metadata rejected cross-origin request to [{}]", contextPath);
```
yasserzamani, line 79: To improve, I think `.warn` is better.

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
JCgH4164838Gh792C124B5: Hi.  Sounds reasonable.  I will try to add one soon.

core/src/main/java/org/apache/struts2/interceptor/FetchMetadataInterceptor.java
JCgH4164838Gh792C124B5: Hi.  It looks like this interceptor could produce a lot of logging output, to the point that it might flood the application's logs.
Maybe the debug-level output is OK, since the developer can usually find the related forbidden (403) response in the access logs of the application server anyway ?
Another option could be to use a `.info` level for the message instead (could still be suppressed if developers set minimum log level to warning, but easier to differentiate than debug-level).  What do you think ?

salcho: Hi,

I think you guys know better what log level is more appropriate for Struts devs. What I can tell you is that this code is not expected to reject many requests under normal circumstances, as every time this happens an endpoint is being fetched cross-origin, which indicates a potential CSRF attack more often than not and developers are expected to add endpoints that are exposed cross-origin to the exemption list. 

That is to say: we wouldn't really expect this log statement to be hit to the point of flooding the logs.

JCgH4164838Gh792C124B5:
Hello Again.  This PR has been updated with the suggested additional unit test and increased the level of the "Fetch metadata rejected" log message to `info` level.  Let me know if anyone thinks additional changes are needed.

yasserzamani:
osm! thanks!

PR #429, opened 2020-07-28
WW-5084: Add Content Security Policy support to Struts

salcho:
Hello Struts Devs!

This PR adds Content Security Policy support for Struts 2. A very popular security mitigation against XSS and other injection vulnerabilities. CSP comes in many flavours, but we've chosen to add support for the most robust of them: nonce-based, strict-dynamic CSP.

Here's a summary of these changes:

- Allows users to configure whether CSP is enabled in reporting or enforcement modes and lets them set a report URI, where violation reports will be sent by the browser.
- Implements a CSP Interceptor that generates a nonce-based, strict-dynamic policy and adds it to HTTP responses according to the user's configuration.
- Implements custom JSP and FTL <script> tags. These (`<s:script>` in taglib, for instance) set the nonce attribute on script blocks automatically, so that they match the nonce set in the policy. This feature allows developers to use both existing and new script blocks that are compatible with CSP with minimal refactoring.
- Provides a default implementation of a CSP violation report collection endpoint. This allows developers to see CSP reports as they happen in their logs out of the box, with minimal effort. This behaviour is extensible, so developers can customise the processing of CSP reports.

With these tools, developers can enable CSP in reporting mode, collect reports and identify and refactor existing code that is incompatible with CSP from these reports. Finally, developers will be able to switch CSP to enforcing mode, which will provide a very robust defense against XSS. 

Co-authored-by: Ecenaz Jen Ozmen <eo2419@columbia.edu>
Co-authored-by: Giannis Chatziveroglou <giannisc@google.com>
Co-authored-by: Sal <salchoman@gmail.com>


salcho:
Sorry, this PR was based off the wrong branch in our fork and has been replaced by https://github.com/apache/struts/pull/430

PR #430, opened 2020-07-28
WW-5084: Add Content Security Policy support to Struts

salcho:
Hello Struts Devs!

This PR adds Content Security Policy support for Struts 2. A very popular security mitigation against XSS and other injection vulnerabilities. CSP comes in many flavours, but we've chosen to add support for the most robust of them: nonce-based, strict-dynamic CSP.

Here's a summary of these changes:

- Allows users to configure whether CSP is enabled in **reporting or enforcement modes** and lets them set a report URI, where violation reports will be sent by the browser.
- Implements a CSP Interceptor that generates a **nonce-based, strict-dynamic policy** and adds it to HTTP responses according to the user's configuration.
- Implements custom `<script>` and `<link>` JSP and FTL tags. These (<s:script> in taglib, for instance) **set the nonce attribute on script and link blocks automatically**, so that they match the nonce set in the policy. This feature allows developers to use both existing and new script blocks that are compatible with CSP with minimal refactoring.
- Provides a default implementation of a CSP violation report collection endpoint. This **allows developers to see CSP reports as they happen in their logs out of the box**, with minimal effort. This behaviour is extensible, so developers can customise the processing of CSP reports.
- With these tools, developers can enable CSP in reporting mode, collect reports and identify and refactor existing code that is incompatible with CSP from these reports. Finally, developers will be able to switch CSP to enforcing mode, which will provide a very robust defense against XSS.

Co-authored-by: Ecenaz Jen Ozmen - eo2419@columbia.edu
Co-authored-by: Giannis Chatziveroglou - giannisc@google.com
Co-authored-by: Santiago Diaz - salchoman@gmail.com

### DIFF
--- a/core/src/main/java/org/apache/struts2/action/CspReportAction.java
+++ b/core/src/main/java/org/apache/struts2/action/CspReportAction.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.action;
+
+import com.opensymphony.xwork2.ActionSupport;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.BufferedReader;
+import java.io.IOException;
+
+import static org.apache.struts2.interceptor.csp.CspSettings.CSP_REPORT_TYPE;
+
+/**
+ * An abstract Action that can be extended to process the incoming CSP violation reports. Performs
+ * necessary checks to extract the JSON string of the CSP report and make sure it's a valid report.
+ * Always returns a 204 response.
+ *
+ * Override the <code>processReport(String jsonCspReport)</code> method to customize how the action processes
+ * the CSP report. See {@link DefaultCspReportAction} for the default implementation.
+ *
+ * Add the action to the endpoint that is the <code>reportUri</code> in the {@link org.apache.struts2.interceptor.csp.CspInterceptor}
+ * to collect the reports.
+ *
+ * <pre>
+ *     &lt;package name="csp-reports" namespace="/" extends="struts-default"&gt;
+ *         &lt;action name="csp-reports" class="org.apache.struts2.action.DefaultCspReportAction"&gt;
+ *             &lt;result type="httpheader"&gt;
+ *                 &lt;param name="statusCode">200&lt;/param&gt;
+ *             &lt;/result&gt;
+ *         &lt;/action&gt;
+ *     &lt;/package&gt;
+ * </pre>
+ *
+ * @see DefaultCspReportAction
+ */
+public abstract class CspReportAction extends ActionSupport implements ServletRequestAware, ServletResponseAware {
+    private HttpServletRequest request;
+
+    public String execute() throws IOException {
+        return "success";
+    }
+
+    @Override
+    public void withServletRequest(HttpServletRequest request) {
+        if (!isCspReportRequest(request)) {
+            return;
+        }
+
+        try {
+            BufferedReader reader = request.getReader();
+            String cspReport = reader.readLine();
+            processReport(cspReport);
+        } catch (IOException ignored) {
+        }
+    }
+
+    private boolean isCspReportRequest(HttpServletRequest request) {
+        if (!"POST".equals(request.getMethod()) || request.getContentLength() <= 0){
+            return false;
+        }
+
+        String contentType = request.getContentType();
+        return CSP_REPORT_TYPE.equals(contentType);
+    }
+
+    @Override
+    public void withServletResponse(HttpServletResponse response) {
+        response.setStatus(204);
+    }
+
+    abstract void processReport(String jsonCspReport);
+
+    public void setServletRequest(HttpServletRequest request) {
+        this.request = request;
+    }
+
+    public HttpServletRequest getServletRequest() {
+        return request;
+    }
+}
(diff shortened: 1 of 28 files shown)


salcho:
That's great news @lukaszlenart! Is there anything else we can do to get these merged? :)

lukaszlenart:
I'm just giving some time to others to allow them review your work :)

aleksandr-m:
@salcho @lukaszlenart Do we need `CspReportAction` in the Struts itself? Seems like it is too implementation specific.

core/src/main/java/org/apache/struts2/action/CspReportAction.java
```diff
@@ -0,0 +1,97 @@
...
+ *         &lt;/action&gt;
+ *     &lt;/package&gt;
+ * </pre>
+ *
+ * @see DefaultCspReportAction
+ */
+public abstract class CspReportAction extends ActionSupport implements ServletRequestAware, ServletResponseAware {
+    private HttpServletRequest request;
+
+    public String execute() throws IOException {
```
aleksandr-m, line 56: Add `@Override` and use `SUCCESS` constant.

aleksandr-m: Or remove this method.

salcho:
Hi @aleksandr-m @lukaszlenart,

Re: **CspReportAction**. In our experience, many developers are put off of adopting CSP in their applications because collecting reports adds more overhead. We decided to add the `CspReportAction` class with a default implementation to give those developers the option to either use an out of the box (and useful, since the reports can be seen in logs) implementation or subclass the action and implement custom logic.

Alex, I've pushed a new commit that addresses your comment. Thanks for flagging!

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        if (!uri.get().isAbsolute() && !reportUri.startsWith("/")) {
+            throw new IllegalArgumentException("Illegal configuration: report URI is not relative to the root. Please set a report URI that starts with /");
+        }
+
+        settings.setReportUri(reportUri);
+    }
+
+    private static Optional<URI> buildUri(String reportUri) {
```
aleksandr-m, line 65: Why static?

core/src/main/resources/struts-default.xml
```diff
@@ -377,6 +378,10 @@
                 <interceptor-ref name="alias"/>
                 <interceptor-ref name="servletConfig"/>
                 <interceptor-ref name="i18n"/>
+                <interceptor-ref name="cspInterceptor">
+                    <param name="enforcingMode">false</param>
+                    <param name="reportUri">/csp-reports</param>
```
aleksandr-m, line 383: What will happen if `reportUri` is set and no action is configured? Maybe we should remove `reportUri` from default xml?

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
salcho: no particular reason, really - just because it has no dependencies with the instance and it's a utility method. I don't have a strong opinion on this, would you like to have this method be non-static?

core/src/main/resources/struts-default.xml
salcho: A user can still configure the `reportUri` value to point to an external host! So even if devs choose not to use/subclass the action that Struts provides, they can still use CSP's reporting features (see services like https://report-uri.io and suchlike). I would say leaving it here gives it visibility and perhaps we can make sure that documentation is clear on this point? 

WDYT?

aleksandr-m: This is the default stack, meaning that report uri will be set for all applications using it. If there is no `/csp-reports` action then what will happen? 404?

Also (slim chances but still) what if application already has action with that exact name, but for some other stuff?

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
aleksandr-m: I would say that if there is no reason why it should be static then it shouldn't be. :)

core/src/main/java/org/apache/struts2/action/DefaultCspReportAction.java
```diff
@@ -0,0 +1,38 @@
...
+ * that contains the details of the CSP violation.
+ *
+ * @see CspReportAction
+ */
+public class DefaultCspReportAction extends CspReportAction {
+
+    protected static final Logger LOG = LogManager.getLogger(DefaultCspReportAction.class);
+
+    @Override
+    void processReport(String jsonCspReport) {
```
aleksandr-m, line 35: I still don't think we should maintain this action in framework core, but I fail to see how `processReport` will be called in the current implementation.

core/src/main/resources/struts-default.xml
salcho: That's a fair point! I've pushed a new commit with these changes so the new behaviour is:

- By default, CSP will be set to report-mode (so all websites continue working as previously) and violations will only show up on the console.
- If a user decides to add a reportUri, violations will be seen on the console and will be sent to the URI.
- A user may choose to turn enforcement mode on and the above two points still hold.

I think this is a less intrusive default config!

core/src/main/java/org/apache/struts2/action/DefaultCspReportAction.java
salcho: The `CspReportAction` ensures that reports sent to the reporting endpoint are correct (e.g. use the right HTTP method, have the right Content-Type, etc.) and that responses to them are correct too (they always return a 204). 

After `CspReportAction` receives a valid request extracts its JSON body, it will call `processReport` so that that report gets processed _somehow_ (i.e. dump to disk, extract a few keys from it and log it, de-duplicate it with other reports, log it, etc.). 

I believe there is value in keeping `CspReportAction`, here's what I'd propose: let's turn `CspReportAction` into a concrete class (not abstract anymore) and remove `DefaultCspReportAction` (which only logs the report anyway). `CspReportAction` would just log the report by default (at DEBUG, perhaps?) and we could let devs override this class if they wanted to customize the processing of the report.

Does this sound reasonable?

salcho:
Friendly ping! @aleksandr-m @lukaszlenart

lukaszlenart:
I'm fine with the current state, LGTM 👍  (but I will let @aleksandr-m hit the button  :)

TheFergus:
@lukaszlenart Hi, for <s: script><s:/script>, if the src of the tags needs variables, what should I do?
<s:script src="${myUrl}"></s:script>

lukaszlenart:
@TheFergus already answered here https://github.com/struts-community-plugins/struts2-jquery/pull/318#issuecomment-1333471447

PR #432, opened 2020-08-11
WW-5085: Add Cross-Origin Opener Policy (COOP) and Cross-Origin Embedder Policy (COEP) support

gchatz22:
Hello Struts Devs!

This PR adds Cross-Origin Opener Policy (COOP) and Cross-Origin Embedder Policy (COEP) support for Struts2. Two very popular mitigations that aim to make a website cross-origin isolated.

COOP is a security mitigation that lets developers isolate their resources against side-channel attacks and information leaks. COEP prevents a document from loading any non-same-origin resources which don't explicitly grant the document permission to be loaded. Both COOP and COEP require adding headers to the response object.

COOP and COEP are independent mechanisms and they can be enabled, tested and deployed separately. Using COEP and COOP together allows developers to safely use powerful features such as `SharedArrayBuffer`, `performance.measureMemory()`, and the JS Self-Profiling API. COOP and COEP are now supported by all major browsers. See https://web.dev/why-coop-coep/ for reference.

Here's a summary of the changes made:

- We created 2 new interceptors, the `CoopInterceptor` and `CoepInterceptor`, that handle adding the response headers for the respective security mitigation. Both were added in the default stack of `struts-default.xml`
- Using the `CoopInterceptor`, developers have the ability to choose the policy they want COOP to operate with (`same-origin`, `same-origin-allow-popups`, or `unsafe-none`), set to `same-origin` by default. An exception is thrown if another policy other than the specified 3 is chosen
- Using the `CoepInterceptor`, developers have the ability to choose between the two modes they want COEP to operate under: enforcing mode (header set as `Cross-Origin-Embedder-Policy`) which blocks resources and reports violation, and reporting mode (header set as `Cross-Origin-Embedder-Policy-Report-Only`) which only reports the violation without blocking resources.
- For both interceptors, developers are able to add exempted paths for which the chosen policy will not be applied to

Here is how developers can customize the interceptors in `struts.xml` :
```
<interceptor-ref name="coopInterceptor">
        <param name="exemptedPaths">path1,path2</param>
        <param name="mode">same-origin</param>
</interceptor-ref>
<interceptor-ref name="coepInterceptor">
         <param name="enforcingMode">false</param>
         <param name="disabled">false</param>
         <param name="exemptedPaths">path1,path2</param>
</interceptor-ref>
```


yasserzamani:
LGTM :+1: thanks a lot!

lukaszlenart:
👏  LGTM 👍

PR #434, opened 2020-08-31
Refactoring Nonce Propagation in Struts Tags for CSP

eozmen410:
Hello Struts devs!

This PR is a follow up to [WW-5084: Add Content Security Policy support to Struts](https://github.com/apache/struts/pull/430) to make all Struts tags CSP ready. After our inital CSP implementation we realized that other Struts tags like `<s:doubleselect>`, `<s:head>` also include `<script>` or `<link/>` blocks, and we wanted to make sure enabling CSP will not compromise any of the functionality for the existing tags! Here's a summary of the changes we made:

* Modify the `UIBean` class to add the nonce value as a parameter so tags that need the nonce value can access it
* Add `nonce.ftl` and `<include />` it for tags that need the nonce attribute 
*  Modify the showcase JSP files to use `<s:script>` and `<s:link/>` instead of `<script>` and `<link/>`
* Add support for FreeMarker tags `<@s.script>` and `<@s.link>`

Co-authored-by: Ecenaz Jen Ozmen - @eozmen410 
Co-authored-by: Giannis Chatziveroglou - @gchatz22 
Co-authored-by: Santiago Diaz - @salcho

### DIFF
--- a/core/src/main/resources/org/apache/struts2/interceptor/debugging/console.ftl
+++ b/core/src/main/resources/org/apache/struts2/interceptor/debugging/console.ftl
@@ -21,7 +21,9 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <script type="text/javascript">
+    <script type="text/javascript" <#rt/>
+    <#include "/${parameters.templateDir}/simple/nonce.ftl" />
+    >
     var baseUrl = "<@s.url value="/struts" includeParams="none"/>";
     window.open(baseUrl+"/webconsole.html", 'OGNL Console','width=500,height=450,status=no,toolbar=no,menubar=no');
     </script>    
--- a/core/src/main/resources/org/apache/struts2/interceptor/debugging/console.ftl
+++ b/core/src/main/resources/org/apache/struts2/interceptor/debugging/console.ftl
@@ -21,9 +21,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <script type="text/javascript" <#rt/>
-    <#include "/${parameters.templateDir}/simple/nonce.ftl" />
-    >
+    <script type="text/javascript" <#include "/${parameters.templateDir}/simple/nonce.ftl" /> >
     var baseUrl = "<@s.url value="/struts" includeParams="none"/>";
     window.open(baseUrl+"/webconsole.html", 'OGNL Console','width=500,height=450,status=no,toolbar=no,menubar=no');
     </script>    
--- a/core/src/main/resources/template/css_xhtml/head.ftl
+++ b/core/src/main/resources/template/css_xhtml/head.ftl
@@ -18,7 +18,5 @@
  * under the License.
  */
 -->
-<link
-        <#include "/${parameters.templateDir}/simple/nonce.ftl" />
-        rel="stylesheet" href="<@s.url value='/struts/css_xhtml/styles.css' includeParams='none' encode='false' />" type="text/css" />
+<link <#include "/${parameters.templateDir}/simple/nonce.ftl" /> rel="stylesheet" href="<@s.url value='/struts/css_xhtml/styles.css' includeParams='none' encode='false' />" type="text/css" />
 <#include "/${parameters.templateDir}/simple/head.ftl" />
--- a/core/src/main/resources/template/css_xhtml/head.ftl
+++ b/core/src/main/resources/template/css_xhtml/head.ftl
@@ -18,5 +18,7 @@
  * under the License.
  */
 -->
-<link rel="stylesheet" href="<@s.url value='/struts/css_xhtml/styles.css' includeParams='none' encode='false' />" type="text/css" />
+<link
+        <#include "/${parameters.templateDir}/simple/nonce.ftl" />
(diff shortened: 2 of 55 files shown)


lukaszlenart:
Looks good good, yet some tests are failing - it looks like out integration tests cannot pass because an app cannot be started.

core/src/main/resources/org/apache/struts2/interceptor/debugging/console.ftl
```diff
@@ -21,7 +21,9 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <script type="text/javascript">
+    <script type="text/javascript" <#rt/>
+    <#include "/${parameters.templateDir}/simple/nonce.ftl" />
```
JCgH4164838Gh792C124B5, line 25: Should the changes being made in this PR be collapsed into single lines (where possible) to match the visual structure of the original tag(s) ?

For example in this case, utilize a single line such as:
`<script type="text/javascript" <#include "/${parameters.templateDir}/simple/nonce.ftl" /> >`

to replace the three lines would eliminate the need for extra `<#rt/>` tag processing in the template and produce a visual result closer to the original tag.

core/src/main/resources/template/css_xhtml/head.ftl
```diff
@@ -18,5 +18,7 @@
  * under the License.
  */
 -->
-<link rel="stylesheet" href="<@s.url value='/struts/css_xhtml/styles.css' includeParams='none' encode='false' />" type="text/css" />
+<link
+        <#include "/${parameters.templateDir}/simple/nonce.ftl" />
```
JCgH4164838Gh792C124B5, line 22: And here:
`<link <#include "/${parameters.templateDir}/simple/nonce.ftl" /> rel="stylesheet" href="<@s.url value='/struts/css_xhtml/styles.css' includeParams='none' encode='false' />" type="text/css" />`

This sort of structure repeats in several of the PR's code changes, so those could be modified in a similar way if it makes sense.

core/src/main/resources/template/simple/form-close-tooltips.ftl
```diff
@@ -24,7 +24,11 @@
 --><#t/>
 <#if (parameters.hasTooltip!false)><#t/>
 	<#lt/><!-- javascript that is needed for tooltips -->
-	<#lt/><script type="text/javascript" src='<@s.url value="/struts/domTT.js" includeParams="none" encode="false" />'></script>
-	<#lt/><link rel="stylesheet" type="text/css" href="<@s.url value="/struts/domTT.css" includeParams="none" encode="false" />"/>
+	<#lt/><script type="text/javascript" src='<@s.url value="/struts/domTT.js" includeParams="none" encode="false" />'<#rt/>
```
JCgH4164838Gh792C124B5, line 27: If the `<#rt/>` tag structure is kept, should this one include a space to be consistent with the others ?
i.e.  `/>' <#rt/>`

JCgH4164838Gh792C124B5 (commented):
Hi.  Just a general comment on the template layout.  It is always something that could be considered in a follow-up PR.

JCgH4164838Gh792C124B5:
Hi.  According to the Travis CI report, only the JDK 8 run failed (JDK 9 and 11 passed).  There does not seem to be any changes in the PR that would pose a problem for JDK 8, so it seems weird.

Maybe @eozmen410 could confirm that their branch completes a local build using JDK 8 as well ?  If a local JDK 8 build works (or fails), it might help narrow down a possible cause for the failure during the CI build.

yasserzamani:
Hi @JCgH4164838Gh792C124B5 , I've enabled integration tests only for JDK8's build (using STRUTS_IT=true variable in travis you can see) and it's the reason behind why it fails only. It really runs showcase app in an embedded jetty and tests the app via jwebunit real http requests/responses.

core/src/main/resources/template/simple/nonce.ftl
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+-->
+<#if parameters.nonce?has_content>
+    nonce="${parameters.nonce}"<#rt/>
```
yasserzamani, line 22: missed var and semicolon? i.e. shouldn't it be `var nonce="${parameters.nonce}";`

yasserzamani (approved):
OK I fixed that :)

Great Job :100: Thanks :1st_place_medal: LGTM

salcho:
Hi @JCgH4164838Gh792C124B5 , 

I've pushed a small commit that removes all the `<#rt>` tags as you mentioned. I agree that it's cleaner this way, thanks for that and thanks @yasserzamani for fixing the JDK issue 👍

JCgH4164838Gh792C124B5 (approved):
Thanks @eozmen410 @salcho @yasserzamani , for your efforts related to this PR.  😄 
The changes LGTM.  👍

PR #435, opened 2020-09-04
[WW-5088] Empty file upload, storeLocation null gives wrong error message.

gregh3269:
On an empty file upload the storeLocation is null, so adds a file new StrutsUploadedFile(null)

JakartaMultiPartRequest

List fileList = new ArrayList<>(items.size());
for (FileItem fileItem : items) {
File storeLocation = ((DiskFileItem) fileItem).getStoreLocation();
if (fileItem.isInMemory() && storeLocation != null && !storeLocation.exists()) {
try {
storeLocation.createNewFile();
} catch (IOException e) {
LOG.error("Cannot write uploaded empty file to disk: {}", storeLocation.getAbsolutePath(), e);
}
}
fileList.add(new StrutsUploadedFile(storeLocation));
}

The FileUploadInterceptor checks for a null file but not null content from the new StrutsUploadedFile(null).

eg for an empty file main.js

Error message should be:
Error uploading: main.js
but gives:
Error setting expression 'uploadedFiles' with value ['org.apache.struts2.dispatcher.multipart.StrutsUploadedFile@202585bc', ]
Error setting expression 'uploadedFiles' with value ['org.apache.struts2.dispatcher.multipart.StrutsUploadedFile@49553150', ]

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/FileUploadInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/FileUploadInterceptor.java
@@ -345,6 +345,14 @@ public class FileUploadInterceptor extends AbstractInterceptor {
             if (LOG.isWarnEnabled()) {
                 LOG.warn(errMsg);
             }
+        } else if (file.getContent() == null) {
+            String errMsg = getTextMessage(action, "struts.messages.error.uploading", new String[]{filename});
+            if (validation != null) {
+                validation.addFieldError(inputName, errMsg);
+            }
+            if (LOG.isWarnEnabled()) {
+                LOG.warn(errMsg);
+            }
         } else if (maximumSize != null && maximumSize < file.length()) {
             String errMsg = getTextMessage(action, "struts.messages.error.file.too.large", new String[]{inputName, filename, file.getName(), "" + file.length(), getMaximumSizeStr(action)});
             if (validation != null) {


JCgH4164838Gh792C124B5:
Hi.

Although it seems unlikely, there might be someone with a current use case for uploading empty files.  For this PR's change, maybe there should be a different error message string lookup for the new check added with this PR (e.g. "struts.messages.error.file.empty") to make it clear why the upload is being rejected ?

Another idea might be to introduce a "minimumSize" property for the file upload that developers can configure (with a default of 0 or 1, depending on whether accepting empty files should be allowed by default).  That sort of feature could replace an empty guard with something configurable.

lukaszlenart:
That may be a good addition but I would target 2.6 for such changes. I don't know if @gregh3269 would like to port this PR back to 2.5.x

gregh3269:
I did consider a struts.messages.error.file.empty, but the patch was more related to the incorrect message rather than the semantics of what JakartaMultiPartRequest does, there maybe other reasons for the null storeLocation  if/IOException.
From the snip the logging is odd as the storeLocation is null for an empty file (I have enctype="multipart/form-data" for the upload form) so never gets to the  IOException.  There may be other implementations?

@lukaszlenart For the PR on 2.5.x I do not have any notes on how to do this :-(

lukaszlenart:
@gregh3269 just branch off of `struts-2-5-x` and then cherry-pick your commit, push your branch and open PR against `struts-2-5-x`

PR #438, opened 2020-09-14
WW-5087 handle Parameter.Empty properly

yasserzamani:
There was a bug with AliasInterceptor not handling the Parameter.Empty that is returned from HttpParameters.get(). Since HttpParameters.get() always returns a non-null value, the Evaluated object is treated as always being defined, which results in the empty value being set incorrectly on the stack.


lukaszlenart (approved):
LGTM 👍

PR #441, opened 2020-10-14
WW-5088: Integration test for empty file upload.

pys-arhs:
Add integration test to detect regressions if commons-fileupload behavior changes in the future.

DiskFileItemFactory#setSizeThreshold should be set to -1 because org.apache.commons.io.output.ThresholdingOutputStream#checkThreshold checks if the current file size is strictly greater than the threshold.


lukaszlenart:
LGTM 👍

PR #442, opened 2020-10-15
TagAttribute

lukaszlenart:
WIP

### DIFF
--- a/core/src/main/java/org/apache/struts2/views/TagAttribute.java
+++ b/core/src/main/java/org/apache/struts2/views/TagAttribute.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/apache/struts2/views/TagAttribute.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String stripedExpression() {
+        if (isExpression()) {
+            return value.substring(2, value.length() - 1);
+        } else {
+            return value;
+        }
+    }
+
+    public TagAttribute escaped(){
```
JCgH4164838Gh792C124B5, line 54: Minor suggestion (spacing/readability):  change `(){`  to `() {`

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

This looks like an interesting change.   😄  
I will try to review it in more detail when I have a chance.  It looks like the change might be introducing a failure (java.lang.ClassCastException) in the Showcase tests, as it stands right now (if I read the Travis CI reports correctly).

JCgH4164838Gh792C124B5 (changes requested):
Hi @lukaszlenart .

I downloaded the branch for these changes, took another look, and attempted to build things locally.

Adding the additional methods suggested in the review and then modifying the components (either using the new conversion method, or relying on the implicit `toString()` addition) seemed to work for me locally.  This allowed the Showcase test that was failing to pass, and interactive tests for the Showcase UI components seemed to work OK.

I do not know if the changes suggested in this review are a good strategy or not, but I think in order to implement the `TagAttribute` there will need to be changes in places where the code assumes a component parameter will be a `TagAttribute` or `String` (so that it can handle either).

Depending on how many different types of `Component` parameter values might be encountered, some sort of general-purpose adapter might be a consideration (to convert from the various parameter types to `String` and back).  It does not look like it would be worth the overhead at the moment, though.

It looks like the idea for your PR works, with just a few boundary cases to deal with first.  Good work.  👍

lukaszlenart:
I'm closing this one - there are some tags that makes implementing this quite hard.

PR #445, opened 2020-11-12
Fix meaningless close

lcc:
Meaningless Close: In several classes, close(), specified in Closeable interface, has no effect and other methods in those classes can be called after close() without IOException.

### DIFF
--- a/plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/JspReader.java
+++ b/plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/JspReader.java
@@ -209,7 +209,6 @@ class JspReader {
         CharArrayWriter caw = new CharArrayWriter();
         while (!stop.equals(mark()))
             caw.write(nextChar());
-        caw.close();
         reset(oldstart);
         return caw.toString();
     }
@@ -595,7 +594,6 @@ class JspReader {
             char buf[] = new char[1024];
             for (int i = 0 ; (i = reader.read(buf)) != -1 ;)
                 caw.write(buf, 0, i);
-            caw.close();
             if (current == null) {
                 current = new Mark(this, caw.toCharArray(), fileid, 
                                    getFile(fileid), master, encoding);
--- a/plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/Parser.java
+++ b/plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/Parser.java
@@ -324,7 +324,6 @@ class Parser implements TagConstants {
                 ++i;
             }
         }
-        cw.close();
         return cw.toString();
     }
 


lukaszlenart:
Tests fail

lcc:
Hi @lukaszlenart thanks for the prompt response. The error that causes the pipeline to fail is "ClassNotFoundException org.apache.maven.surefire.booter.ForkedBooter" which also happens in the current master build. 

This is a [well known error](https://issues.apache.org/jira/browse/SUREFIRE-1588) in the maven surefire plugin, in order to "fix" this one must update to a surefire version to one that corrected this, such as 3.0.0-M1.

After updating it on my surefire-plugin, this issue ceases to exist and all tests pass.

This PR was purely meant to fix unecessary uses of the close method, but if you want I can write a patch for the surefire-plugin.

I didn't do it before because there may be some underlying reason to use it the way it's being used.

lukaszlenart:
As far I see there is a compilation error
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/travis/build/apache/struts/plugins/portlet-mocks/src/test/java/org/apache/struts2/StrutsSpringPortletMockObjectsTest.java:[225,78] unreported exception java.io.IOException; must be caught or declared to be thrown
```

https://travis-ci.org/github/apache/struts/jobs/743068291

plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/JspReader.java
```diff
@@ -209,7 +209,6 @@ String getText(Mark start, Mark stop) throws JasperException {
         CharArrayWriter caw = new CharArrayWriter();
         while (!stop.equals(mark()))
             caw.write(nextChar());
-        caw.close();
```
JCgH4164838Gh792C124B5, line 212: The API Docs and JDK source confirm that for `CharArrayWriter` instances, calls to `close()` are no-ops (no effect).
So I think this should be a safe change.

plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/JspReader.java
```diff
@@ -595,7 +594,6 @@ private void pushFile(String file, String encoding,
             char buf[] = new char[1024];
             for (int i = 0 ; (i = reader.read(buf)) != -1 ;)
                 caw.write(buf, 0, i);
-            caw.close();
```
JCgH4164838Gh792C124B5, line 598: Same goes for here.

plugins/embeddedjsp/src/main/java/org/apache/struts2/jasper/compiler/Parser.java
```diff
@@ -324,7 +324,6 @@ private String parseScriptText(String tx) {
                 ++i;
             }
         }
-        cw.close();
```
JCgH4164838Gh792C124B5, line 327: And here.

plugins/portlet-mocks/src/test/java/org/apache/struts2/StrutsSpringPortletMockObjectsTest.java
```diff
@@ -222,7 +222,8 @@ public void testMockClientDataRequest() {
         assertEquals("Content-type not as expected ?", TEST_CONTENT_TYPE, mockClientDataRequest.getContentType());
         mockClientDataRequest.setMethod(TEST_METHOD);
         assertEquals("Method not as expected ?", TEST_METHOD, mockClientDataRequest.getMethod());
-        try ( InputStream inputStream = mockClientDataRequest.getPortletInputStream() ) {
+	InputStream inputStream = mockClientDataRequest.getPortletInputStream();
```
JCgH4164838Gh792C124B5, line 225: This change in the PR should probably be reverted.  The previous try-with-resources call is safer, and restoring that code should eliminate the `IOException` issue currently being reported during the compilation.

JCgH4164838Gh792C124B5 (changes requested):
Hello Lucas.

I think that the `CharArrayWriter` `close()` changes are OK, being no-ops as you pointed out.
The change to `StrutsSpringPortletMockObjectsTest` should probably be reverted, though, since it causes the compilation of your PR to fail.

Thanks,
James.

plugins/portlet-mocks/src/test/java/org/apache/struts2/StrutsSpringPortletMockObjectsTest.java
lcc: Thanks for the suggestion, the PR has now been updated!

lukaszlenart (approved):
LGTM 👍

PR #446, opened 2020-11-13
[WW-5096] Fix StaticParametersInterceptor param overwrite

lukaszlenart:
Fixes [WW-5096](https://issues.apache.org/jira/browse/WW-5096)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/StaticParametersInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/StaticParametersInterceptor.java
@@ -227,22 +227,13 @@ public class StaticParametersInterceptor extends AbstractInterceptor {
     protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
         HttpParameters previousParams = ac.getParameters();
 
-        HttpParameters.Builder combinedParams = HttpParameters.create();
+        HttpParameters.Builder combinedParams;
         if (overwrite) {
-            if (previousParams != null) {
-                combinedParams = combinedParams.withParent(previousParams);
-            }
-            if (newParams != null) {
-                combinedParams = combinedParams.withExtraParams(newParams);
-            }
+            combinedParams = HttpParameters.create().withParent( previousParams);
+            combinedParams = combinedParams.withExtraParams(newParams);
         } else {
-            if (newParams != null) {
-                HttpParameters newHttpParameters = HttpParameters.create(newParams).build();
-                combinedParams = combinedParams.withParent(newHttpParameters);
-            }
-            if (previousParams != null) {
-                combinedParams = combinedParams.withExtraParams(previousParams);
-            }
+            combinedParams = HttpParameters.create(newParams);
+            combinedParams = combinedParams.withExtraParams(previousParams);
         }
         ac.setParameters(combinedParams.build());
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/interceptor/StaticParametersInterceptor.java
```diff
@@ -237,7 +237,8 @@ protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams
             }
         } else {
             if (newParams != null) {
-                combinedParams = combinedParams.withExtraParams(newParams);
+                HttpParameters newHttpParameters = HttpParameters.create(newParams).build();
+                combinedParams = combinedParams.withParent(newHttpParameters);
             }
             if (previousParams != null) {
                 combinedParams = combinedParams.withParent(previousParams);
```
akash5885, line 244: @lukaszlenart I think for this one means for previous params, combinedParams.withParent(previousParams) needs to replace with combinedParams.withExtraParams(previousParams) otherwise extra parameters which are in newParams and not in previousParams will be lost. 
Please let me know if I'm missing anything here.

lukaszlenart: Not sure if I got your point, yet please check the latest change. The problem is that I don't have a test case to reproduce the problem.

JCgH4164838Gh792C124B5:
Hi @lukaszlenart  and @akash5885 .

It looks like the original code of `StaticParametersInterceptor` `addParametersToContext()` was doing the same operations if `overwrite` was `true `or `false`, just in reverse-order depending on the `boolean `value. I think that was what @akash5885 stated in the JIRA ?

Did the different ordering of the old code blocks have any effect at all, or was the end result the same whether `overwrite` was `true` or `false` ?

Is there a possibility for a manual test example or test case that clarifies how the original code was not working as intended, and how the updated code does ?

I am not sure I am understanding the original problem correctly, so I figured I would ask.  😄

akash5885:
HI @JCgH4164838Gh792C124B5 @lukaszlenart 
Apologies I didn't mention exact scenario that triggered this bug. But now I have added scenario in JIRA ticket [WW-5096](https://issues.apache.org/jira/browse/WW-5096) which triggered this bug.

@JCgH4164838Gh792C124B5
yes, you are correct.original code of StaticParametersInterceptor addParametersToContext() was doing the same operations irrespective of overwrite is true or false.

akash5885 (approved):
LGTM !

yasserzamani:
Looks like the bug has emerged by [here](https://github.com/apache/struts/commit/5508352ddb46417ccd44033064ea337da509c021#diff-680ecf9ee762502e27c18bab70f879a7772ed76880d1473061133487d2bc75b7) and we can just restore previous logic.

One test fails. And not sure if we simple have new test case from provided scenario?

lukaszlenart:
@akash5885 thanks a lot for you description and test case - based on that I was able to prepare a few unit tests to cover the logic. Please review and let me know.

akash5885 (approved):
@lukaszlenart Thanks for your quick code fix and unit test cases. Code fix and Unit test cases looks good to me !

yasserzamani:
LGTM :100: Merely I can't recall if integration tests were done in PR builds but currently I see they aren't. Basically I remember `STRUTS_IT=true` was defined by me and was required to be set in Travis params, any idea?

yasserzamani:
Ooops sorry that was on only master branch and PRs; I realized now

yasserzamani:
OK I cherry picked against master; Lets merge if passed integration tests there as well :)

PR #448, opened 2020-11-17
[WW-5098] Upgrade asm to 9.0

sepe81:
with JDK 16 support (sealed classes)
according to https://asm.ow2.io/versions.html


lukaszlenart:
Would you mind creating a JIRA ticket to control what was upgraded? Thanks in advance!

yasserzamani:
LGTM :100:

PR #449, opened 2020-11-19
[WW-5096] Fix StaticParametersInterceptor param overwrite

yasserzamani:
Fixes [WW-5096](https://issues.apache.org/jira/browse/WW-5096)


lukaszlenart:
LGTM 👍

lukaszlenart:
I have noticed that this changes #447 need to be merged first to avoid conflicts.

PR #450, opened 2020-11-21
[WW-3730] Avoids conversion to String[] of passed params

lukaszlenart:
Fixes [WW-3730](https://issues.apache.org/jira/browse/WW-3730)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/ActionComponent.java
+++ b/core/src/main/java/org/apache/struts2/components/ActionComponent.java
@@ -214,24 +214,10 @@ public class ActionComponent extends ContextBean {
             parentParams = getStack().getActionContext().getParameters();
         }
 
-        HttpParameters.Builder builder = HttpParameters.create();
-        if (parentParams != null) {
-            builder = builder.withParent(parentParams);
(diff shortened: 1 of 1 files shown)


yasserzamani:
I wonder if it's a breaking change why still [this test](https://github.com/apache/struts/blob/a4793d27cf19546516167b1749e8d124c731e93a/core/src/test/java/org/apache/struts2/components/ActionComponentTest.java#L33) passes. Couldn't find why!

lukaszlenart:
Because everything is an Object right now, there is no conversion to String, which was not needed tbh. And it was potentially  breaking change, maybe it isn't :)

yasserzamani:
Thanks! I found it :) Test has been changed through [WW-4572 Http parameters](https://github.com/apache/struts/pull/58) and I was watching original test. So basically yes then it's a breaking change, because the modified test doesn't try anymore that the value is always a `string[]` (The original test was doing via `((String[])params.get("foo"))[0]` which implicitly tests conversion to `string[]`).

yasserzamani (approved):
LGTM :+1:

PR #452, opened 2020-11-29
[WW-5078] Removes support for <xwork> DTDs

lukaszlenart:
Implements [WW-5078](https://issues.apache.org/jira/browse/WW-5078)


lukaszlenart:
I assume a silence consensus here, LGTM 👍

PR #453, opened 2020-12-06
[WW-3877] Drops altSyntax option

lukaszlenart:
`altSyntax` was an old approach from WebWork to support `%{..}` expressions. Right now it doesn't make sense to disable this option (set `false`) as in most cases this means an app won't work. Keeping support for the `altSyntax` option is useless as all expression must use `%{..}` syntax to work.

Implements [WW-3877](https://issues.apache.org/jira/browse/WW-3877)
It will be merged under [Lazy Consensus](http://community.apache.org/committers/lazyConsensus.html)


lukaszlenart:
LGTM 👍

PR #455, opened 2020-12-07
sort getMethods for deterministic iteration order

Yilinwang:
The tests `org.apache.struts2.json.SMDMethodInterfaceTest#testBaseClassOnly` and `org.apache.struts2.json.SMDMethodInterfaceTest#testInterfaces` can fail due to a different iteration order of `Object.getMethods()`. The failures are as follows:
```
[ERROR] org.apache.struts2.json.SMDMethodInterfaceTest.testBaseClassOnly  Time elapsed: 0.008 s  <<< FAILURE!
junit.framework.ComparisonFailure: expected:<get[Z]> but was:<get[X]>
    at org.apache.struts2.json.SMDMethodInterfaceTest.testBaseClassOnly(SMDMethodInterfaceTest.java:136)

[ERROR] org.apache.struts2.json.SMDMethodInterfaceTest.testInterfaces  Time elapsed: 0 s  <<< FAILURE!
junit.framework.ComparisonFailure: expected:<get[Z]> but was:<get[X]>
    at org.apache.struts2.json.SMDMethodInterfaceTest.testInterfaces(SMDMethodInterfaceTest.java:155)
```

The fix is to sort the result of `getMethods()` so that the iteration order remains stable and the failure will not occur any more. In this way, the test will be more stable.

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
@@ -411,9 +411,9 @@ public class JSONUtil {
     public static Method[] listSMDMethods(Class clazz, boolean ignoreInterfaces) {
         final List<Method> methods = new LinkedList<>();
         if (ignoreInterfaces) {
-            Method[] SMDMethods = clazz.getMethods();
-            Arrays.sort(SMDMethods, (a, b) -> a.toString().compareTo(b.toString()));
-            for (Method method : SMDMethods) {
+            Method[] classMethods = clazz.getMethods();
+            Arrays.sort(classMethods, (a, b) -> a.toString().compareTo(b.toString()));
+            for (Method method : classMethods) {
                 SMDMethod smdMethodAnnotation = method.getAnnotation(SMDMethod.class);
                 if (smdMethodAnnotation != null) {
                     methods.add(method);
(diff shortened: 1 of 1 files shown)


plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
```diff
@@ -410,7 +411,9 @@ else if (serializationParams.getErrorCode() > 0)
     public static Method[] listSMDMethods(Class clazz, boolean ignoreInterfaces) {
         final List<Method> methods = new LinkedList<>();
         if (ignoreInterfaces) {
-            for (Method method : clazz.getMethods()) {
+            Method[] SMDMethods = clazz.getMethods();
```
lukaszlenart, line 414: Could you use `methods` instead of `SMDMethods`?

plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
```diff
@@ -421,7 +424,9 @@ else if (serializationParams.getErrorCode() > 0)
             // order encountered
             JSONUtil.visitInterfaces(clazz, new JSONUtil.ClassVisitor() {
                 public boolean visit(Class aClass) {
-                    for (Method method : aClass.getMethods()) {
+                    Method[] SMDMethods = aClass.getMethods();
```
lukaszlenart, line 427: Same here, `methods` instead of `SMDMethods`?

lukaszlenart (commented):
Looks good, just two minor things

plugins/json/src/main/java/org/apache/struts2/json/JSONUtil.java
Yilinwang: I think methods is used at 412 `final List<Method> methods = new LinkedList<>();`

Yilinwang (commented):
Thanks for replying. I think methods is used before, how about classMethods

lukaszlenart:
`classMethods` works as well 👍

Yilinwang:
fixed!

lukaszlenart:
LGTM 👍

PR #462, opened 2021-01-02
[WW-5021] Introduces a new constant to allow define a path to static content

lukaszlenart:
Implements [WW-5021](https://issues.apache.org/jira/browse/WW-5021)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/UIBean.java
+++ b/core/src/main/java/org/apache/struts2/components/UIBean.java
@@ -30,6 +30,7 @@ import org.apache.struts2.components.template.Template;
 import org.apache.struts2.components.template.TemplateEngine;
 import org.apache.struts2.components.template.TemplateEngineManager;
 import org.apache.struts2.components.template.TemplateRenderingContext;
+import org.apache.struts2.dispatcher.StaticContentLoader;
 import org.apache.struts2.util.TextProviderHelper;
 import org.apache.struts2.views.annotations.StrutsTagAttribute;
 import org.apache.struts2.views.util.ContextUtil;
@@ -526,8 +527,8 @@ public abstract class UIBean extends Component {
     }
 
     @Inject(StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH)
-    public void setUiStaticContentPath(String uiStaticContentPath) {
-        this.uiStaticContentPath = uiStaticContentPath;
+    public void setStaticContentPath(String uiStaticContentPath) {
+        this.uiStaticContentPath = StaticContentLoader.Validator.validateStaticContentPath(uiStaticContentPath);
     }
 
     @Inject
--- a/core/src/main/java/org/apache/struts2/components/UIBean.java
+++ b/core/src/main/java/org/apache/struts2/components/UIBean.java
@@ -317,7 +317,7 @@ import java.util.Map;
  *    <tr>
  *      <td>tooltipIcon</td>
  *      <td>String</td>
- *      <td>/struts/static/tooltip/tooltip.gif</td>
+ *      <td>/static/tooltip/tooltip.gif</td>
  *      <td>The url to the tooltip icon</td>
  *   <tr>
  *      <td>tooltipDelay</td>
@@ -506,6 +506,8 @@ public abstract class UIBean extends Component {
     protected String defaultTemplateDir;
     protected String defaultUITheme;
     protected String uiThemeExpansionToken;
+    protected String uiStaticContentPath;
+
     protected TemplateEngineManager templateEngineManager;
 
     @Inject(StrutsConstants.STRUTS_UI_TEMPLATEDIR)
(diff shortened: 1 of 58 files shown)


JCgH4164838Gh792C124B5 (changes requested):
Hi Lukasz.
The change looks pretty good to me.  :)
The only suggestion to consider is the possibility of adding some extra guards (or log warnings) against "bad" values for the path parameter (something that was not needed when it was always a constant "/static").

core/src/main/java/org/apache/struts2/components/UIBean.java
```diff
@@ -523,6 +525,11 @@ public void setUIThemeExpansionToken(String uiThemeExpansionToken) {
         this.uiThemeExpansionToken = uiThemeExpansionToken;
     }
 
+    @Inject(StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH)
+    public void setUiStaticContentPath(String uiStaticContentPath) {
+        this.uiStaticContentPath = uiStaticContentPath;
```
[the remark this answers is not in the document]
lukaszlenart: Right, added warning and fallback

core/src/main/java/org/apache/struts2/config/entities/ConstantConfig.java
```diff
@@ -1344,4 +1346,12 @@ public Integer getOgnlAutoGrowthCollectionLimit() {
     public void setOgnlAutoGrowthCollectionLimit(Integer ognlAutoGrowthCollectionLimit) {
         this.ognlAutoGrowthCollectionLimit = ognlAutoGrowthCollectionLimit;
     }
+
+    public String getStaticContentPath() {
+        return staticContentPath;
+    }
+
+    public void setStaticContentPath(String staticContentPath) {
+        this.staticContentPath = staticContentPath;
```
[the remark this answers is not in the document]
lukaszlenart: Same here

core/src/main/java/org/apache/struts2/dispatcher/DefaultStaticContentLoader.java
```diff
@@ -100,20 +110,23 @@
...
+     * @param serveStaticContent New setting
      */
     @Inject(StrutsConstants.STRUTS_SERVE_STATIC_CONTENT)
     public void setServeStaticContent(String serveStaticContent) {
         this.serveStatic = BooleanUtils.toBoolean(serveStaticContent);
     }
 
+    @Inject(StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH)
+    public void setStaticContentPath(String staticContentPath) {
+        this.staticContentPath = staticContentPath;
```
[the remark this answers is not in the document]
lukaszlenart: Right, done

core/src/main/java/org/apache/struts2/dispatcher/StaticContentLoader.java
```diff
@@ -57,4 +67,32 @@
...
+                    uiStaticContentPath,
+                    StaticContentLoader.DEFAULT_STATIC_CONTENT_PATH);
+                return StaticContentLoader.DEFAULT_STATIC_CONTENT_PATH;
+            } else if ("/".equals(uiStaticContentPath)) {
+                LOG.warn("\"{}\" cannot be set to \"{}\", falling back into default value \"{}\"",
+                    StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH,
+                    uiStaticContentPath,
+                    StaticContentLoader.DEFAULT_STATIC_CONTENT_PATH);
+                return StaticContentLoader.DEFAULT_STATIC_CONTENT_PATH;
+            } else if(!uiStaticContentPath.startsWith("/")) {
```
JCgH4164838Gh792C124B5, line 87: Minor spacing item: "} else if(" could be "} else if ("

core/src/main/java/org/apache/struts2/dispatcher/StaticContentLoader.java
```diff
@@ -57,4 +67,32 @@
...
+                LOG.warn("\"{}\" must start with \"/\", but has been set to \"{}\", prepending the missing \"/\"!",
+                    StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH,
+                    uiStaticContentPath);
+                return "/" + uiStaticContentPath;
+            } else {
```
JCgH4164838Gh792C124B5, line 92: I think there may be one validation condition missing from the checks (trailing "/"), as it looks like that would break processing as well.  Adding a new clause like this:
```
} else if (uiStaticContentPath.endsWith("/")) {
    LOG.warn("\"{}\" must not end with \"/\", but has been set to \"{}\", removing all trailing \"/\"!",
        StrutsConstants.STRUTS_UI_STATIC_CONTENT_PATH,
        uiStaticContentPath);
    return StringUtils.stripEnd(uiStaticContentPath, "/");
}
```
before the final else might work.  What do you think ?

JCgH4164838Gh792C124B5 (changes requested):
Hi Lukasz.
Thanks for reviewing and adding the static path validator.  :)
There are two small suggestions for consideration in the most recent PR changes, which you can look at and decide whether they make sense or not.

core/src/main/java/org/apache/struts2/dispatcher/StaticContentLoader.java
lukaszlenart: fixed!

core/src/main/java/org/apache/struts2/dispatcher/StaticContentLoader.java
lukaszlenart: Right, on the first thought I want to make the trailing "/" required, but after all I decided the above approach is better.

JCgH4164838Gh792C124B5 (approved):
The latest changes look good to me.  👍

PR #463, opened 2021-01-02
[WW-4742] Stops evaluating default message as a OGNL expression

lukaszlenart:
and uses proper order for escaping

Fixes [WW-4742](https://issues.apache.org/jira/browse/WW-4742)


lukaszlenart:
LGTM 👍

PR #464, opened 2021-01-02
[WW-5000] Extracts string literals as constants

lukaszlenart:
Implements [WW-5000](https://issues.apache.org/jira/browse/WW-5000)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
@@ -31,6 +31,7 @@ import com.opensymphony.xwork2.util.ClassLoaderUtil;
 import com.opensymphony.xwork2.validator.Validator;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.struts2.StrutsConstants;
 
 import java.io.Serializable;
 import java.util.Map;
@@ -68,7 +69,7 @@ public class ObjectFactory implements Serializable {
         this.container = container;
     }
 
-    @Inject(value="objectFactory.classloader", required=false)
+    @Inject(value= StrutsConstants.STRUTS_OBJECT_FACTORY_CLASSLOADER, required=false)
     public void setClassLoader(ClassLoader cl) {
         this.ccl = cl;
     }
--- a/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
@@ -69,7 +69,7 @@ public class ObjectFactory implements Serializable {
         this.container = container;
     }
 
-    @Inject(value= StrutsConstants.STRUTS_OBJECT_FACTORY_CLASSLOADER, required=false)
+    @Inject(value=StrutsConstants.STRUTS_OBJECT_FACTORY_CLASSLOADER, required=false)
     public void setClassLoader(ClassLoader cl) {
         this.ccl = cl;
     }
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
@@ -284,7 +284,7 @@ public class DefaultConfiguration implements Configuration {
         builder.factory(ValueSubstitutor.class, EnvsValueSubstitutor.class, Scope.SINGLETON);
 
         builder.constant(StrutsConstants.STRUTS_DEVMODE, "false");
-        builder.constant(StrutsConstants.STRUTS_LOG_MISSING_PROPERTIES, "false");
+        builder.constant(StrutsConstants.STRUTS_OGNL_LOG_MISSING_PROPERTIES, "false");
         builder.constant(StrutsConstants.STRUTS_ENABLE_OGNL_EVAL_EXPRESSION, "false");
         builder.constant(StrutsConstants.STRUTS_ENABLE_OGNL_EXPRESSION_CACHE, "true");
         builder.constant(StrutsConstants.STRUTS_CONFIGURATION_XML_RELOAD, "false");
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/opensymphony/xwork2/ObjectFactory.java
```diff
@@ -68,7 +69,7 @@ public void setContainer(Container container) {
         this.container = container;
     }
 
-    @Inject(value="objectFactory.classloader", required=false)
+    @Inject(value= StrutsConstants.STRUTS_OBJECT_FACTORY_CLASSLOADER, required=false)
```
sepe81, line 72:
```suggestion
    @Inject(value=StrutsConstants.STRUTS_OBJECT_FACTORY_CLASSLOADER, required=false)
```

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -284,7 +284,7 @@ protected Container createBootstrapContainer(List<ContainerProvider> providers)
         builder.factory(ValueSubstitutor.class, EnvsValueSubstitutor.class, Scope.SINGLETON);
 
         builder.constant(StrutsConstants.STRUTS_DEVMODE, "false");
-        builder.constant(StrutsConstants.STRUTS_LOG_MISSING_PROPERTIES, "false");
+        builder.constant(StrutsConstants.STRUTS_OGNL_LOG_MISSING_PROPERTIES, "false");
```
sepe81, line 287: Does this mean, the OGNL log constant is preferred over the non OGNL constant?

sepe81: Oh, I see it is a rename under https://github.com/apache/struts/pull/464/files#diff-2de2f780034a74f8de7c95a0a38f6bea1c97cff74d7060064f28dc6069a24a44

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -346,4 +346,9 @@
 
     /** See {@link com.opensymphony.xwork2.config.impl.AbstractMatcher#appendNamedParameters */
     public static final String STRUTS_MATCHER_APPEND_NAMED_PARAMETERS = "struts.matcher.appendNamedParameters";
+
+    public static final String STRUTS_CHAINING_COPY_ERRORS = "struts.chaining.copyErrors";
```
sepe81, line 350: I guess according to https://issues.apache.org/jira/browse/WW-5000?focusedCommentId=17257465&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17257465 the removal of `xwork.` is intentional?

struts.xwork.chaining.copyErrors gets struts.chaining.copyErrors

sepe81 (approved):
LGTM if my assumptions are correct

core/src/main/java/org/apache/struts2/StrutsConstants.java
lukaszlenart: Yes, this is a major release so I think it's a good time to change this. XWork doesn't exist as a seperated artefact anymore.

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
lukaszlenart: Yes, just a rename to match the same pattern.

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -250,7 +250,7 @@
     public static final String STRUTS_EL_THROW_EXCEPTION = "struts.el.throwExceptionOnFailure";
 
     /** Logs properties that are not found (very verbose) */
-    public static final String STRUTS_LOG_MISSING_PROPERTIES = "struts.ognl.logMissingProperties";
```
JCgH4164838Gh792C124B5, line 253: Hi Lukasz.
It is possible (if unlikely) that someone has been referencing `STRUTS_LOG_MISSING_PROPERTIES` from the constants file, and removing it will cause breakage.
Probably not a big concern, but maybe consider leave the old constant, have it set to the new one's value, and just mark the old one as deprecated ?
Something like:
```
/** Logs properties that are not found (very verbose) */
@deprecated as of 2.6.  Use {@link STRUTS_OGNL_LOG_MISSING_PROPERTIES} instead.
public static final String STRUTS_LOG_MISSING_PROPERTIES = STRUTS_OGNL_LOG_MISSING_PROPERTIES;

/** Logs properties that are not found (very verbose) */
public static final String STRUTS_OGNL_LOG_MISSING_PROPERTIES = "struts.ognl.logMissingProperties";
```
It might not really be necessary, but just something to consider.  :)

lukaszlenart: Done

core/src/test/java/com/opensymphony/xwork2/ognl/accessor/XWorkListPropertyAccessorTest.java
```diff
@@ -85,7 +85,7 @@ public void testAutoGrowthCollectionLimit() {
 
     public void testDeprecatedAutoGrowCollectionLimit() {
```
JCgH4164838Gh792C124B5, line 86: With the deprecated `setDeprecatedAutoGrowCollectionLimit()` being dropped, I think `testDeprecatedAutoGrowCollectionLimit()` now becomes a duplicate of `testAutoGrowthCollectionLimit()`.
If that is the case, then you might be able to drop `setDeprecatedAutoGrowCollectionLimit()` completely, if you want.

JCgH4164838Gh792C124B5 (approved):
LGTM. 👍

core/src/test/java/com/opensymphony/xwork2/ognl/accessor/XWorkListPropertyAccessorTest.java
lukaszlenart: Good catch, dropped!

PR #467, opened 2021-01-03
Proposed TextProvider feature addition (and cleanup)

JCgH4164838Gh792C124B5:
TextProvider feature addition and cleanup:
- Introduce (optional) control flag STRUTS_I18N_SEARCH_DEFAULTBUNDLES_FIRST to request TextProviders read from default resource bundles first, instead of their standard lookup ordering.  Defaults to false.
- Minor refactor of AbstractLocalizedTextProvider hierarchy to introduce getDefaultMessageWithAlternateKey() method that repackages existing logic used by the SrutsLocalizedTextProvider and GlobalLocalizedTextProvider.
- Update GlobalLocalizedTextProvider method comments to properly reflect the actual logic of some of its methods.
- New unit tests to confirm standard and default bundle first lookup ordering.
- Cleanup of some unused imports, a few typos, and code formatting items.
- Cleanup of default.properties comments to make them more consistent.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
@@ -61,6 +61,7 @@ abstract class AbstractLocalizedTextProvider implements LocalizedTextProvider {
     protected final ConcurrentMap<String, ResourceBundle> bundlesMap = new ConcurrentHashMap<>();
     protected boolean devMode = false;
     protected boolean reloadBundles = false;
+    protected boolean searchDefaultBundlesFirst = false;  // Search default resource bundles first.  Note: This flag may not be meaningful to all implementations.
 
     private final ConcurrentMap<MessageFormatKey, MessageFormat> messageFormats = new ConcurrentHashMap<>();
     private final ConcurrentMap<Integer, List<String>> classLoaderMap = new ConcurrentHashMap<>();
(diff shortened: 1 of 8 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This is an initial attempt to add a feature to the I18N TextProvider hierarchy to read resource entries from the default bundles first, and then continue with "normal" lookup if nothing is found.  It arose from a discussion on the developer list raised by G. Huber concerning the number of lookup calls when most resource entries are defined in the main application bundle.

This implementation allows a developer to "opt-in" to an alternative lookup strategy where the default bundles take precedence over entries anywhere else (if an entry does not exist in the default bundles, the class hierarchy bundles are then searched normally).  It should be slightly more efficient for applications that only use a "global" properties file (default resource bundle), while still providing lookups along the class hierarchy for any entries that are not defined in the default (global) resource bundles.  The default value of the flag is false, which preserves the normal (original) behaviour for resource lookup (requires explicitly setting the configuration to use the alternative lookup flow).

There is no JIRA yet for this item.  If after reviewing this PR there is support for the change, a JIRA can be opened to cover it for tracking.

lukaszlenart (approved):
Looks good to me, just add the JIRA ticket and I'm good :)

JCgH4164838Gh792C124B5:
A new JIRA WW-5112 has been opened to track the proposed change in this PR.
If there are any additional changes to consider, please let me know.

lukaszlenart (approved):
LGTM 👍

PR #468, opened 2021-01-10
[WW-4799] Adding struts.date.format as a conversion format for DateConverter

lukaszlenart:
Implements [WW-4799](https://issues.apache.org/jira/browse/WW-4799)


lukaszlenart:
LGTM 👍

PR #469, opened 2021-01-21
WW-5115 Reduce logging for DMI excluded parameters.

gregh3269:
There are unnecessary log warning when DMI is enabled, from the ParametersInterceptor.

WARN com.opensymphony.xwork2.interceptor.ParametersInterceptor ParametersInterceptor:isAccepted - Parameter [action:myAction!save] didn't match accepted pattern [[\w+((\.\w+)|(\[\d+])|((\d+))|(['(\w|[\u4e00-\u9fa5])'])|(('(\w|[\u4e00-\u9fa5])')))*]]! See Accepted / Excluded patterns at https://struts.apache.org/security/#accepted--excluded-patterns

eg the property 'action:myAction!save' should not be considered as a bean/property parameter, as its used as part of DMI to submit the form.

Any property which matches the DMI method invocation "^(action|method):.*" needs to be silently ignored and not logged in devMode=true.

DMI_AWARE_ACCEPTED_PATTERNS can also be dropped from DefaultAcceptedPatternsChecker as the DMI action|method would never be a form property.


core/src/main/java/com/opensymphony/xwork2/security/DefaultExcludedPatternsChecker.java
```diff
@@ -84,6 +85,14 @@ protected void setDynamicMethodInvocation(String dmiValue) {
...
             LOG.debug("DMI is disabled, adding DMI related excluded patterns");
             setAdditionalExcludePatterns("^(action|method):.*");
+        } else {
+            LOG.debug("DMI is enabled, adding DMI related ignored patterns");
+            ignoredPatterns = new HashSet<>();
+            try {
+                ignoredPatterns.add(Pattern.compile("^(action|method):.*", Pattern.CASE_INSENSITIVE));
+            } finally {
+            	ignoredPatterns = Collections.unmodifiableSet(ignoredPatterns);
+            }
```
lukaszlenart, line 95: Basically right now it doesn't matter if DMI is enabled or not, `action|method` will be excluded. Wouldn't be better to just extend the base RegEx? Or setup this pattern in constructor?

```java
public DefaultExcludedPatternsChecker() {
    setExcludedPatterns(EXCLUDED_PATTERNS);
    setAdditionalExcludePatterns("^(action|method):.*");
}
```

So then additional `ignoredPatterns` field is not needed.

core/src/main/java/com/opensymphony/xwork2/security/ExcludedPatternsChecker.java
```diff
@@ -66,18 +66,24 @@
 
         private final boolean excluded;
         private final String excludedPattern;
+        private final boolean log;
```
lukaszlenart, line 69: I'm not a fun of such flags as they are quite misleading and control something out of scope of this class.

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -324,12 +324,12 @@ protected boolean isAccepted(String paramName) {
     protected boolean isExcluded(String paramName) {
         ExcludedPatternsChecker.IsExcluded result = excludedPatterns.isExcluded(paramName);
         if (result.isExcluded()) {
-            if (devMode) { // warn only when in devMode
+            if (devMode && result.isLog()) { // warn only when in devMode and required
```
lukaszlenart, line 327: What's the reason of ignoring some patterns and some not in logging?

yasserzamani:
I think your DMI isn't enabled at all because I see that DefaultAcceptedPatternsChecker setAcceptedPatterns(DMI_AWARE_ACCEPTED_PATTERNS) when DMI is enabled but in same time I see that in this PR description, the logged accepted pattern isn't DMI_AWARE_ACCEPTED_PATTERNS. It is ACCEPTED_PATTERNS which starts with \w+((\., provided DMI_AWARE_ACCEPTED_PATTERNS starts with \w+([:].

Otherwise (if it's enabled really) it should work as per tested [testDmiIsEnabled](https://github.com/apache/struts/blob/09f969a9bebe31370df64702a61420f14ead6271/core/src/test/java/com/opensymphony/xwork2/security/DefaultAcceptedPatternsCheckerTest.java#L205).

lukaszlenart:
I'm closing this PR as it changes nothing.

PR #475, opened 2021-03-06
[WW-5117] Reorders stack

lukaszlenart:
This PR reorders objects on `ValueStack` and puts action always on top just before tag. In such case any `%{..}` will be evaluated against action first and then against the tag itself.

Fixes [WW-5117](https://issues.apache.org/jira/browse/WW-5117)


core/src/test/java/org/apache/struts2/views/jsp/ui/HiddenTest.java
```diff
@@ -62,6 +62,23 @@ public void testDisabled() throws Exception {
...
+        tag.setPageContext(pageContext);
+        tag.setId("einszwei");
+        tag.setName("first");
+        tag.setValue("%{id}");
+        tag.setDynamicAttribute("", "data-wuffmiauww", "%{id}");
+
+        tag.doStartTag();
+        tag.doEndTag();
+
+        verify(TextFieldTag.class.getResource("Hidden-3.txt"));
```
yasserzamani, line 79: :clap: maybe also `assertTrue(valueStack.pop()===testAction)` and then `assertFalse(valueStack.pop()===tag)` to explicitly verify whole logic

yasserzamani (approved):

core/src/test/java/org/apache/struts2/views/jsp/ui/HiddenTest.java
lukaszlenart: I would like to have a better API to always search action first, but that requires a bit of refactoring

yasserzamani: So did you mean to merge this one for now?:)

lukaszlenart: I added an additional test to be sure I didn't break anything else

yasserzamani: Excuse me I can't remember exactly, is this situation possible: local variable action != null and an action is already on top of stack? If so then currently it ends up in a ValueStack which has duplicate actions in stack.

lukaszlenart: Not sure if I understand your question. Action is pushed on stack in `AbstractTagTest#createMocks` which happens on `setUp`, this isn't a full blown action execution loop, rather tag tests mimic the whole execution logic.
Secondly, in `FreemarkerTemplateEngine` which is responisble for rendering tags (and takes part in the test logic) there is a check

```
Object action = (ai == null) ? null : ai.getAction();
```

and I have no idea if this was because of some possible solution or just to support testing - in normal situation `ActionInvocation` cannot be null and we are triggering an exception in such situation (not sure if this was already implemented in Struts 2.5.x).

I was trying to re-configure this tag test to use proper action execution logic, but somehow I was ending up with two instances of the same action on the stack, after spending a few hours I gave up and added the above assertions.

core/src/main/java/org/apache/struts2/components/template/FreemarkerTemplateEngine.java
```diff
@@ -144,15 +144,23 @@ public void close() throws IOException {
...
+        }
         try {
             stack.push(templateContext.getTag());
+            stack.push(action);
             template.process(model, writer);
         } finally {
-            stack.pop();
+            stack.pop(); // removes action
+            stack.pop(); // removes tag
+            stack.push(action); // puts back action
```
yasserzamani, line 159: I meant this part. This lines only makes sense when `action` had been null previously. Otherwise I think it'll end up with a ValueStack which has duplicate actions in stack from a same instance.

core/src/test/java/org/apache/struts2/views/jsp/ui/HiddenTest.java
yasserzamani: Thanks a lot for clarification! I resolved this one as I've been commenting on exact place.

core/src/main/java/org/apache/struts2/components/template/FreemarkerTemplateEngine.java
lukaszlenart: Good catch, fixed!

core/src/main/java/org/apache/struts2/components/template/FreemarkerTemplateEngine.java
```diff
@@ -144,15 +144,20 @@ public void close() throws IOException {
             }
         };
 
+        LOG.debug("Puts action on the top of ValueStack, just before the tag");
+        action = stack.pop();
         try {
             stack.push(templateContext.getTag());
+            stack.push(action);
```
yasserzamani, line 151: I realized right now that I think these two lines must be outside of the try after line 148.

lukaszlenart: Could you elaborate a bit more?

yasserzamani: Yes sure thanks. I meant the `finally` block is here to undo those two pushes so consequently those two pushes must be done outside of the `try...finally`'s `try`. For example currently consider `stack.push(templateContext.getTag());` fails (it's unlikely but anyway just to describe) then the `finally` block pops three objects while we've only pushed two objects.

lukaszlenart: Hm.. right, unlikely to happen but let's be safe, fixed!

lukaszlenart:
Please squash this PR if so

core/src/main/java/org/apache/struts2/components/template/FreemarkerTemplateEngine.java
```diff
@@ -121,6 +121,10 @@ public void renderTemplate(TemplateRenderingContext templateContext) throws Exce
         ActionInvocation ai = ActionContext.getContext().getActionInvocation();
 
         Object action = (ai == null) ? null : ai.getAction();
+        if (action == null) {
+            LOG.warn("Rendering tag {} out of Action scope, accessing directly JSPs is not recommended! " +
+                    "Please read https://struts.apache.org/security/#never-expose-jsp-files-directly", templateName);
```
yasserzamani, line 126: :clap:

PR #476, opened 2021-03-15
[WW-5119] Fix: remove contention during localized text lookup (JDK 1.8+)

davoustp:
The localized resources lookup shows high contention during significant stress in an enterprise application delivering the UI using Struts 2.5.20.

Profiling the app while under load stress using the excellent https://github.com/jvm-profiling-tools/async-profiler (using the `lock` mode and `--reverse` option to aggregate on the various contention code locations) shows the following picture:

image

No other high contention code paths was detected (profiler obviously shows high lock wait times on EPoll and other `java.util.concurrent`-related waits, which are irrelevant here): this code path is the main contention culprit.

Switching to the concurrent version of `Set` fixes entirely the issue: no more contention detected in this area after applying the fix.

Fixes [WW-5119](https://issues.apache.org/jira/browse/WW-5119)

PR #477, opened 2021-03-15
[WW-5119] Fix: remove contention during localized text lookup (JDK 1.7+)

davoustp:
Backport of pull request #476 for Struts 2.5.x (JDK 1.7 still supported in this branch).

The `ConcurrentHashMap.newKeySet()` usage (available only in JDK 1.8+) is replaced by a `ConcurrentHashMap<String,Boolean>` to guarantee JDK 1.7 compatibility.

Fixes [WW-5119](https://issues.apache.org/jira/browse/WW-5119)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
@@ -31,8 +31,6 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.text.MessageFormat;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -61,7 +59,7 @@ abstract class AbstractLocalizedTextProvider implements LocalizedTextProvider {
 
     private final ConcurrentMap<MessageFormatKey, MessageFormat> messageFormats = new ConcurrentHashMap<>();
     private final ConcurrentMap<Integer, List<String>> classLoaderMap = new ConcurrentHashMap<>();
-    private final Set<String> missingBundles = Collections.synchronizedSet(new HashSet<String>());
+    private final ConcurrentMap<String,Boolean> missingBundles = new ConcurrentHashMap<>();
     private final ConcurrentMap<Integer, ClassLoader> delegatedClassLoaderMap = new ConcurrentHashMap<>();
 
     /**
@@ -390,7 +388,7 @@ abstract class AbstractLocalizedTextProvider implements LocalizedTextProvider {
         ClassLoader classLoader = getCurrentThreadContextClassLoader();
         String key = createMissesKey(String.valueOf(classLoader.hashCode()), aBundleName, locale);
 
-        if (missingBundles.contains(key)) {
+        if (missingBundles.containsKey(key)) {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
```diff
@@ -61,7 +59,7 @@
 
     private final ConcurrentMap<MessageFormatKey, MessageFormat> messageFormats = new ConcurrentHashMap<>();
     private final ConcurrentMap<Integer, List<String>> classLoaderMap = new ConcurrentHashMap<>();
-    private final Set<String> missingBundles = Collections.synchronizedSet(new HashSet<String>());
+    private final ConcurrentMap<String,Boolean> missingBundles = new ConcurrentHashMap<>();
```
JCgH4164838Gh792C124B5, line 62: Minor formatting suggestion:  Add a space before `Boolean` for: `ConcurrentMap<String, Boolean>`.

core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
```diff
@@ -413,11 +411,11 @@ public ResourceBundle findResourceBundle(String aBundleName, Locale locale) {
                     }
                 } catch (MissingResourceException e) {
                     LOG.debug("Missing resource bundle [{}]!", aBundleName, e);
-                    missingBundles.add(key);
+                    missingBundles.put(key, Boolean.TRUE);
```
JCgH4164838Gh792C124B5, line 414: Consider changing `put()` to `putIfAbsent()` as the closest behaviour match to the previous `add()` method usage:
`missingBundles.putIfAbsent(key, Boolean.TRUE);`

core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
```diff
@@ -413,11 +411,11 @@ public ResourceBundle findResourceBundle(String aBundleName, Locale locale) {
...
                 }
             } else {
                 LOG.debug("Missing resource bundle [{}]!", aBundleName);
-                missingBundles.add(key);
+                missingBundles.put(key, Boolean.TRUE);
```
JCgH4164838Gh792C124B5, line 418: And the same `missingBundles.putIfAbsent(key, Boolean.TRUE);` consideration here.

JCgH4164838Gh792C124B5 (changes requested):
Hi @davoustp .  Good work on locating a concurrency bottleneck and thanks for providing a PR that attempts to address it.  👏 

There does not appear to be any obvious issue with moving away from a fully synchronized collection for the missing bundles (that I can see for the current code).

I have provided some minor suggestions for this PR, but beyond that it looks good to me.  👍 
The PR #476 equivalent change (for JDK 8+) was able maintain the same type, so the comments in this review are not relevant to it.

core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
davoustp: Not sure that it makes any difference, but let's stick to the closest behaviour as mentioned.

core/src/main/java/com/opensymphony/xwork2/util/AbstractLocalizedTextProvider.java
davoustp: Same as above.

davoustp (commented):
Done - commited.

lukaszlenart:
nice 👏 
LGTM 👍

PR #478, opened 2021-03-22
[WW-5121] Fix: remove contention during Scope.SINGLETON injection

davoustp:
Fixes the contention issue described in WW-5121 by applying a double-null-check-with-volatile pattern to avoid the `synchronized` cost on each and every injection of the singleton.

This pull request applies to branch master (Struts 2.6.x).


davoustp:
Also available for Struts 2.5.x with #479

lukaszlenart:
Looks good to me 💯

yasserzamani:
Awesome :100: thanks!

PR #479, opened 2021-03-22
[WW-5121] Fix: remove contention during Scope.SINGLETON injection

davoustp:
Fixes the contention issue described in WW-5121 by applying a double-null-check-with-volatile pattern to avoid the synchronized cost on each and every injection of the singleton.

Same patch as #478 but against the Struts 2.5.x branch.


yasserzamani:
Awesome :clap: thanks a lot!

PR #484, opened 2021-04-25
[WW-5125] use .get(...) instead of [...] for fieldErrors

yasserzamani:
All [ occurrences in all ftl files reviewed. Only fieldErrors had used [...] which replaced by .get(...) to fix WW-5125 as well as aligning with other ftl files conduct.


lukaszlenart:
LGTM 👍

PR #486, opened 2021-05-02
[WW-5126] use == instead of .equals in ModelDrivenInterceptor

yasserzamani:
due to refreshing model regardless of potential overridden Object.equals


lukaszlenart (approved):
LGTM 👍

PR #492, opened 2021-06-27
[WW-5132] [WW-5123] labelposition - cherry-pick

lukaszlenart:
Cherry pick of https://github.com/apache/struts/pull/490 plus some additional small changes

Refs [WW-5123](https://issues.apache.org/jira/browse/WW-5123)
Refs [WW-5132](https://issues.apache.org/jira/browse/WW-5132)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/UIBean.java
+++ b/core/src/main/java/org/apache/struts2/components/UIBean.java
@@ -645,7 +645,7 @@ public abstract class UIBean extends Component {
     public void evaluateParams() {
         String templateDir = getTemplateDir();
         String theme = getTheme();
-        
+
         addParameter("templateDir", templateDir);
         addParameter("theme", theme);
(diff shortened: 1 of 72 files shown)


core/src/main/java/org/apache/struts2/components/UIBean.java
```diff
@@ -689,7 +689,9 @@ public void evaluateParams() {
         }
 
         if (labelPosition != null) {
-            addParameter("labelposition", findString(labelPosition));
+            String labelPosition = findString(this.labelPosition);
+            addParameter("labelposition", labelPosition);
```
yasserzamani, line 693: Not sure but I think not required anymore as all usages in all ftl files are removed.

lukaszlenart: Yet someone can depend on that in his custom theme.

PR #493, opened 2021-06-27
[WW-5136] Makes class attribute deprecated

lukaszlenart:
Fixes [WW-5136](https://issues.apache.org/jira/browse/WW-5136)


yasserzamani:
LGTM :+1:

PR #494, opened 2021-07-04
[WW-5136] Makes class attribute deprecated - cherry-pick

lukaszlenart:
Refs [WW-5136](https://issues.apache.org/jira/browse/WW-5136)


yasserzamani:
thanks! LGTM :+1:

PR #495, opened 2021-07-05
[WW-5129] Supports dynamic attributes in second control

lukaszlenart:
Implements [WW-5129](https://issues.apache.org/jira/browse/WW-5129)

### DIFF
--- a/core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
+++ b/core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
@@ -18,27 +18,19 @@
  * under the License.
  */
 -->
+<#macro prefixedDynamicAttributes prefix>
 <#if (parameters.dynamicAttributes?? && parameters.dynamicAttributes?size > 0)><#rt/>
 <#assign aKeys = parameters.dynamicAttributes.keySet()><#rt/>
 <#list aKeys as aKey><#rt/>
-<#if dynamic_attributes_ignore??>
-<#if !aKey?starts_with(dynamic_attributes_ignore)>
+<#if aKey?starts_with(prefix)>
 <#assign keyValue = parameters.dynamicAttributes.get(aKey)/>
 <#if keyValue?is_string>
     <#assign value = struts.translateVariables(keyValue)!keyValue/>
 <#else>
     <#assign value = keyValue?string/>
 </#if>
- ${aKey}="${value?html}"<#rt/>
-</#if>
-<#else>
-<#assign keyValue = parameters.dynamicAttributes.get(aKey)/>
-<#if keyValue?is_string>
-    <#assign value = struts.translateVariables(keyValue)!keyValue/>
-<#else>
-    <#assign value = keyValue?string/>
-</#if>
- ${aKey}="${value?html}"<#rt/>
+ ${aKey}="${value}"<#rt/>
 </#if>
 </#list><#rt/>
 </#if><#rt/>
+</#macro>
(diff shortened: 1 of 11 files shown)


core/src/main/resources/template/simple/dynamic-attributes.ftl
```diff
@@ -21,12 +21,24 @@
 <#if (parameters.dynamicAttributes?? && parameters.dynamicAttributes?size > 0)><#rt/>
 <#assign aKeys = parameters.dynamicAttributes.keySet()><#rt/>
 <#list aKeys as aKey><#rt/>
-  <#assign keyValue = parameters.dynamicAttributes.get(aKey)/>
```
yasserzamani, line 24: I think it wasn't required to delete these lines and add following duplicate new lines. Adding just this new line before here is more brief and should work: `<#if dynamic_attributes_ignore?? && aKey?starts_with(dynamic_attributes_ignore)><#continue></#if>`

core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
```diff
@@ -0,0 +1,36 @@
...
+<#assign aKeys = parameters.dynamicAttributes.keySet()><#rt/>
+<#list aKeys as aKey><#rt/>
+<#if aKey?starts_with(prefix)>
+<#assign keyValue = parameters.dynamicAttributes.get(aKey)/>
+<#if keyValue?is_string>
+    <#assign value = struts.translateVariables(keyValue)!keyValue/>
+<#else>
+    <#assign value = keyValue?string/>
+</#if>
+ ${aKey}="${value}"<#rt/>
```
yasserzamani, line 32: I think we should strip the prefix from `aKey` before printing to output e.g. `second-foo: bar` should produce `< ... foo="bar"` not `< ... second-foo="bar"` because we've used `second-` internally as a flag to address issue.

yasserzamani (changes requested):

core/src/main/resources/template/simple/dynamic-attributes.ftl
lukaszlenart: As far I see `<#continue>` is [deprecated](https://freemarker.apache.org/docs/ref_directive_list.html#ref_list_continue)

core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
lukaszlenart: Hmm... maybe I should add an option `stripPrefix` and allow users control the behaviour.

core/src/main/resources/template/simple/dynamic-attributes.ftl
yasserzamani: I see thanks. But still that link's note proposes other solutions which fits here well, e.g. `?filter(`.

core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
yasserzamani: IMHO it's our internal flag. I don't think that user might be interested in it at all.

core/src/main/resources/template/simple/dynamic-attributes.ftl
lukaszlenart: Done, I had to use `<#function>`

core/src/main/resources/template/simple/prefixed-dynamic-attributes.ftl
lukaszlenart: Done

PR #497, opened 2021-09-22
[WW-5073] - simplify AbstractMatcher convertParam

sdutry:
[WW-5073](https://issues.apache.org/jira/browse/WW-5073)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
@@ -19,12 +19,19 @@
 package com.opensymphony.xwork2.config.impl;
 
 import com.opensymphony.xwork2.util.PatternMatcher;
+
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.Serializable;
-import java.util.*;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
```diff
@@ -204,31 +211,22 @@ public E match(String potentialMatch) {
...
-        }
-        
-        return ret.toString();
-    }
+	protected String convertParam(String val, Map<String, String> vars) {
+		if (val == null) {
+			return null;
+		}
+
+		Pattern wildcardPattern = Pattern.compile("\\{(.)\\}");
```
lukaszlenart, line 219: Just a small optimisation, could you move this pattern compilation into a field? Right now each time the `convertParam` is called, the same pattern is compiled over and over.

lukaszlenart (changes requested):

core/src/main/java/com/opensymphony/xwork2/config/impl/AbstractMatcher.java
sdutry: Moved.

lukaszlenart: 👏

lukaszlenart (approved):
LGTM 👍

PR #503, opened 2021-11-10
Cache all OGNL expressions (when enabled) irrespective of their execution status [Fixes WW-5147]

davoustp:


lukaszlenart:
As I already posted in #504, I would keep this change in 2.6.x branch only and avoid changing behaviour in 2.5.x

PR #505, opened 2021-11-11
Add test codes to check the encoding of response

powercomsmt:
Add test codes to check the encoding of response


lukaszlenart (approved):
LGTM 👍

PR #506, opened 2021-11-13
Fix flaky test testDynamicAttributesSupport.

yyfMichaelYan:
Fixed the flaky test testDynamicAttributesSupport due to non-deterministic order of two attributes.

This flaky test is found by using [Nondex](https://github.com/TestingResearchIllinois/NonDex), a tool used to find flaky tests in Java projects. In the first two elements of the expected string, the order of the two attributes, ```foo``` and ```placeholder```, is non-deterministic, which means ```foo``` can be either before or after ```placeholder```. To make the test pass every time, all 4 possible orders are covered in the assertion.

### DIFF
--- a/core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
+++ b/core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
@@ -113,15 +113,16 @@ public class FreemarkerResultMockedTest extends StrutsInternalTestCase {
         ActionMapping mapping = container.getInstance(ActionMapper.class).getMapping(request, configurationManager);
         dispatcher.serviceAction(request, response, mapping);
 
-        String expected =
-            "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" foo=\"bar\" placeholder=\"input\"/>"
-                + "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" foo=\"bar\" placeholder=\"input\"/>"
-                + "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" break=\"true\"/>"
-                + "<input type=\"text\" name=\"required\" value=\"\" id=\"required\" required=\"true\"/>";
+        String input1 = "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" foo=\"bar\" placeholder=\"input\"/>";
+        String input2 = "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" placeholder=\"input\" foo=\"bar\"/>";
+        String input34 = "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" break=\"true\"/>"
+                       + "<input type=\"text\" name=\"required\" value=\"\" id=\"required\" required=\"true\"/>";
 
         String result = stringWriter.toString();
-
-        assertEquals(expected, result);
+        assertTrue(result.equals(input1 + input1 + input34)
+                || result.equals(input1 + input2 + input34)
+                || result.equals(input2 + input1 + input34)
+                || result.equals(input2 + input2 + input34));
     }
 
     public void testManualListInTemplate() throws Exception {
--- a/core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
+++ b/core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
@@ -38,6 +38,11 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 
 import static org.apache.struts2.views.jsp.AbstractUITagTest.normalize;
+import static org.hamcrest.CoreMatchers.allOf;
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.endsWith;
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
```diff
@@ -113,15 +113,16 @@ public void testDynamicAttributesSupport() throws Exception {
...
+        String input34 = "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" break=\"true\"/>"
+                       + "<input type=\"text\" name=\"required\" value=\"\" id=\"required\" required=\"true\"/>";
 
         String result = stringWriter.toString();
-
-        assertEquals(expected, result);
+        assertTrue(result.equals(input1 + input1 + input34)
+                || result.equals(input1 + input2 + input34)
+                || result.equals(input2 + input1 + input34)
+                || result.equals(input2 + input2 + input34));
```
lukaszlenart, line 125: It is very hard to understand why this logic was introduced, there is no explanation neither in parameter names nor in a comment to the test.

I would extract the changing parts and assert them with `.contains()` and then use `.startsWith()` and `.endsWith()` to assert the constant parts.

lukaszlenart (changes requested):

core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
sepe81: The project also includes hamcrest. I guess, that some test in the sense of https://stackoverflow.com/a/1092241/153337 would work.

sepe81 (changes requested):

core/src/test/java/org/apache/struts2/views/freemarker/FreemarkerResultMockedTest.java
yyfMichaelYan: Hi Lukasz! Thanks for your suggestion! Now, the two possible orders of ```foo``` and ```placeholder``` are covered by the two ```.contains()``` method, and the last two ```input``` elements (i.e. the constant part) is covered by the ```.endsWith()``` method.

yyfMichaelYan: Hi Sebastian! Thanks for your advice! Yes, hamcrest can work. However, I didn't use it because that would require multiple ```assertThat``` and essentially it's the same as using one assertion with ```.contains()``` and ```.endsWith()```.

yyfMichaelYan: Hi Sebastian! Actually, you can also add an error message to ```assertTrue```. The signature of this method is ```TestCase.assertTrue(String message, boolean condition) : void```. I didn't add it because the initial assertion doesn't come with one. If you think it's better to have an error message, I can definitely add it!

lukaszlenart: @sepe81 idea was something like this
```java
assertThat(result)
    .startsWith("<input type=\"text\" name=\"test\" value=\"\" id=\"test\"")
    .contains("foo=\"bar\"")
    .contains("placeholder=\"input\"")
    .endsWith(
        "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" break=\"true\"/>"
            + "<input type=\"text\" name=\"required\" value=\"\" id=\"required\" required=\"true\"/>"
    );
```

sepe81: Exactly. I was not worried about the fixed assertion message but the dynamic message on failing assertions. With @lukaszlenart's proposal, IMHO we would see the exact setting, if something is going wrong.

yyfMichaelYan: @lukaszlenart @sepe81 Thank you both! I found the method ```allOf(java.lang.Iterable<Matcher<? super T>> matchers)``` which can put all matchers together and be used as the second argument to ```assertThat```.

The dynamic messages are a little different (if my understanding of "dynamic message" is correct). With the new assertion methods, there are extra text like ```a string starting with...```.
Original:
```
expected:<<input[]type="text" name="te...> but was:<<input[ ]type="text" name="te...>
```
New:
```
Expected: (a string starting with "<inputtype=\"text\" name=\"test\" value=\"\" id=\"test\"" and a string containing "foo=\"bar\"" and a string containing "placeholder=\"input\"" and a string ending with "<input type=\"text\" name=\"test\" value=\"\" id=\"test\" break=\"true\"/><input type=\"text\" name=\"required\" value=\"\" id=\"required\" required=\"true\"/>")
     but: a string starting with "<inputtype=\"text\" name=\"test\" value=\"\" id=\"test\"" was "<input type="text" name="test" value="" id="test" placeholder="input" foo="bar"/><input type="text" name="test" value="" id="test" foo="bar" placeholder="input"/><input type="text" name="test" value="" id="test" break="true"/><input type="text" name="required" value="" id="required" required="true"/>"
```

lukaszlenart (approved):
Thanks a lot for your patience and changes, LGTM 👍

PR #509, opened 2021-11-16
[WW-5148] Upgrade ASM to version 9.2

sepe81:
with JDK 17 and 18 support according to https://asm.ow2.io/versions.html


lukaszlenart:
Thanks!

PR #510, opened 2021-11-23
[WW-5149] Puts back "labelposition" with a warning of deprecation

lukaszlenart:
Fixes [WW-5149](https://issues.apache.org/jira/browse/WW-5149)

### DIFF
--- a/core/src/main/java/org/apache/struts2/views/jsp/ui/AbstractUITag.java
+++ b/core/src/main/java/org/apache/struts2/views/jsp/ui/AbstractUITag.java
@@ -30,6 +30,7 @@ import java.util.Map;
  * Abstract base class for all UI tags.
  */
 public abstract class AbstractUITag extends ComponentTagSupport implements DynamicAttributes {
+
     protected String cssClass;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/struts2/views/jsp/ui/AbstractUITag.java
```diff
@@ -90,7 +93,11 @@ protected void populateParams() {
         uiBean.setDisabled(disabled);
         uiBean.setLabel(label);
         uiBean.setLabelSeparator(labelSeparator);
-        uiBean.setLabelPosition(labelPosition);
+        if (labelposition != null) {
```
yasserzamani, line 96: Not sure but since labelposition is deprecated, shouldn't we check for `labelPosition` here instead? to give it preference and priority higher than the deprecated one.

lukaszlenart: Hm... right :)

yasserzamani (approved):
LGTM :+1:

PR #520, opened 2021-12-28
REST Showcase fix for broken JSPs using S2 link tags

JCgH4164838Gh792C124B5:
Update:
- Fix REST Showcase JSP issue for s:link tags that replaced HTML link tags.
- s:link tags cannot use JSP EL in them.  Modified JSPs to use scriptlet, JSP expression, and OGNL expression to work with s:link tag.

### DIFF
--- a/apps/rest-showcase/src/main/webapp/WEB-INF/content/orders-deleteConfirm.jsp
+++ b/apps/rest-showcase/src/main/webapp/WEB-INF/content/orders-deleteConfirm.jsp
@@ -27,8 +27,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Orders</title>
-    <s:link href="${pageContext.request.contextPath}/css/bootstrap.min.css" rel="stylesheet"></s:link>
-    <s:link href="${pageContext.request.contextPath}/css/app.css" rel="stylesheet"></s:link>
+    <!-- Using a standard HTML link tag with JSP EL to get the contextPath may be simpler, but this is an equivalent for s:link -->
+    <s:set var="pageContextPath"><%=((HttpServletRequest)request).getContextPath()%></s:set>
+    <s:link href="%{#pageContextPath}/css/bootstrap.min.css" rel="stylesheet"></s:link>
+    <s:link href="%{#pageContextPath}/css/app.css" rel="stylesheet"></s:link>
 
     <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
     <!-- WARNING: Respond.js doesn't work if you view the page via file:// -->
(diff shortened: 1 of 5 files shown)


JCgH4164838Gh792C124B5:
Hi Apache Struts Team.

I noticed a small bug with the REST Showcase when testing S2 6.0RC1, which prevented the JSPs from rendering.  A JIRA can be opened for this PR, if it is considered OK after review.

apps/rest-showcase/src/main/webapp/WEB-INF/content/orders-deleteConfirm.jsp
```diff
@@ -27,8 +27,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Orders</title>
-    <s:link href="${pageContext.request.contextPath}/css/bootstrap.min.css" rel="stylesheet"></s:link>
-    <s:link href="${pageContext.request.contextPath}/css/app.css" rel="stylesheet"></s:link>
+    <!-- Using a standard HTML link tag with JSP EL to get the contextPath may be simpler, but this is an equivalent for s:link -->
+    <s:set var="pageContextPath"><%=((HttpServletRequest)request).getContextPath()%></s:set>
+    <s:link href="%{#pageContextPath}/css/bootstrap.min.css" rel="stylesheet"></s:link>
```
lukaszlenart, line 32: I think this should work as well:
```
<s:link href="%{#request.pageContext}/css/bootstrap.min.css" rel="stylesheet"></s:link>
```
and it's more [Struts'ish](https://struts.apache.org/tag-developers/ognl)

JCgH4164838Gh792C124B5: Hi @lukaszlenart .

Thanks for the feedback and suggestion.  :) 

I think for the above s:link tag suggestion:
`<s:link href="%{#request.pageContext}/css/bootstrap.min.css" rel="stylesheet"></s:link>`
you may have meant to suggest `contextPath` rather than `pageContext`, so something like:
`<s:link href="%{#request.contextPath}/css/bootstrap.min.css" rel="stylesheet"></s:link>`  ?

Unfortunately, neither worked (nor did similar attempts to access the `contextPath` via the value stack).  I had tried various expressions to try and get the information, but every attempt to get it from the value stack resulted in an empty string.

Since the TLD does not allow for JSP EL usage either, the only solution I could find on the JSP was to use a JSP expression.  If there was an action class to use in the example, we could make it request aware and implement a getter (a more "Struts-ish" option).

So far, the PR's solution was the only one I could find that worked with `s:link`.  If there is a cleaner solution, please let me know and I will update the PR.

lukaszlenart: Ok, right, I made a mistake. `#request` isn't `HttpServletRequest` but `RequestMap` - a wrapper around the request which only exposes attributes of the request and not the whole object.

lukaszlenart (approved):
LGTM 👍

PR #523, opened 2022-01-04
[WW-5022] Sets escapeHtmlBody to false by default and defines a new flag

lukaszlenart:
This PR turns `escapeHtmlBody` into false by default and introduces a new flag `struts.ui.escapeHtmlBody` to allow globally switch body escaping to `true` if needed.

This change popped up during discussion on the new Struts 2 6.0.0 version
https://lists.apache.org/thread/dp0zdtn17pj64gv8rdc8sjs1f6q4rd8z

Refs [WW-5022](https://issues.apache.org/jira/browse/WW-5022)

### DIFF
--- a/apps/showcase/src/main/webapp/WEB-INF/decorators/main.jsp
+++ b/apps/showcase/src/main/webapp/WEB-INF/decorators/main.jsp
@@ -113,7 +113,7 @@
 
         <div id="navbar" class="navbar-collapse collapse">
                 <ul class="nav navbar-nav">
-                    <li><s:a value="%{home}" escapeHtmlBody="false"><i class="glyphicon glyphicon-home"></i> Home</s:a></li>
+                    <li><s:a value="%{home}"><i class="glyphicon glyphicon-home"></i> Home</s:a></li>
                     <li class="dropdown">
                         <a href="#" class="dropdown-toggle" data-toggle="dropdown">
                             <i class="glyphicon glyphicon-cog"></i> Configuration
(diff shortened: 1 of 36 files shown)


gregh3269:
For the java templates plugin to work, possibly need to add the escapeHtmlBody parameter to the evaluateExtraParams() of the anchor and submit tag, otherwise its not passed to the plugin.  

org.apache.struts2.components.Anchor

@Override
    protected void evaluateExtraParams() {
        super.evaluateExtraParams();

        if (href != null) {
            addParameter("href", ensureAttributeSafelyNotEscaped(findString(href)));
        } else {
            //no href, build it from URL attributes
            StringWriter sw = new StringWriter();
            urlRenderer.beforeRenderUrl(urlProvider);
            urlRenderer.renderUrl(sw, urlProvider);
            String builtHref = sw.toString();
            if (StringUtils.isNotEmpty(builtHref)) {
                addParameter("href", ensureAttributeSafelyNotEscaped(builtHref));
            }
        }
        **addParameter("escapeHtmlBody", escapeHtmlBody)**;
    }

Maybe there is an alternative way, as escapeHtmlBody are not part of UIBean.

lukaszlenart:
I wonder if we need that in Java Templates, `escapeHtmlBody` is only needed by Freemarker based tags to avoid auto-escaping, which is default in the latest version of Freemarker. Java Templates use they own escaping mechanism which didn't change with introducing the latest Freemarker.

gregh3269:
Possibly does not do any escaping, as encode is always false.
 
Looking at the code it does :

org.apache.struts2.views.java.XHTMLTagSerializer
public void characters(String text, boolean encode) throws IOException {
        writer.write(encode ? StringUtils.defaultString(StringEscapeUtils.escapeHtml4(text)) : text);
 }

org.apache.jasper.runtime.JspWriterImpl.write(String s, int off, int len) throws IOException

lukaszlenart:
@gregh3269 thanks, I added the flag to JavaTemplates

lukaszlenart:
Time to merge :)

PR #524, opened 2022-01-04
[WW-5117] Evaluates dynamic attributes

lukaszlenart:
This PR introduces conditional lazy evaluation of some tag's attributes (e.g.: _dynamic attributes_) to avoid double evaluation vulnerability.

Reverts changes in #475 and fixes [WW-5117](https://issues.apache.org/jira/browse/WW-5117)

PR #525, opened 2022-01-13
[WW-5117] Evaluates dynamic attributes - cherrypick

lukaszlenart:
Cherry picks #524

### DIFF
--- a/core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
+++ b/core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
@@ -346,7 +346,8 @@ public class TextfieldTest extends AbstractUITagTest {
         tag.setName("myname");
         tag.setValue("%{foo}");
         tag.setSize("10");
-        tag.setDynamicAttribute(null, "anotherAttr", "another_%{foo}");
+        tag.setDynamicAttribute(null, "anotherAttr", "%{foo}");
+        tag.setDynamicAttribute(null, "secondAttr", "second_%{foo}");
 
         tag.doStartTag();
         tag.doEndTag();
@@ -372,7 +373,8 @@ public class TextfieldTest extends AbstractUITagTest {
         tag.setName("myname");
         tag.setValue("%{foo}");
         tag.setSize("10");
-        tag.setDynamicAttribute(null, "anotherAttr", "another_%{foo}");
+        tag.setDynamicAttribute(null, "anotherAttr", "%{foo}");
+        tag.setDynamicAttribute(null, "secondAttr", "second_%{foo}");
 
         tag.doStartTag();
         setComponentTagClearTagState(tag, true);  // Ensure component tag state clearing is set true (to match tag).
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
```diff
@@ -346,7 +346,7 @@ public void testSimple_recursionTest() throws Exception {
         tag.setName("myname");
         tag.setValue("%{foo}");
         tag.setSize("10");
-        tag.setDynamicAttribute(null, "anotherAttr", "%{foo}");
+        tag.setDynamicAttribute(null, "anotherAttr", "another_%{foo}");
```
yasserzamani, line 349: That's intentional by me to start value with %{ to always have a monitor and test for double evaluation. This change eliminate that I think. Why had you to change it? to pass that test?

core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
```diff
@@ -372,7 +372,7 @@ public void testSimple_recursionTest_clearTagStateSet() throws Exception {
...
```
yasserzamani, line 375: (same here)

core/src/test/resources/org/apache/struts2/views/jsp/ui/Textfield-5.txt
```diff
@@ -1,4 +1,4 @@
 <tr>
     <td class="tdLabel"><label for="myname" class="label">mylabel:</label></td>
-    <td class="tdInput"><input type="text" name="myname" size="10" value="%{1+1}" id="myname" anotherAttr="%{1+1}"/></td>
+    <td class="tdInput"><input type="text" name="myname" size="10" value="%{1+1}" id="myname" anotherAttr="another_%{1+1}"/></td>
```
yasserzamani, line 3: (same here)

yasserzamani (changes requested):

core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
yasserzamani: Thanks for clarification! Just wanted to note and clarify it. So we can go ahead :) 

p.s. however I think it's a good practice to not touch current tests and just add tests, just for more confident that changes aren't breaking previous assertions. e.g. in this case I think we can simply add a new attr **anotherAttr2**:
```tag.setDynamicAttribute(null, "anotherAttr2", "another_%{foo}");```
and add expected attr to the current expected file. I see your point that dynAttr are normally `static_text_%{variable}` but this test is just a special not normal case test which assumes that attacker had been successful to inject such value.

yasserzamani (approved):

core/src/test/java/org/apache/struts2/views/jsp/ui/TextfieldTest.java
lukaszlenart: Sure, I can do it, no problem :)

lukaszlenart: Done!

yasserzamani: thanks so much! LGTM :+1: will wait a few days and merge

PR #482, opened 2021-04-03
[WW-5115] Reduce logging of ignored DMI params

lukaszlenart:
@gregh3269 what about such simple solution?

Resolves [WW-5115](https://issues.apache.org/jira/browse/WW-5115)

PR #526, opened 2022-01-23
[WW-5115] Reduces logging for ignored DMI related params when DMI is disabled

lukaszlenart:
@gregh3269 what about such simple solution? I have cherry-picked commits from https://github.com/apache/struts/pull/482

Resolves [WW-5115](https://issues.apache.org/jira/browse/WW-5115)


lukaszlenart:
LGTM 👍

PR #529, opened 2022-02-06
[WW-5016] Adds support for LocalDate and adjusts tests to use the new Java 8 API

lukaszlenart:
Relates to [WW-5016](https://issues.apache.org/jira/browse/WW-5016)


JCgH4164838Gh792C124B5 (commented):
Hi @lukaszlenart.
The changes and updates to the unit tests seem to make sense, based on a migration from `SimpleDateFormat` to `DateTimeFormatter`.  👍 
The only other consideration might be whether there would be any (or enough) value in providing a backwards-compatibility mode for the old date formatting ?
Probably not, but just raising it for consideration.

lukaszlenart:
Hm... I can implement a flag to either use a new or an old formatting mechanism. Yet this will introduce unnecessary complexity and I think it's a good moment to break backward compatibility while releasing a new major version.

lukaszlenart:
I'm working on a version which uses `@Inject` to inject a porper formatter.

lukaszlenart:
Done, I've extended the change to use to a formatter injected by the framework based on user choice.

PR #530, opened 2022-02-07
Add submitUnchecked parameter

gregh3269:
Add missing attribute submitUnchecked to java templates checkbox tag so it matches the main freemarker template.  See WW-5168.


gregh3269:
think I got this wrong :-(.  end("input");  Will try again.

gregh3269:
OK #531

PR #532, opened 2022-03-28
Basic LocalDateTime support

gregh3269:
Add support for LocalDateTime for the LocalDateTime.now() variant. DateTimeFormatter.ISO_LOCAL_DATE_TIME.

See [WW-5175](https://issues.apache.org/jira/browse/WW-5175)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
+++ b/core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
@@ -18,28 +18,34 @@
  */
 package com.opensymphony.xwork2.conversion.impl;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.apache.struts2.conversion.TypeConversionException;
-import com.opensymphony.xwork2.ActionContext;
-import com.opensymphony.xwork2.TextProvider;
-import com.opensymphony.xwork2.util.ValueStack;
-
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
 import java.text.DateFormat;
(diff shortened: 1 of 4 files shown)


lukaszlenart:
What's the status here? Do we still have a problem with `struts.date.format`?

gregh3269:
Don't think so.  I did another patch (#3) to remove the !"".equals(globalFormat) check, and the globalDateFormat lookup on the LocalDateTime as it will break it.

lukaszlenart:
Ok, so only tests need to be fixed

gregh3269:
ok, I will have a look.

core/src/main/java/com/opensymphony/xwork2/conversion/impl/DateConverter.java
```diff
@@ -156,20 +192,37 @@ private DateFormat[] getDateFormats(ActionContext context, Locale locale) {
...
 
+    /**
+     * Retrieves the list of date time formats to be used when converting dates
+     * 
+     * @param context the current ActionContext
+     * @param locale  the current locale of the action
+     * 
+     * @return a list of DateTimeFormatter to be used for date conversion
+     */
+    private DateTimeFormatter[] getDateTimeFormats(ActionContext context, Locale locale) {
```
lukaszlenart, line 217: Can this be `protected` to allow override?

lukaszlenart (commented):
Looks good, just one suggestion

lukaszlenart:
LGTM 👍

gregh3269:
...maybe the formatting should be like java.sql.Timestamp, in line, as it only parses with a yyyy-mm-ddThh:mm:ss.  Early days here as its a big topic.

PR #533, opened 2022-04-03
fix WW-5165 Update spring to 5.3.x b/c 4.3.x is EOL

victorsosa:
fix for issue [WW-5165](https://issues.apache.org/jira/browse/WW-5165)

Update spring to 5.3.x b/c 4.3.x is EOL


pom.xml
```diff
@@ -103,6 +103,20 @@
...
 
     <properties>
+        <currentVersion>${project.version}</currentVersion>
+        <struts2.springPlatformVersion>5.3.18.RELEASE</struts2.springPlatformVersion>
+        <ognl.version>3.1.8</ognl.version>
+        <asm.version>3.3</asm.version>
+        <asm5.version>5.0.2</asm5.version>
+        <tiles.version>3.0.5</tiles.version>
+        <tiles-request.version>1.0.6</tiles-request.version>
+        <log4j2.version>2.5</log4j2.version>
```
lukaszlenart, line 113: You have duplicated these properties and some of them have been reverted to very old versions. Also this changes nothing as the same properties are defined a few lines below which means they will overwrite your changes - that's why all the tests have passed.

lukaszlenart:
I would close this PR as it was branched off old master, you can follow this step to learn how to prepare a fresh branch out of Struts master branch - it's a recommended approach.

https://struts.apache.org/submitting-patches.html#contributing-with-github

PR #537, opened 2022-04-22
safety update with EXCLUDED_PATTERNS

k4n5ha0:
to avoid attack pocs like st2-061 st2-062 to clean excludedPackageNamePatterns excludedClasses and excludedPackageNames


k4n5ha0:
i chose other way to protect my app

PR #542, opened 2022-05-01
Update SecurityMemberAccess.java fix  security bugs

k4n5ha0:
fix bugs
A.isAssignableFrom(B) this judge if B belong A
deb0b13e394317ab49e132004abcb576_166144853-f70e78d4-a359-4082-bd86-34c10e8d99f2

if not fix it will always return  false


lukaszlenart:
@k4n5ha0 did you ever run tests? You will notice tones of WARNs and it won't work.

PR #546, opened 2022-05-03
update isClassExcluded logic

k4n5ha0:
1) all test pass

2)i send to jira 
https://issues.apache.org/jira/browse/WW-5180

3)read this paper below 
relink:https://mc0wn.blogspot.com/2021/04/exploiting-struts-rce-on-2526.html
we can know if we use excludedClass.isAssignableFrom(clazz) to judge class is dangerous or not
st2's excludedClass list will very hard to be bypassed. 
so any other exp like paper's #@org.apache.commons.collections.BeanMap@{} to clean excludedPackageNames and excludedClasses will be block（so i add java.lang.Cloneable）

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -45,6 +45,7 @@ public class SecurityMemberAccess extends DefaultMemberAccess {
     private Set<Pattern> excludedPackageNamePatterns = Collections.emptySet();
     private Set<String> excludedPackageNames = Collections.emptySet();
     private boolean disallowProxyMemberAccess;
+    protected String packageName;
 
     /**
      * SecurityMemberAccess
@@ -55,6 +56,7 @@ public class SecurityMemberAccess extends DefaultMemberAccess {
      */
     public SecurityMemberAccess(boolean allowStaticMethodAccess) {
         super(false);
+        packageName = this.getClass().getPackage().getName();
         this.allowStaticMethodAccess = allowStaticMethodAccess;
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -168,8 +170,20 @@ protected boolean isClassExcluded(Class<?> clazz) {
             return true;
         }
         for (Class<?> excludedClass : excludedClasses) {
-            if (clazz.isAssignableFrom(excludedClass)) {
-                return true;
+            if(excludedClass != Object.class) {
+                if(clazz.getName().startsWith(packageName)){
```
lukaszlenart, line 174: Why do you need this check? The whole `com.opensymphony.xwork2.ognl.` package is already excluded and [package exclusion is tested before class exclusion](https://github.com/apache/struts/pull/546/files#diff-bd4f5dc22219d389a455107a004139c3fa3a94b9493946dc9b0c927f26217853L85-L99)

k4n5ha0: if i don't add this check, the unit test can't pass.

lukaszlenart: Also this change make that all the tests have passed as they are located in `com.opensymphony.xwork2.ognl.`

k4n5ha0: the old check in com.opensymphony.xwork2.ognl.   isn't change , also use  clazz.isAssignableFrom(excludedClass)
the new check use excludedClass.isAssignableFrom(clazz)

lukaszlenart: Yes, basically you adjusted the code just to satisfy tests, it will break users' applications as they do not use `com.opensymphony.xwork2.ognl.` in their code.

k4n5ha0: this change is for safety.
many other class can replace #@org.apache.commons.collections.BeanMap@{} to clean excludedPackageNames and excludedClasses.
we can't allow excludedClasses geting longer and longer.

lukaszlenart: This exact change is not for safety, it's to omit tests and it degrades our _safety net_. Could you provide a use case for you PR? I would like to implement an unit test to cover those changes.

lukaszlenart: I just copy/paste `SecurityMemberAccessTest` into a custom package (eg. `com.test`) to avoid messing with `com.opensymphony.xwork2.ognl.` package and to double check your change - it fails on `testMiddleOfInheritanceExclusion1`.

k4n5ha0: ok , i give up. =)
i am just web security guy , i don't good at coding

PR #550, opened 2022-05-14
[WW-5182] Upgrade to Servlet API 3.1

lukaszlenart:
Refs [WW-5182](https://issues.apache.org/jira/browse/WW-5182)


yasserzamani:
LGTM:+1:

PR #534, opened 2022-04-04
fix WW-5165 Update spring to 5.3.x b/c 4.3.x is EOL

victorsosa:
Refs [WW-5165](https://issues.apache.org/jira/browse/WW-5165)

PR #552, opened 2022-05-17
[WW-5165] Upgrades Spring to version 5.3.20

lukaszlenart:
Refs [WW-5165](https://issues.apache.org/jira/browse/WW-5165)
Closes #534

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -242,6 +242,16 @@
             <artifactId>spring-web</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context-support</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <dependency>
             <groupId>junit</groupId>
--- a/plugins/portlet-mocks/pom.xml
+++ b/plugins/portlet-mocks/pom.xml
@@ -51,6 +51,7 @@
     </dependencies>
 
     <properties>
+        <spring.platformVersion>4.3.30.RELEASE</spring.platformVersion>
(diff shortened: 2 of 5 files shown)


plugins/portlet/pom.xml
```diff
@@ -130,11 +130,7 @@
             <artifactId>spring-test</artifactId>
             <optional>true</optional>
         </dependency>
-        <dependency>
```
sepe81, line 133: Maybe we could add some additional infos to the commit to clarify the context

Remove unused spring-webmvc-portlet
Spring 5.x dropped support for spring-webmvc-portlet
https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x#dropped-support

core/pom.xml
```diff
@@ -242,6 +242,16 @@
             <artifactId>spring-web</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
```
sepe81, line 247: background: Add spring-context for testing with StaticApplicationContext

plugins/portlet-mocks/pom.xml
```diff
@@ -51,6 +51,7 @@
     </dependencies>
 
     <properties>
+        <spring.platformVersion>4.3.30.RELEASE</spring.platformVersion>
```
sepe81, line 54: good idea, to freeze the spring version here

sepe81 (approved):

core/pom.xml
lukaszlenart: Do you mean to use `<scope>test</scope>`?

plugins/portlet-mocks/pom.xml
lukaszlenart: It's the only way to avoid conflict of dependencies :D

core/pom.xml
sepe81: I guess, for the test suite to work, this would suffice -- but my comment was just meant as a review note.

PR #555, opened 2022-05-25
upgrade Felix m-bundle-p: reproducible builds

hboutemy:
see last releases rebuild results: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/struts/README.md


lukaszlenart:
LGTM 👍

PR #559, opened 2022-06-05
WW-5184 - Add optional parameter value check to ParametersInterceptor

brianandle:
Implements [WW-5184](https://issues.apache.org/jira/browse/WW-5184)
* Add optional parameter value check to ParametersInterceptor
* Add Unit Tests

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
@@ -359,7 +359,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
     protected boolean acceptableValue(String name, String value) {
     	boolean accepted = (value == null || value.isEmpty() || (!isParamValueExcluded(value) && isParamValueAccepted(value)));
         if (!accepted) {
-            LOG.info("Parameter [{}] was not accepted with value [{}] and will be dropped!", name, value);
+            LOG.warn("Parameter [{}] was not accepted with value [{}] and will be dropped!", name, value);
         }
         return accepted;
     }
@@ -452,7 +452,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
 			for (Pattern excludedPattern : excludedValuePatterns) {
 				if (value != null) {
 					if (excludedPattern.matcher(value).matches()) {
-						LOG.info("Parameter value [{}] matches excluded pattern [{}] and will be dropped.", value,
+						LOG.warn("Parameter value [{}] matches excluded pattern [{}] and will be dropped.", value,
 								excludedPattern);
 						return true;
 					}
@@ -475,7 +475,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
 			// acceptedValuePatterns not defined so anything is allowed
 			return true;
 		}
-		LOG.info("Parameter value [{}] did not match any acceptedValuePattern pattern and will be dropped.", value);
+		LOG.warn("Parameter value [{}] did not match any acceptedValuePattern pattern and will be dropped.", value);
 		return false;
 	}
     
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
@@ -191,7 +191,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
         for (Map.Entry<String, Parameter> entry : params.entrySet()) {
             String parameterName = entry.getKey();
             boolean isAcceptableParameter = isAcceptableParameter(parameterName, action);
-            isAcceptableParameter &= isAcceptableParameterValue(entry.getValue());
+            isAcceptableParameter &= isAcceptableParameterValue(entry.getValue(), action);
 
             if (isAcceptableParameter) {
                 acceptableParameters.put(parameterName, entry.getValue());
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -354,7 +407,86 @@ protected boolean isExcluded(String paramName) {
         }
         return false;
     }
+    
 
+    public void setAcceptedValuePatterns(String commaDelimitedPatterns) {
```
yasserzamani, line 412: BTW another methods annotated with `@Inject` as well as some other changes are required enabling user to configure these patterns via XML configuration.

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -354,7 +407,86 @@ protected boolean isExcluded(String paramName) {
...
+        	excludedValuePatterns = Collections.unmodifiableSet(excludedValuePatterns);
+        }
+    }
+    
+	protected boolean isParamValueExcluded(String value) {
+		if (hasParamValuesToExclude()) {
+			for (Pattern excludedPattern : excludedValuePatterns) {
+				if (value != null) {
+					if (excludedPattern.matcher(value).matches()) {
+						LOG.info("Parameter value [{}] matches excluded pattern [{}] and will be dropped.", value,
```
yasserzamani, line 455: same here

yasserzamani (changes requested):
Thanks! Here are some new requests and a fix for my previous comments in previous PR please...

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
brianandle: No matter what I did I couldn't get Inject to work on these, and I had assumed that's because I was using it inappropriately. The other setters that are documented by Struts, setParamNameMaxLength, setAcceptParamNames, setExcludeParams don't utilize the @Inject annotation and are injected via the following params and thus I followed suit

acceptParamNames, excludeParams, paramNameMaxLength

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
```diff
@@ -311,6 +346,24 @@ private boolean isIgnoredDMI(String name) {
...
+     * * Value is accepted
+     * 
+     * @param name - Param name (for logging)
+     * @param value - value to check
+     * @return true if accepted
+     */
+    protected boolean acceptableValue(String name, String value) {
+    	boolean accepted = (value == null || value.isEmpty() || (!isParamValueExcluded(value) && isParamValueAccepted(value)));
+        if (!accepted) {
+            LOG.info("Parameter [{}] was not accepted with value [{}] and will be dropped!", name, value);
```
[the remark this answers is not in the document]
brianandle: Changed to warn and updated pull request

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
yasserzamani: Ah I see , thanks for clarification!

yasserzamani (approved):

core/src/main/java/com/opensymphony/xwork2/interceptor/ParametersInterceptor.java
yasserzamani: Hmm I think it makes sense to have some default patterns set here however. Currently it's empty and user has to set it in all interceptors stacks as far as I understood right? If so then like what is done for parameters names, better to have some defaults patterns set here e.g. your already provided patterns in tests for %{} and ${}

brianandle: My concern would be the regression impact. I would sure hope that Struts users aren't ever expecting users to pass in params with values with %{} and ${} but that level of change seems best to include with a major (which ironically I just missed).

I do think we could prevent an entire attack vector by excluding %{} and ${} in param values but again I'd be worried about the regression impact.

If the Struts team wants to do that I wouldn't object but I wasn't going to go that far on my own.

brianandle: And yes it would need to be defined such as:

    <package name="default" extends="struts-default">
        <interceptors>
            <interceptor-stack name="myCustomStack">
                <interceptor-ref name="defaultStack">
                	       <param name="params.excludeValuePatterns">.*\$\{.*?\}.*,.*%\{.*?\}.*</param>
                </interceptor-ref>
            </interceptor-stack>
        </interceptors>
  </package>

brianandle:
@lukaszlenart, sorry just noticed the comment, so you're thinking of a ParameterValueAware similar to ParameterNameAware correct? If so I can work on that in the morning (I'm in PST timezone).

brianandle:
Added ParameterValueAware interface and unit test

lukaszlenart:
I would postpone this change and introduce it into 6.1.0 - right now there is a few bugs targeting 6.0.1 which is a priority for me

lukaszlenart:
LGTM 👍

lukaszlenart:
@brianandle could you add this new options to https://struts.apache.org/core-developers/parameters-interceptor ? Thanks in advance!

brianandle:
Sure. I'll work on that on Monday.
________________________________

brianandle:
Created https://github.com/apache/struts-site/pull/170 for the doc changes

PR #566, opened 2022-06-10
[WW-5185] Reintroduces proper implementation of Tiles resource to support wildcard matching of Tiles definitions

lukaszlenart:
Refs [WW-5185](https://issues.apache.org/jira/browse/WW-5185)

### DIFF
--- a/apps/rest-showcase/src/main/webapp/WEB-INF/web.xml
+++ b/apps/rest-showcase/src/main/webapp/WEB-INF/web.xml
@@ -19,12 +19,13 @@
  * under the License.
  */
 -->
-<web-app id="starter" version="2.5" xmlns="http://java.sun.com/xml/ns/javaee"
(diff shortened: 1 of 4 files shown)


yasserzamani:
Thanks LGTM :+1: just one thing: Does this mean that Tiles itself `URLApplicationResource` is not implemented properly to support wildcard matching? Or is it a bug in Tiles `URLApplicationResource` impl? ... if so then shouldn't it be fixed there instead of here at Struts?

lukaszlenart:
`URLApplicationResource` was intended to work with multiple localized Tiles definitions, eg.: `tiles.xml, tiles_pl.xml, tiles_fr.xml` and so one, `StrutsApplicationResource` is connected with `StrutsWildcardServletApplicationContext` to support wildcard matching, but without localisation support - it's done via [`I18N:`](https://struts.apache.org/plugins/tiles/#i18n) prefix. Probably I should clarify that in the docs.

PR #568, opened 2022-06-14
Add async and velocity plugin to bom

tobiasstadler:
Fixes [WW-5187](https://issues.apache.org/jira/browse/WW-5187)


tobiasstadler:
Thank You!

PR #569, opened 2022-06-14
[WW-5188] Replaces since 2.6 with 6.0

lukaszlenart:
Fixes [WW-5188](https://issues.apache.org/jira/browse/WW-5188)

### DIFF
--- a/core/src/test/resources/struts.xml
+++ b/core/src/test/resources/struts.xml
@@ -20,8 +20,8 @@
  */
 -->
 <!DOCTYPE struts PUBLIC
-          "-//Apache Software Foundation//DTD Struts Configuration 2.6//EN"
-          "http://struts.apache.org/dtds/struts-2.6.dtd">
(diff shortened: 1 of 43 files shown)


core/src/test/resources/struts.xml
```diff
@@ -20,8 +20,8 @@
  */
 -->
 <!DOCTYPE struts PUBLIC
-          "-//Apache Software Foundation//DTD Struts Configuration 2.6//EN"
-          "http://struts.apache.org/dtds/struts-2.6.dtd">
+        "-//Apache Software Foundation//DTD Struts Configuration 6.0//EN"
+        "struts-6.0.dtd">
```
sepe81, line 24: Isn't it possible to provide the dtd under https://struts.apache.org/dtds/ anymore?

lukaszlenart: It is, just forgot to add it here https://github.com/apache/struts-site/tree/master/source/dtds

PR #571, opened 2022-06-15
[WW-5190] Fixes StackOverflowException when dispatching request

lukaszlenart:
This PR addresses a problem with forwarding to another JSP or action, but only in case if `StrutsExecuteAndPrepareFilter` is used - when combination of `StrutsPrepareFilter` & `StrutsExecuteFilter` is used; forwarding to actions won't work - it's a large problem and will be [addressed](https://issues.apache.org/jira/browse/WW-5199) separately, this never worked tbh. 

Fixes [WW-5190](https://issues.apache.org/jira/browse/WW-5190)

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
@@ -83,6 +83,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.regex.Pattern;
@@ -612,20 +613,12 @@ public class Dispatcher {
         }
 
         try {
-            String namespace = mapping.getNamespace();
-            String name = mapping.getName();
-            String method = mapping.getMethod();
+            String actionNamespace = mapping.getNamespace();
+            String actionName = mapping.getName();
+            String actionMethod = mapping.getMethod();
 
-            ActionProxy proxy;
-
-            //check if we are probably in an async resuming
-            ActionInvocation invocation = ActionContext.getContext().getActionInvocation();
-            if (invocation == null || invocation.isExecuted()) {
-                proxy = getContainer().getInstance(ActionProxyFactory.class).createActionProxy(namespace, name, method,
-                    extraContext, true, false);
-            } else {
-                proxy = invocation.getProxy();
-            }
+            LOG.trace("Processing action, namespace: {}, name: {}, method: {}", actionNamespace, actionName, actionMethod);
+            ActionProxy proxy = prepareActionProxy(extraContext, actionNamespace, actionName, actionMethod);
 
             request.setAttribute(ServletActionContext.STRUTS_VALUESTACK_KEY, proxy.getInvocation().getStack());
 
@@ -654,6 +647,36 @@ public class Dispatcher {
                 throw new ServletException(e);
             }
         }
+    }
+
+    private ActionProxy prepareActionProxy(Map<String, Object> extraContext, String actionNamespace, String actionName, String actionMethod) {
+        ActionProxy proxy;
+        //check if we are probably in an async resuming
+        ActionInvocation invocation = ActionContext.getContext().getActionInvocation();
+        if (invocation == null || invocation.isExecuted()) {
+            LOG.trace("Creating a new action, namespace: {}, name: {}, method: {}", actionNamespace, actionName, actionMethod);
+            proxy = createActionProxy(actionNamespace, actionName, actionMethod, extraContext);
+        } else {
+            proxy = invocation.getProxy();
+            if (isSameAction(proxy, actionNamespace, actionName, actionMethod)) {
+                LOG.trace("Proxy: {} matches requested action, namespace: {}, name: {}, method: {} - reusing proxy", proxy, actionNamespace, actionName, actionMethod);
+            } else {
+                LOG.trace("Proxy: {} doesn't match action namespace: {}, name: {}, method: {} - creating new proxy", proxy, actionNamespace, actionName, actionMethod);
+                proxy = createActionProxy(actionNamespace, actionName, actionMethod, extraContext);
+            }
+        }
+        return proxy;
+    }
+
+    private ActionProxy createActionProxy(String namespace, String name, String method, Map<String, Object> extraContext) {
+        ActionProxyFactory actionProxyFactory = getContainer().getInstance(ActionProxyFactory.class);
+        return actionProxyFactory.createActionProxy(namespace, name, method, extraContext, true, false);
+    }
+
+    private boolean isSameAction(ActionProxy actionProxy, String namespace, String actionName, String method) {
+        return Objects.equals(namespace, actionProxy.getNamespace())
+            && Objects.equals(actionName, actionProxy.getActionName())
+            && Objects.equals(method, actionProxy.getMethod());
     }
 
     /**
--- a/apps/showcase/pom.xml
+++ b/apps/showcase/pom.xml
@@ -121,6 +121,11 @@
             <artifactId>log4j-jcl</artifactId>
             <version>${log4j2.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
+            <version>${log4j2.version}</version>
+        </dependency>
 
         <dependency>
             <groupId>opensymphony</groupId>
@@ -191,10 +196,6 @@
                     <stopKey>CTRL+C</stopKey>
                     <stopPort>8999</stopPort>
                     <systemProperties>
-                        <systemProperty>
-                            <name>log4j.configuration</name>
-                            <value>file:${basedir}/src/main/resources/log4j2.xml</value>
-                        </systemProperty>
                         <systemProperty>
                             <name>slf4j</name>
                             <value>false</value>
--- a/apps/showcase/src/main/java/org/apache/struts2/showcase/source/ViewSourceAction.java
+++ b/apps/showcase/src/main/java/org/apache/struts2/showcase/source/ViewSourceAction.java
@@ -22,16 +22,13 @@ package org.apache.struts2.showcase.source;
 
 import com.opensymphony.xwork2.ActionSupport;
 import com.opensymphony.xwork2.util.ClassLoaderUtil;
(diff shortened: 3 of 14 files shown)


core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -656,6 +649,36 @@ public void serviceAction(HttpServletRequest request, HttpServletResponse respon
...
+        if (invocation == null || invocation.isExecuted()) {
+            LOG.trace("Creating a new action, namespace: {}, name: {}, method: {}", actionNamespace, actionName, actionMethod);
+            proxy = createActionProxy(actionNamespace, actionName, actionMethod, extraContext);
+        } else {
+            proxy = invocation.getProxy();
+            if (isSameAction(proxy, actionNamespace, actionName, actionMethod)) {
+                LOG.trace("Proxy: {} matches requested action, namespace: {}, name: {}, method: {} - reusing proxy", proxy, actionNamespace, actionName, actionMethod);
+            } else {
+                LOG.trace("Proxy: {} doesn't match action namespace: {}, name: {}, method: {} - creating new proxy", proxy, actionNamespace, actionName, actionMethod);
+                proxy = createActionProxy(actionNamespace, actionName, actionMethod, extraContext);
```
yasserzamani, line 665: to avoid StackOverflowException, shouldn't this line go to if(isSame) statement instead of else statement? because I think StackOverflow happens when we forward to same action again and again, but the if statement is empty, doesn't do anything functional, just logs.

yasserzamani:
Is it simple to have a testWW5190 test method which tests that this issue is fixed? Sorry I couldn't understand the root cause to write a test. My best guess is that it's because of Struts FORWARD happens in same thread, is it?

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
lukaszlenart: No, the problem here is with reusing the same `ActionInvocation` which is bound to the current thread (which means it wasn't recreated to handle a new action) and all this happens in the same thread: initial action -> JSP forward -> final action. Because of that the same result is executed again and again. Before introducing support for Async the action proxy was created each time from scratch. And `ActionInvocation#isExecuted()` flag won't help as this happens on a result level (the `executed` flag is set to `true` once a result has been executed, which won't happen here).

```
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /dispatch is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /dispatch as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:139) - Found mapping ActionMapping{name='dispatch', namespace='/', method='null', extension='null', params={}, result=null} for /dispatch
[DEBUG] interceptor.I18nInterceptor (I18nInterceptor.java:120) - Intercept '//dispatch'
[DEBUG] interceptor.I18nInterceptor (I18nInterceptor.java:167) - Using LocaleFinder implementation org.apache.struts2.interceptor.I18nInterceptor$SessionLocaleHandler
[DEBUG] interceptor.I18nInterceptor$RequestLocaleHandler (I18nInterceptor.java:247) - Searching locale in request under parameter request_only_locale
[DEBUG] interceptor.I18nInterceptor$SessionLocaleHandler (I18nInterceptor.java:318) - Searching locale in request under parameter request_locale
[DEBUG] interceptor.I18nInterceptor$SessionLocaleHandler (I18nInterceptor.java:345) - Checks session for saved locale
[DEBUG] interceptor.I18nInterceptor$SessionLocaleHandler (I18nInterceptor.java:360) - No Locale defined in session, fetching from current request and it won't be stored in session!
[DEBUG] interceptor.I18nInterceptor$RequestLocaleHandler (I18nInterceptor.java:264) - Searching current Invocation context
[DEBUG] interceptor.I18nInterceptor$RequestLocaleHandler (I18nInterceptor.java:268) - Applied invocation context locale: en_US
[DEBUG] interceptor.I18nInterceptor (I18nInterceptor.java:136) - Before action invocation Locale=en_US
[DEBUG] interceptor.FileUploadInterceptor (FileUploadInterceptor.java:239) - Bypassing //dispatch
[DEBUG] result.ServletDispatcherResult (ServletDispatcherResult.java:127) - Forwarding to location: /dispatch-result.jsp
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /dispatch-result.jsp is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /dispatch-result.jsp as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:136) - Cannot find mapping for /dispatch-result.jsp, passing to other filters
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /index.action is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /index.action as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:139) - Found mapping ActionMapping{name='index', namespace='/', method='null', extension='action', params={}, result=null} for /index.action
[DEBUG] result.ServletDispatcherResult (ServletDispatcherResult.java:127) - Forwarding to location: /dispatch-result.jsp
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /dispatch-result.jsp is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /dispatch-result.jsp as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:136) - Cannot find mapping for /dispatch-result.jsp, passing to other filters
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /index.action is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /index.action as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:139) - Found mapping ActionMapping{name='index', namespace='/', method='null', extension='action', params={}, result=null} for /index.action
[DEBUG] result.ServletDispatcherResult (ServletDispatcherResult.java:127) - Forwarding to location: /dispatch-result.jsp
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /dispatch-result.jsp is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /dispatch-result.jsp as a normal action
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:136) - Cannot find mapping for /dispatch-result.jsp, passing to other filters
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:126) - Checking if /index.action is a static resource
[TRACE] filter.StrutsPrepareAndExecuteFilter (StrutsPrepareAndExecuteFilter.java:129) - Assuming uri /index.action as a normal action
```

lukaszlenart:
> Is it simple to have a testWW5190 test method which tests that this issue is fixed? Sorry I couldn't understand the root cause to write a test. My best guess is that it's because of Struts FORWARD happens in same thread, is it?

The simplest was is to define a dispatcher example in Showcase app and define an integration test for that.

lukaszlenart:
There is some discrepancy between `StrutsPrepareAndExcuteFilter` and  `StrutsPrepareFilter` & `StrutsExcuteFilter` which I'm trying to figure out

lukaszlenart:
I have integration test but it won't work with the Showcase App because of those problems.

yasserzamani:
So it makes sense to have those in this PR please. awesome!! thank you very much!

lukaszlenart:
@yasserzamani I extend description of the PR as the problem is a bit more complicated. My fix addresses only the broken support for forwarding to an another action when using `StrutsPrepareAndExecuteFilter`, yet the Showcase App is using combination of `StrutsPrepareFilter`/`StrutsExecuteFilter` and this never worked properly - I created a dedicated ticket to fix this with the next Minor release.

burtonrhodes:
This change is breaking my ExecuteAndWait implementations using the [recommended refresh implementation](https://struts.apache.org/core-developers/execute-and-wait-interceptor). The resulting url appends the file with "!methodName" making the url invalid.

A real world example is below:

```
<action name="LetterMergeBulk_contactList" class="com.afs.web.struts.action.letters.LetterMergeBulkAction" method="contactList">
    <interceptor-ref name="openSessionExecuteAndWaitInterceptor" />
    <result name="wait">/struts/letters/letterMergeBulkWait_popup.jsp</result>
    <result>/struts/letters/letterMerge_popup.jsp</result>
</action>
```

Results in (v6.0.3)
```
/afs/app/LetterMergeBulk_contactList!contactList.action?struts.token.name=token&token=A9SIKBABK17IZFM6AHSP0QSFIJHLZKJB&letterTemplateId=14773
```

Instead of (v6.0.0)
```
/afs/app/LetterMergeBulk_contactList.action?struts.token.name=token&token=A9SIKBABK17IZFM6AHSP0QSFIJHLZKJB&letterTemplateId=14773
```

Using HTML Header
```
<head>
   <title>Please wait</title>
   <meta http-equiv="refresh" content="5;url=<s:url includeParams="all" />"/>
</head>
```

lukaszlenart:
@burtonrhodes do you use DMI?

burtonrhodes:
I don't use DMI, but have no specific setting in my struts.xml file regarding it. However, given your suggestion, I specifically disabled DMI to see if that would help - but the result is still the same.

`<constant name="struts.enable.DynamicMethodInvocation" value="false"/>`

lukaszlenart:
Ok, so please register a new JIRA ticket and we can start working on solution after figuring out what's wrong.

burtonrhodes:
Done! [https://issues.apache.org/jira/browse/WW-5241](https://issues.apache.org/jira/browse/WW-5241)

PR #581, opened 2022-07-30
[WW-5173] - Fix DI behaviour for custom cache factories

yasserzamani:


yasserzamani:
already included in #573 :)

PR #583, opened 2022-08-06
WW-5208 Update hibernate-validator to 6.2.4

sepe81:
b/c 6.1 has reached its end-of-life according to
https://hibernate.org/validator/releases/6.1/

6.2 is compatible with the given Java and Jakarta
versions and supports `javax.validation.*` according to
https://hibernate.org/validator/releases/6.2/


lukaszlenart:
LGTM 👍

PR #582, opened 2022-08-06
Update maven wrapper to 3.8.6

sepe81:

PR #584, opened 2022-08-06
Add openjdk17 to build, remove oraclejdk9 (EOL) and switch from oraclejdk to openjdk to fix Travis CI build

sepe81:
In reference to https://github.com/apache/struts/pull/582#issuecomment-1207209619 and to fix the build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: java
 sudo: false
 
 jdk:
-  - oraclejdk9
-  - oraclejdk11
+  - openjdk11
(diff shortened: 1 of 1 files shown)


lukaszlenart:
I switched to OpenJDK and now it works - I think the same has to be done here
https://github.com/apache/struts-examples/blob/master/.travis.yml

lukaszlenart:
Build passed 💯

lukaszlenart:
I assume is _ready to go_, so LGTM 👍

PR #589, opened 2022-08-22
@Before and @After related question in StrutsJUnit4TestCase class

fp024:
Hello.

StrutsJUnit4TestCase class is not using JUnit4's `@Before`, `@After`.

It works because JUnit4 doesn't even check the package name, it only checks the annotation name.
 
But semantically, I think it's correct to use JUnit4's annotations.

There is no issue tab page, so I made a Pull Request.

Thanks you.


fp024:
commit-message-issue

Github parses the Commit Message strangely, so I'll close it.

I'm sorry.

PR #590, opened 2022-08-23
[WW-5215] Checks if session was already created before applying CSP settings

lukaszlenart:
Fixes [WW-5215](https://issues.apache.org/jira/browse/WW-5215)


sepe81:
Maybe change "Checks is session was already..." to "Checks if session was already..." for better understanding

lukaszlenart:
> Maybe change "Checks is session was already..." to "Checks if session was already..." for better understanding

Right, typo :)

PR #593, opened 2022-09-01
[WW-5218] Allows to disable CSP related interceptors

lukaszlenart:
Fixes [WW-5218](https://issues.apache.org/jira/browse/WW-5218)


yasserzamani (approved):
LGTM thanks!!

PR #608, opened 2022-10-01
[WW-5233] Include Apache Tiles code base in the Tiles plugin

lukaszlenart:
Fixes [WW-5233](https://issues.apache.org/jira/browse/WW-5233)


lukaszlenart:
@gregh3269 @JCgH4164838Gh792C124B5 @kusalk any objections to merge this and release a new version?

kusalk:
We are not using Tiles or these plugins so no objections here :)

gregh3269:
+1.  But then i am standard wrt tiles.

HittiGubbi:
Is there some instructions how to migrate to 6.3.0 from an external Tiles dependency? The 6.3.0 release notes state that "when you use the Struts Tiles plugin please remove any external dependencies to the Apache Tiles". Does this mean that I should remove the dependency to tiles-core from pom.xml? If I do that, the web app (which works fine on 6.2.0) fails to start due to an error:

`java.lang.ClassNotFoundException: org.apache.tiles.extras.complete.CompleteAutoloadTilesListener`

gregh3269:
@ HittiGubbi  what is this used for?

kusalk:
Yes that listener is from the old deprecated plugin - see [here](https://struts.apache.org/plugins/tiles/) for updated instructions

HittiGubbi:
Thank you. I tried this but now get the error:

`java.lang.NoSuchMethodError: 'void org.apache.commons.digester.Digester.setXIncludeAware(boolean)`

kusalk:
Looks like you're using a different version of `commons-digester` to the one included by Struts

gregh3269:
..By using the org.apache.tiles.extras.complete.CompleteAutoloadTilesListener you get Wild Cards, EL, OGNL, MVEL support in your tiles.xml

https://stackoverflow.com/questions/10050433/ognl-and-wildcards-working-in-tiles-definitions-with-struts2-tiles-plugin/10067367#10067367

Can you share your tiles.xml please?

gregh3269:
Can you try dropping the struts2-tiles-plugin and replace with tiles extras as it does not look like you are using the struts2 tiles plugin ie StrutsTilesListener.
```
<dependency>
   	<groupId>org.apache.tiles</groupId>
    	<artifactId>tiles-extras</artifactId>
    	<version>3.0.8</version>
</dependency>
```
I tested this and it will start, but it brings a very old velocity 1.6, which won't work for me.

In the struts2-tiles-plugin <6.2.0 it did not have many classes,  possibly none were being used.

PR #610, opened 2022-10-07
Add missing classes and tld definition.

gregh3269:
Works great for me.


gregh3269:
Coveralls thingy

GetAsStringTag
from: private GetAsStringModel model = new GetAsStringModel());
to: private GetAsStringModel model = new GetAsStringModel(new DefaultAttributeResolver());

and then remove change:
GetAsStringModel:
 /**
     * Constructor that uses the defaut attribute resolver.
     *
     * @since 3.0.0
     */
    public GetAsStringModel() {
        this(new DefaultAttributeResolver());
    }
It does not say why it does not like the constructor?

lukaszlenart (approved):

PR #611, opened 2022-10-13
[WW-5241] Fixes issue with includeParams=all when using exec & wait

lukaszlenart:
Fixes [WW-5241](https://issues.apache.org/jira/browse/WW-5241)


lukaszlenart:
@burtonrhodes fixed, the bug was in `DefaultActionMapper` which ignored DMI flag and was appending `!method` every time.

burtonrhodes:
This fix is working nicely on my end. Many thanks.

PR #642, opened 2022-12-08
WW-5199 Allow forwarding from/to actions

kusalk:
Fixes [WW-5199](https://issues.apache.org/jira/browse/WW-5199)


lukaszlenart:
Thanks a lot! I had no idea that the fix would so easy :)
LGTM 👍

PR #648, opened 2022-12-19
WW-5270 Fix forwarding from Struts excluded URL

kusalk:
Implements WW-5270

### DIFF
--- a/apps/showcase/src/main/java/org/apache/struts2/showcase/servlet/TestServlet.java
+++ b/apps/showcase/src/main/java/org/apache/struts2/showcase/servlet/TestServlet.java
@@ -1,0 +1,41 @@
+/*
+ * $Id$
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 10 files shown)


kusalk:
My initial ThreadLocal based approach to reworking the request cleanup process was not functioning as intended. I've reverted to tracking filter recursion as a request attribute but have still retained the reworked logic.

This means:
- ActionContext is not cleared until the first StrutsPrepareFilter execution for that request has completed - this is irrespective of whether the ActionContext was generated there or not. This can occur when a Struts excluded URL forwards to a non-excluded one. Waiting until the ActionContext is cleared is required by SiteMesh.
- The MultiPart component of a request is only cleaned up in the filter execution in which it was wrapped. This is retaining the original behaviour as before. I believe if we try to cleanup at any other stage, it will fail as the `instanceof` check will fail - although I haven't investigated this closely.

The implementation is a little confusing at first glance (and was also why I favoured the ThreadLocal based approach) but it's the best I've got right now - open to suggestions.

lukaszlenart (approved):
Sorry, I totally missed this PR, LGTM 👍

lukaszlenart:
@kusalk could you resolve conflicts? I see nothing but GH blocks merging this PR :\

kusalk:
@lukaszlenart All good, should be sorted now. We shipped this change in Confluence 8.0.2 on Dec 22nd and there have been no reported issues.

lukaszlenart (approved):

PR #655, opened 2023-01-22
Http interceptor

lukaszlenart:
Implements `HttpInterceptor` to allow limit handling actions to given HTTP methods
Closes [WW-4404](https://issues.apache.org/jira/browse/WW-4404)


lukaszlenart:
@jogep @davelnewton please take a look in a spare time as you had been commenting on the previous PRs with the same :)

PR #657, opened 2023-01-30
WW-5279 Improve readability of XmlConfigurationProvider class

kusalk:
WW-5279

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
@@ -44,7 +44,6 @@ import com.opensymphony.xwork2.inject.Scope;
 import com.opensymphony.xwork2.util.ClassLoaderUtil;
 import com.opensymphony.xwork2.util.ClassPathFinder;
 import com.opensymphony.xwork2.util.DomHelper;
-import com.opensymphony.xwork2.util.TextParseUtil;
 import com.opensymphony.xwork2.util.location.LocatableProperties;
 import com.opensymphony.xwork2.util.location.Location;
 import com.opensymphony.xwork2.util.location.LocationUtils;
@@ -75,6 +74,17 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.Vector;
+import java.util.function.Consumer;
+
+import static com.opensymphony.xwork2.util.TextParseUtil.commaDelimitedStringToSet;
+import static java.lang.Boolean.parseBoolean;
+import static java.lang.Character.isLowerCase;
+import static java.lang.Character.toUpperCase;
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static org.apache.commons.lang3.StringUtils.defaultString;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.trimToNull;
 
 
 /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
```diff
@@ -164,13 +174,10 @@ public boolean equals(Object o) {
...
             return true;
         }
-
         if (!(o instanceof XmlConfigurationProvider)) {
             return false;
         }
-
-        final XmlConfigurationProvider xmlConfigurationProvider = (XmlConfigurationProvider) o;
-
+        XmlConfigurationProvider xmlConfigurationProvider = (XmlConfigurationProvider) o;
```
sepe81, line 180:
```suggestion
        final XmlConfigurationProvider xmlConfigurationProvider = (XmlConfigurationProvider) o;
```

this could remain final

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
```diff
@@ -439,28 +445,16 @@ protected void addAction(Element actionElement, PackageConfig.Builder packageCon
...
-            } else {
-                className = ActionSupport.class.getName();
-            }*/
-
-        } else {
-            if (!verifyAction(className, name, location)) {
-                LOG.error("Unable to verify action [{}] with class [{}], from [{}]", name, className, location);
-                return;
-            }
+        if (!className.isEmpty() && !verifyAction(className, name, location)) {
```
sepe81, line 455: Maybe we can remove the unused param `name` from verifyAction L:490

Sonar says: Parameter 'name' is never used

sepe81 (commented):
not easy to overlook the big diff, but good to see the cognitive complexity reduced afterwards

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
kusalk: I find it clutters the code and goes against [present common practice](https://stackoverflow.com/questions/154314/when-should-one-use-final-for-method-parameters-and-local-variables) - but I can revert these changes if that is the preferred code style.

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
kusalk: There's multiple methods in this class with unused params - I've kept them to maintain compatibility with potential subclasses. Perhaps something to revisit when releasing the next major version.

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
sepe81: https://stackoverflow.com/questions/154314/when-should-one-use-final-for-method-parameters-and-local-variables/154510#154510 LGTM, but for this existing line "If I'm in someone else's code, I'm not going to pull them out but if I'm writing new code I won't put them in." would fit.

With your explanation, both styles would be ok for me. Maybe @lukaszlenart has some recommendation according to the Struts coding style?

lukaszlenart: I'm fine with dropping `final` in this context, I like using `final`s for fields and function arguments, where it makes sense :)

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlConfigurationProvider.java
lukaszlenart: @sepe81 I'm right now creating a task to mark a given code `@Deprecated` and then another task (targeting major/minor release) to remove the code. I found such approach more useful and more informative for the users :)

lukaszlenart (approved):
LGTM 👍

PR #656, opened 2023-01-30
WW-5278 Collect duplicated code into AbstractActionValidatorManager

kusalk:
WW-5278

PR #659, opened 2023-02-14
WW-5284 Refactor ActionValidatorManager implementations

kusalk:
WW-5284
--
Following on from #656 I realised the implementation could be simplified further. I've merged AbstractActionValidationManager into DefaultActionValidationManager and made AnnotationActionValidationManager an extension of that. Other than that I've only made a handful of functional changes which I'll comment below.


core/src/main/java/com/opensymphony/xwork2/validator/DefaultActionValidatorManager.java
```diff
@@ -43,64 +60,292 @@
...
+
+    protected Validator getValidatorFromValidatorConfig(ValidatorConfig config, ValueStack stack) {
+        Validator validator = validatorFactory.getValidator(config);
+        validator.setValidatorType(config.getType());
+        validator.setValueStack(stack);
+        return validator;
     }
 
     @Override
     public synchronized List<Validator> getValidators(Class clazz, String context, String method) {
```
kusalk, line 140: Given that both `AnnotationActionValidatorManager` and `DefaultActionValidatorManager` use the same caching mechanism I presume both should be synchronised on this method. Previously `AnnotationActionValidatorManager` was not. Regardless, I don't think it will hurt to add it here for the purpose of simplifying the code.

core/src/main/java/com/opensymphony/xwork2/validator/DefaultActionValidatorManager.java
```diff
@@ -43,64 +60,292 @@
...
         String validatorKey = buildValidatorKey(clazz, context);
 
-        if (validatorCache.containsKey(validatorKey)) {
-            if (reloadingConfigs) {
-                validatorCache.put(validatorKey, buildValidatorConfigs(clazz, context, true, null));
-            }
-        } else {
+        if (!validatorCache.containsKey(validatorKey)) {
```
kusalk, line 143: Same logic, just reduced nesting

core/src/test/java/com/opensymphony/xwork2/validator/DefaultActionValidatorManagerTest.java
```diff
@@ -87,7 +87,7 @@ protected void tearDown() throws Exception {
 
 
     public void testBuildValidatorKey() {
-        String validatorKey = DefaultActionValidatorManager.buildValidatorKey(SimpleAction.class, alias);
+        String validatorKey = actionValidatorManager.buildValidatorKey(SimpleAction.class, alias);
```
kusalk, line 90: Changed this to an instance method for consistency between the 2 implementations

core/src/main/java/com/opensymphony/xwork2/validator/AnnotationActionValidatorManager.java
```diff
@@ -120,10 +66,20 @@ protected String buildValidatorKey(Class clazz, String context) {
...
         }
-
         return sb.toString();
     }
 
+    @Override
+    protected Validator getValidatorFromValidatorConfig(ValidatorConfig config, ValueStack stack) {
+        Validator validator = validatorFactory.getValidator(
+                new ValidatorConfig.Builder(config)
+                        .removeParam("methodName")
```
kusalk, line 76: Is this override necessary? I'm not completely clear on what it's achieving. Would be nice to delete it if it doesn't do anything.

core/src/main/java/com/opensymphony/xwork2/validator/AbstractActionValidatorManager.java
```diff
@@ -1,289 +0,0 @@
...
-                            break;
-                        }
-                    }
-
-                    continue;
-                }
-
-                validator.validate(object);
-            } finally {
-                validator.setValidatorContext(null);
```
kusalk, line 176: As far as I can tell this doesn't do anything. Probably a remnant from when we used to cache the Validators instead of the ValidatorConfigs.

core/src/main/java/com/opensymphony/xwork2/validator/AnnotationActionValidatorManager.java
```diff
@@ -37,56 +37,11 @@
  * @author Rainer Hermanns
  * @author jepjep
  */
-public class AnnotationActionValidatorManager extends AbstractActionValidatorManager {
+public class AnnotationActionValidatorManager extends DefaultActionValidatorManager {
```
kusalk, line 40: It's now much clearer how this class differs to the default one.

core/src/main/java/com/opensymphony/xwork2/validator/AnnotationActionValidatorManager.java
lukaszlenart: Looks like it does nothing right now, you can try to remove it.

kusalk: Done :)

lukaszlenart (approved):
LGTM 👍

PR #662, opened 2023-02-25
[WW-5285] Limits max number of files to upload at once

lukaszlenart:
Upgrades commons-fileupload to ver. 1.5 and sets default limit to 256 files

Closes [WW-5285](https://issues.apache.org/jira/browse/WW-5285)

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
@@ -55,6 +55,12 @@ public abstract class AbstractMultiPartRequest implements MultiPartRequest {
     protected boolean maxSizeProvided;
 
     /**
+     * Specifies the maximum number of files in one request.
+     */
+    protected long maxFiles;
+    protected boolean maxFilesProvided;
+
+    /**
      * Specifies the buffer size to use during streaming.
      */
     protected int bufferSize = BUFFER_SIZE;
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -54,6 +54,12 @@ public abstract class AbstractMultiPartRequest implements MultiPartRequest {
     protected long maxSize;
     protected boolean maxSizeProvided;
 
+    /**
+     * Specifies the maximum number of files in one request.
+     */
+    protected long maxFiles;
+    protected boolean maxFilesProvided;
```
yasserzamani, line 61: nitpick: can't we use Long instead of long here? and a null value will mean that it's not provided

yasserzamani (approved):

aleksandr-m (approved):

core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
lukaszlenart: I have used the same approach as with `maxSize`, but maybe you are right to use `null` as a _not set_ indicator

lukaszlenart: Done

yasserzamani (approved):

PR #663, opened 2023-03-01
WW-5267 Add option to generate ActionContext for excluded URLs

kusalk:
WW-5267
--
SiteMesh can be applied to URLs that are not Struts actions, in which case we need to still ensure the ActionContext exists.

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
@@ -45,7 +45,7 @@ public class StrutsPrepareFilter implements StrutsStatics, Filter {
 
     protected PrepareOperations prepare;
     protected List<Pattern> excludedPatterns = null;
-    protected boolean alwaysCreateActionContext = false;
+    private Dispatcher dispatcher;
 
     public void init(FilterConfig filterConfig) throws ServletException {
         InitOperations init = createInitOperations();
@@ -107,7 +107,7 @@ public class StrutsPrepareFilter implements StrutsStatics, Filter {
             prepare.trackRecursion(request);
             if (excludedPatterns != null && prepare.isUrlExcluded(request, excludedPatterns)) {
                 request.setAttribute(REQUEST_EXCLUDED_FROM_ACTION_MAPPING, true);
-                if (alwaysCreateActionContext) {
+                if (alwaysCreateActionContext()) {
                     prepare.createActionContext(request, response);
                 }
             } else {
@@ -126,6 +126,11 @@ public class StrutsPrepareFilter implements StrutsStatics, Filter {
             }
             prepare.cleanupRequest(request);
         }
+    }
+
+    private boolean alwaysCreateActionContext() {
+        return Boolean.parseBoolean(dispatcher.getContainer()
+                .getInstance(String.class, StrutsConstants.STRUTS_ALWAYS_CREATE_ACTION_CONTEXT));
     }
 
     public void destroy() {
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/apache/struts2/dispatcher/TwoFilterIntegrationTest.java
```diff
@@ -19,87 +19,133 @@
...
         MockHttpServletResponse response = run("/hello.action", filterPrepare, middle, filterExecute, failFilter);
         assertEquals(200, response.getStatus());
     }
 
+    /**
+     * It is possible for a Struts excluded URL to be forwarded to a Struts URL. If this happens, the ActionContext
+     * should not be cleared until the very first execution of the StrutsPrepareFilter, otherwise SiteMesh will malfunction.
+     */
+    @Test
+    public void testActionContextNotClearedUntilEndWhenForwardedFromExcludedUrl() throws ServletException, IOException {
```
kusalk, line 113: This integration test is actually for [WW-5270](https://issues.apache.org/jira/browse/WW-5270). (We already have acceptance test coverage for that fix.)

core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
```diff
@@ -44,6 +46,12 @@ public class StrutsPrepareFilter implements StrutsStatics, Filter {
 
     protected PrepareOperations prepare;
     protected List<Pattern> excludedPatterns = null;
+    protected boolean alwaysCreateActionContext = false;
+
+    @Inject(value = StrutsConstants.STRUTS_ALWAYS_CREATE_ACTION_CONTEXT, required = false)
```
kusalk, line 51: Actually - we can't inject here can we? 😬

kusalk: All sorted, ready for review :)

core/src/test/java/org/apache/struts2/dispatcher/PrepareOperationsTest.java
```diff
@@ -73,6 +73,7 @@ public void testWrappedRequestCleanup() {
         });
         IntStream.range(0, mockedRecursions - 1).forEach(i -> prepare.cleanupWrappedRequest(req));
 
+        // Assert org.apache.struts2.dispatcher.Dispatcher#cleanUpRequest has not yet run
```
lukaszlenart, line 76: ?

lukaszlenart (commented):
One doubt about commented out code and as far I understand, right now it would be possible to access `ActionContext` out of action, directly from JSP? Is it a security vulnerability?

core/src/test/java/org/apache/struts2/dispatcher/PrepareOperationsTest.java
kusalk: This comment was meant to clarify the intent of line 77. This unit test is testing that if prepare#wrapRequest is run X times, then prepare#cleanupWrappedRequest must also be run X times before the request is cleaned up. We do not want to clean up too early (line 77) and we don't want to forget to clean up (line 79).

kusalk:
> right now it would be possible to access `ActionContext` out of action, directly from JSP?

Not clear on what you mean, how so?

lukaszlenart:
> > right now it would be possible to access `ActionContext` out of action, directly from JSP?
> 
> Not clear on what you mean, how so?

If the flag is "on" and the request matches excluded urls, the `ActionContext` will be available in non-Struts managed endpoints. By design all the requests should be handled by the actions first and then forwarded into view layer (like JSP or Freemarker) - this also involves the whole security mechanism embedded into _normal_ flow (interceptors).

With this change it is possible to overuse this functionality by having an excluded url and still accessing `ActionContext` out of action scope directly from JSP or Freemarker. User is "escaping" from Struts sandbox, yet having option to operate on `ActionContext` like `ActionContext.getContext().getContainer()`.

This raises security concerns tbh.

kusalk:
> User is "escaping" from Struts sandbox, yet having option to operate on `ActionContext` like `ActionContext.getContext().getContainer()`.

Ah, I see what you mean. Let me rethink this problem and see if I can come up with a solution that doesn't violate `struts.action.excludePattern`. Thanks for the feedback

PR #665, opened 2023-03-02
WW-5266 Implement struts.multipart.maxFileSize

kusalk:
WW-5266
--

### DIFF
--- a/core/src/main/resources/org/apache/struts2/default.properties
+++ b/core/src/main/resources/org/apache/struts2/default.properties
@@ -69,6 +69,7 @@ struts.multipart.parser=jakarta
 struts.multipart.saveDir=
 struts.multipart.maxSize=2097152
 struts.multipart.maxFiles=256
+struts.multipart.maxFileSize=2097152
 
 ### Load custom property files (does not override struts.properties!)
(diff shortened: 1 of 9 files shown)


core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -69,6 +69,7 @@ struts.multipart.parser=jakarta
 struts.multipart.saveDir=
 struts.multipart.maxSize=2097152
 struts.multipart.maxFiles=256
+struts.multipart.maxFileSize=2097152
```
kusalk, line 72: We could always make this `-1` so it's unused by default and won't break backwards compatibility for existing Struts applications.

lukaszlenart: Makes sense, and also this option somehow duplicates what we have on the interceptor level - so this is a global config, which takes precedence over the interceptor configuration.

I would just comment out this line and use `null` as a indicator to mark this option disabled to match the same behaviour for other options.

kusalk: Done :)

lukaszlenart (approved):
LGTM 👍

PR #668, opened 2023-03-12
WW-5293 Allow loading XML configuration from other than filesystem

kusalk:
WW-5293
--
I've split XmlConfigurationProvider into a superclass XmlDocsConfigurationProvider. The original XmlConfigurationProvider now extends XmlDocsConfigurationProvider to provide filesystem specific loading functionality. The new XmlDocsConfigurationProvider can be instantiated on its own by feeding in a parsed XML document directly.

Additionally:
- Deprecated some old constructors/methods so that they can be safely deleted in Struts 7.0
- Allowed greater flexibility in subclassing by splitting the build action/interceptor/result configs into their own methods
- Fixed a minor existing bug where `#registerBeanSelection` did not respect the ObjectFactory configured class loader
- Fixed a minor bug introduced in #657 where an unhandled exception is thrown when calling `#loadPackages` of an XmlConfigurationProvider of an already loaded XML file

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlDocConfigurationProvider.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/providers/XmlDocConfigurationProvider.java
@@ -426,8 +426,7 @@ public abstract class XmlDocConfigurationProvider implements ConfigurationProvid
     protected void addAction(Element actionElement, PackageConfig.Builder packageContext) throws ConfigurationException {
         String name = actionElement.getAttribute("name");
         String className = actionElement.getAttribute("class");
-        //methodName should be null if it's not set
-        String methodName = trimToNull(actionElement.getAttribute("method"));
+
         Location location = DomHelper.getLocationObject(actionElement);
 
         if (location == null) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/opensymphony/xwork2/config/providers/XmlDocConfigurationProvider.java
```diff
@@ -0,0 +1,981 @@
...
+import static com.opensymphony.xwork2.util.TextParseUtil.commaDelimitedStringToSet;
+import static java.lang.Boolean.parseBoolean;
+import static java.lang.Character.isLowerCase;
+import static java.lang.Character.toUpperCase;
+import static java.lang.String.format;
+import static org.apache.commons.lang3.StringUtils.defaultString;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.trimToNull;
+
+public abstract class XmlDocConfigurationProvider implements ConfigurationProvider {
```
lukaszlenart, line 77: I would add a minimal JavaDoc and `@since` tag

lukaszlenart (changes requested):

core/src/main/java/com/opensymphony/xwork2/config/providers/XmlDocConfigurationProvider.java
kusalk: Done and rebased :)

lukaszlenart (approved):
LGTM 👍

PR #680, opened 2023-04-09
[WW-5304] Drops deprecated methods and fields in ActionContext

lukaszlenart:
Closes [WW-5304](https://issues.apache.org/jira/browse/WW-5304)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ActionContext.java
@@ -58,79 +58,52 @@ import java.util.Map;
  */
 public class ActionContext implements Serializable {
 
-    static ThreadLocal<ActionContext> actionContext = new ThreadLocal<>();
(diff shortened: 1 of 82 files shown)


core/src/main/java/com/opensymphony/xwork2/ActionContext.java
```diff
@@ -157,6 +129,15 @@ public static ActionContext of(Map<String, Object> context) {
         return new ActionContext(context);
     }
 
+    /**
+     * Creates a new ActionContext based on empty Map
+     *
+     * @return new ActionContext
+     */
+    public static ActionContext of() {
```
kusalk, line 137: `create()` might be a more descriptive name for this method

lukaszlenart: There is already `of(Map<String, Object> context)` so I stick to the same convention

PR #682, opened 2023-05-05
Improve [WW-4434] - add documentation and rename existing ftl to achieve the wanted behaviour

fischey:
I stumbled across the ticket https://issues.apache.org/jira/browse/WW-4434 during development, because I wanted to use the s:datetextfield.
I still got an exception In version 2.5.30 as I used the tag in the simple theme.
With this PR I want to improve the documentation and fix the current available ftl so that others will not get this error too :-)

Best regards
fischey


lukaszlenart (approved):
LGTM 👍

PR #683, opened 2023-05-09
WW-5308 Java templates plugin, add minlength and maxlength to textarea.

gregh3269:
Java templates plugin, add minlength and maxlength to textarea.


lukaszlenart (approved):
LGTM 👍

PR #688, opened 2023-05-20
[WW-5312] Attempt to fix ExecuteAndWaitInterceptor inconsistent processing

JCgH4164838Gh792C124B5:
Update:
- Improve ExecuteAndWaitInterceptor state behaviour, added debug and trace logging.
- Ensure StrutsBackgroundProcess thread done state always set on completion.
- Fix SessionMap processing issue caused by divergence of method signatures, which can result in the ancestor methods getting called (bypassing actual session processing).
- Update SessionMapTest to correspond to changes, add tests to try and detect some (put/remove) method signature behaviour changes.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/ExecuteAndWaitInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ExecuteAndWaitInterceptor.java
@@ -228,6 +228,7 @@ public class ExecuteAndWaitInterceptor extends MethodFilterInterceptor {
     /* (non-Javadoc)
      * @see com.opensymphony.xwork2.interceptor.MethodFilterInterceptor#doIntercept(com.opensymphony.xwork2.ActionInvocation)
      */
+    @Override
     protected String doIntercept(ActionInvocation actionInvocation) throws Exception {
         ActionProxy proxy = actionInvocation.getProxy();
         String name = getBackgroundProcessName(proxy);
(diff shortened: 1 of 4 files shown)


JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

The commit summary basically covers what is being attempted in the PR.  Other than some minor cleanup (in the modules where changes were made), the changes were an attempt to fix the issue found with WW-5312.

The regression tests pass (locally), as well as interactive testing using the Struts2 Showcase application, but the changes might still need improvements.  The ```SessionMap``` method signature changes appear to be needed no matter what, though, in order to ensure the descendant method gets called even when type erasure takes place (and having the ```@Override``` annotations present should help ensure the method signatures stay consistent).

Please review and advise if the PR looks OK, and if there are any recommended changes.

_Note_: It looks like there was a "sonar-maven-plugin" failure for the Java 11 build environment run, but the build itself seems to have completed OK.  Possibly an issue with the automated build ?

core/src/main/java/org/apache/struts2/interceptor/ExecuteAndWaitInterceptor.java
```diff
@@ -394,7 +404,10 @@ public void init() {
 
     @Override
     public void destroy() {
-        super.destroy();
-        executor.shutdown();
+        try {
+          executor.shutdown();
+        } finally {
+          super.destroy();
+        }
```
lukaszlenart, line 411: 👏

lukaszlenart (commented):
Nice, I like your changes. One question about `SessionMap` - wouldn't be better to use na internal map and delegate calls to it instead of extending `AbstractMap`?

core/src/main/java/org/apache/struts2/interceptor/ExecuteAndWaitInterceptor.java
JCgH4164838Gh792C124B5: Hi Lukasz.  Thanks for reviewing the proposed fix changes.  😄 

Looking at the commit history, `SessionMap`  has extended `AbstractMap` since its earliest commits (many, many years ago).  Since `SessionMap` is _public_ and has been structured a certain way for so long, changing its design to delegate (to a non-accessible internal map, instead of extending) could have unexpected impacts (especially for anything that might try to use/extend it in user-space code).

Staying with the long-standing (preexisting) design might be safer in terms of potential side-effects.  Keeping `@overrides` annotations in place along with some unit tests that attempt to detect breaking-changes should help.  Considering a `SessionMap` redesign would probably need to be a standalone issue/work item, if people think it is needed (taking into account the existing design seems to have worked sufficiently for a long time).

Hopefully my reasoning above makes sense.  If it seems like other changes/improvements should be made to the PR, please let me know.

lukaszlenart: Thanks a lot for your comment, yet I was thinking about redesigning the current *Map approach (there is ApplicationMap, RequestMap as well) and use native objects to have more control over them. Right now we must prevent extra cuotion to secure the framweork (excluded `#application` or `#request` in ParametersInterceptor).

Anyway this will happen in Struts 7.0 at leas :)

lukaszlenart (approved):
LGTM 👍

PR #692, opened 2023-05-29
[WW-5310] Supports fragment in URL

lukaszlenart:
Closes [WW-5310](https://issues.apache.org/jira/browse/WW-5310)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/ServletUrlRenderer.java
+++ b/core/src/main/java/org/apache/struts2/components/ServletUrlRenderer.java
@@ -159,10 +159,10 @@ public class ServletUrlRenderer implements UrlRenderer {
             }
         }
 
-        Map<String, Object> actionParams = null;
+        QueryStringParser.Result queryStringResult = queryStringParser.empty();
         if (action != null && action.indexOf('?') > 0) {
             String queryString = action.substring(action.indexOf('?') + 1);
-            actionParams = queryStringParser.parse(queryString, false);
+            queryStringResult = queryStringParser.parse(queryString);
             action = action.substring(0, action.indexOf('?'));
         }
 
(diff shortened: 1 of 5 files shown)


nikosdimitrakas:
Shouldn't the places that call QueryStringParser.parse keep track of the fragment identifier since it will no longer be included in the result of the method parse (as the end of the value of the last parameter, which previously only worked when the fragment identifier did not contain special characters)? So that the fragment identifier can be appended when the url has been recomposed. Sending a url with a fragment identifier to an s:url tag should give it back without losing the fragment identifier. At least as long as the s:url does not have a new fragment identifier specified with the anchor attribute.

lukaszlenart:
Yet I'm not sure if this is the parser's responsibility, the whole idea with using `<s:url/>` is about constructing an url based on action's values and not to apply an existing url and recompose it back - in such case you can simple use `<a href/>` tag.

nikosdimitrakas:
We combine s:url and a as below in order to get the url escaped correctly. (We have complete urls provided by the users)
```
<s:url var="urlValue" value="%{url}"/>
<a href="${urlValue}">click me</a>
```
Without the s:url the escaping is wrong (s:property does not escape urls correctly, s:a doesn't either). Unfortunately I do not remember the exact characters that were mishandled without s:url.

lukaszlenart:
To support this properly I would have to change signature to return an object like `ParserResult(Map<String, String> params, String fragment)` but this will break backward compatibility. I will close this PR and left implementation _as is_ and register a ticket to fix that in Struts 7.x

nikosdimitrakas:
You could fix this without changing the signature of the method by defining that the incoming string must be the query string and nothing more. The fact that the few callers of the parse method provide everything after the ? in the url is the problem in my opinion. The class QueryStringParser should only be working with the query string. This way you can maintain backwards compatibility and make sure that the query string is handled correctly in the QuerystringParser and that the fragment identifier is handled correctly in the other classes (ServletUrlRenderer, ServlerDispatcherResult and DefaultUrlHelper).

lukaszlenart:
Right, but this will be rather a workaround

lukaszlenart:
@nikosdimitrakas what do you think about the latest changes?

nikosdimitrakas:
For me , it would have made more sense to handle the fragment identifier outside the QueryStringParser, as the fragment identifier is technically not part of the query string. That way you could keep the backwards compatibility. To avoid having to extract the fragment identifier in multiple places (even though it would only be a couple of lines) you could have some URLSplitter class with static methods like String getQueryString(String url) and String getFragmentIdentifier(String url). Then send the query string (and only the query string) to the QueryStringParser.
That said, I think you way will do the job.

lukaszlenart:
We can treat this as a step in a right direction, I would like to refactor the whole UrlProvider but to do this I need to throw away support for portlets.

PR #696, opened 2023-07-11
upgrade Felix Maven Bundle Plugin

hboutemy:
fixes last Reproducible Builds issue after #555
= https://issues.apache.org/jira/browse/FELIX-6602

see checks https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/struts/README.md

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -287,7 +287,7 @@
                 <plugin>
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
-                    <version>5.1.6</version>
+                    <version>5.1.9</version>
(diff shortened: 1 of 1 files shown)


lukaszlenart:
Could you create a ticket in JIRA to address this upgrade? Thanks in advance :)

hboutemy:
https://issues.apache.org/jira/browse/WW-5320 created and PR updated

lukaszlenart (approved):

lukaszlenart:
@hboutemy thanks for the ticket, yet I just noticed this is upgrade of a maven plugin and ticket was not needed, my bad, sorry :(

PR #715, opened 2023-07-30
[WW-5327] Stop using JavaBeans notation for setters in SecurityMemberAccess & MemberAccessValueStack

lukaszlenart:
Closes [WW-5327](https://issues.apache.org/jira/browse/WW-5327)
Closes #567

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -872,22 +872,22 @@ public class OgnlUtil {
         }
 
         SecurityMemberAccess memberAccess = new SecurityMemberAccess(allowStaticFieldAccess);
-        memberAccess.setDisallowProxyMemberAccess(disallowProxyMemberAccess);
+        memberAccess.disallowProxyMemberAccess(disallowProxyMemberAccess);
 
         if (devMode) {
             if (!warnReported.get()) {
                 warnReported.set(true);
                 LOG.warn("Working in devMode, using devMode excluded classes and packages!");
             }
-            memberAccess.setExcludedClasses(devModeExcludedClasses);
-            memberAccess.setExcludedPackageNamePatterns(devModeExcludedPackageNamePatterns);
-            memberAccess.setExcludedPackageNames(devModeExcludedPackageNames);
-            memberAccess.setExcludedPackageExemptClasses(devModeExcludedPackageExemptClasses);
+            memberAccess.useExcludedClasses(devModeExcludedClasses);
+            memberAccess.useExcludedPackageNamePatterns(devModeExcludedPackageNamePatterns);
+            memberAccess.useExcludedPackageNames(devModeExcludedPackageNames);
+            memberAccess.useExcludedPackageExemptClasses(devModeExcludedPackageExemptClasses);
         } else {
-            memberAccess.setExcludedClasses(excludedClasses);
-            memberAccess.setExcludedPackageNamePatterns(excludedPackageNamePatterns);
-            memberAccess.setExcludedPackageNames(excludedPackageNames);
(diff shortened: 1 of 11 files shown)


lukaszlenart:
@JCgH4164838Gh792C124B5 @gregh3269 @kusalk could you take a look?

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
```diff
@@ -482,12 +482,22 @@ public void clearContextValues() {
         ((OgnlContext) context).getValues().clear();
     }
 
+    @Deprecated
     public void setAcceptProperties(Set<Pattern> acceptedProperties) {
-        securityMemberAccess.setAcceptProperties(acceptedProperties);
+        throw new UnsupportedOperationException("Please use #useAcceptProperties instead!");
```
kusalk, line 487: Doesn't this defeat the purpose of deprecating and retaining these methods? May as well delete them if they're not functional

gregh3269:
Seems OK.  Running from the branch directly.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
lukaszlenart: My idea was to give a clear signal that these methods shouldn't be used anymore as they can degrade security. I will remove them in the [next version](https://issues.apache.org/jira/browse/WW-5328)

kusalk: I'd personally just delete them if that's the signal you want to give. Deprecating methods traditionally means they can still be used and will redirect to the replacing method where possible.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -194,7 +194,7 @@ private Set<Class<?>> parseClasses(String commaDelimitedClasses) {
             try {
                 classes.add(Class.forName(className));
             } catch (ClassNotFoundException e) {
-                throw new ConfigurationException("Cannot load class for exclusion/exemption configuration: " + className, e);
+                LOG.warn("Class: {} doesn't exist, ignoring it!", className);
```
kusalk, line 197: I would strongly advise against this - a ConfigurationException prevents the application from starting when an invalid exclusion list is provided (e.g. accidentally left out a comma or misspelt a class).

If we're going to change this to a warning, I think we need additional validation to prevent mistakes that lead to decreased security.

Alternatively, consider implementing an `optionalExcludedClasses` instead for classes that may not be present in every Struts application.

lukaszlenart: Fair point, I was wondering what was my intention when I had been introducing this change

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStack.java
lukaszlenart: I will keep them operational till the next version

kusalk (approved):

core/src/main/resources/struts-excluded-classes.xml
```diff
@@ -37,7 +37,9 @@
                 java.lang.ProcessBuilder,
                 java.lang.Thread,
                 sun.misc.Unsafe,
-                com.opensymphony.xwork2.ActionContext"/>
+                com.opensymphony.xwork2.ActionContext,
+                com.opensymphony.xwork2.ognl.SecurityMemberAccess,
```
kusalk, line 41: Latter 2 already excluded by `struts.excludedPackageNames`

lukaszlenart: Thanks, addressed in #729

PR #727, opened 2023-08-17
[WW-5331] Uses proper signature of get()

lukaszlenart:
Fixes [WW-5331](https://issues.apache.org/jira/browse/WW-5331)


core/src/main/java/org/apache/struts2/dispatcher/ApplicationMap.java
```diff
@@ -117,12 +117,15 @@ public Object setValue(final Object obj) {
      * @param key the entry key.
      * @return the servlet context attribute or init parameter or <tt>null</tt> if the entry is not found.
      */
-    public Object get(final String key) {
+    public Object get(final Object key) {
```
kusalk, line 120: Add Override annotation as per Sonar, otherwise LGTM

lukaszlenart: Done

PR #728, opened 2023-08-17
Defines a proper CODEOWNERS file

lukaszlenart:
[About code owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# Request PR review from any Apache Struts committer
+* @apache/struts-committers


lukaszlenart:
//cc: @apache/struts-committers

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Request PR review from any Apache Struts committer
+* @apache/struts-committers
```
kusalk, line 2: I think GitHub doesn't like the `*`

kusalk: Ah it might be complaining because it's a private team

lukaszlenart: Right... not sure why this team is private as all committers are publicly known

sepe81: But where exactly are all struts-committers defined as a group on GitHub? I couldn't find any reference.

lukaszlenart: I assume this is autogenerated group by ASF Infra base on the ASF records, yet I'm not sure which one is used for that as it doesn't include all the committers
https://people.apache.org/phonebook.html?unix=struts

kusalk: I see a list of 11 people - presumably the Struts committers who have made their Apache org membership public (private by default)?

lukaszlenart: As far I understand you must link your GH account to the ASF account to be added to the team, yet still there is a difference.

lukaszlenart: For reference https://issues.apache.org/jira/browse/INFRA-24905

PR #729, opened 2023-08-17
[WW-5327] Removes duplicated exclusion

lukaszlenart:
Refs [WW-5327](https://issues.apache.org/jira/browse/WW-5327)
Refs #715

### DIFF
--- a/core/src/main/resources/struts-excluded-classes.xml
+++ b/core/src/main/resources/struts-excluded-classes.xml
@@ -37,8 +37,7 @@
                 java.lang.ProcessBuilder,
                 java.lang.Thread,
                 sun.misc.Unsafe,
-                com.opensymphony.xwork2.ActionContext,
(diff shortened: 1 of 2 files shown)


core/src/main/resources/struts-excluded-classes.xml
```diff
@@ -38,8 +38,7 @@
                 java.lang.Thread,
                 sun.misc.Unsafe,
                 com.opensymphony.xwork2.ActionContext,
-                com.opensymphony.xwork2.ognl.SecurityMemberAccess,
-                com.opensymphony.xwork2.ognl.OgnlValueStack"/>
+                com.opensymphony.xwork2.ognl.SecurityMemberAccess"/>
```
kusalk, line 41: You missed removing this one :)

lukaszlenart: Good catch, thanks :)

kusalk (approved):

PR #730, opened 2023-08-17
Uses Java 17 to perform Code Quality check

lukaszlenart:
Addresses warning:
> The version of Java (11.0.20) you have used to run this analysis is deprecated and we will stop accepting it soon. Please update to at least Java 17.
Read more [here](https://docs.sonarcloud.io/appendices/scanner-environment/)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -49,7 +49,8 @@ jobs:
           distribution: adopt
           java-version: ${{ matrix.java }}
       - name: Build with Maven on Java ${{ matrix.java }}
-        run: mvn -B -V -DskipAssembly test --no-transfer-progress
+        if: matrix.java != '17'
+        run: mvn -B -V -DskipAssembly verify --no-transfer-progress
       - name: Code coverage on Java ${{ matrix.java }}
         if: matrix.java == '17'
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/maven.yml
```diff
@@ -49,10 +49,9 @@ jobs:
           distribution: adopt
           java-version: ${{ matrix.java }}
       - name: Build with Maven on Java ${{ matrix.java }}
-        if: matrix.java != '11'
```
kusalk, line 52: If we don't have `if: matrix.java != '17'`, we will be running the tests twice for JDK17

.github/workflows/maven.yml
```diff
@@ -49,10 +49,9 @@ jobs:
...
         run: mvn -B -V -DskipAssembly test --no-transfer-progress
```
kusalk, line 52: Shouldn't this be `verify` to ensure we run the integration tests for every JDK?

.github/workflows/maven.yml
lukaszlenart: 👍

.github/workflows/maven.yml
lukaszlenart: Right, will change that

kusalk (approved):

.github/workflows/maven.yml
lukaszlenart: This can lead to failed tests as quite often builds of different PRs are run on the same build agent and Jetty can not bind to port 8090 which already used. I will keep `test` phase for the JDK 8 & 11.

PR #731, opened 2023-08-17
WW-5334 Misc VelocityManager code cleanup

kusalk:
WW-5334
--

### DIFF
--- a/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
+++ b/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
@@ -47,12 +47,16 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
-import java.util.StringTokenizer;
+
+import static java.lang.String.format;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
+import static org.apache.struts2.views.util.ContextUtil.STRUTS;
 
 /**
  * Manages the environment for Velocity result types
@@ -60,8 +64,6 @@ import java.util.StringTokenizer;
 public class VelocityManager {
 
     private static final Logger LOG = LogManager.getLogger(VelocityManager.class);
-
-    public static final String STRUTS = "struts";
 
     private ObjectFactory objectFactory;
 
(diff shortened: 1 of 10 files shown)


plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -488,62 +468,18 @@ private void applyDefaultConfiguration(ServletContext context, Properties proper
...
         }
-
         properties.setProperty("userdirective", userdirective);
     }
 
-    private void addDirective(StringBuilder sb, Class<?> clazz) {
-        sb.append(clazz.getName()).append(",");
-    }
-
-    private static String replace(String string, String oldString, String newString) {
```
kusalk, line 516: This must be ancient code to be implementing String#replace from scratch

kusalk:
Hmm we need to rethink the Maven module setup to avoid cyclic dependencies

lukaszlenart:
> Hmm we need to rethink the Maven module setup to avoid cyclic dependencies. I'll create a separate PR addressing this.

What do you mean by that? I was planning to drop support for Portlets in Struts 7.x

kusalk:
> What do you mean by that?

Since `struts2-junit-plugin` depends on other plugins (such as `struts2-convention-plugin`, `struts2-portlet-plugin`, `struts2-velocity-plugin`), it's impossible to use `struts2-junit-plugin` within those plugins to write unit tests (it creates a cyclic dependency). We either need to move the unit tests for these affected plugins into their own modules (quite odd), or what would be better is to not have `struts2-junit-plugin` depend on any other plugins.

> I was planning to drop support for Portlets in Struts 7.x

That's fine, I'll try reorganise the modules slightly until then - otherwise I can't unit test `struts2-velocity-plugin`!

kusalk:
Will rebase and reopen after review and merge of #732

plugins/velocity/src/test/java/org/apache/struts2/views/velocity/result/VelocityResultTest.java
```diff
@@ -18,93 +18,100 @@
...
-import com.mockobjects.dynamic.Mock;
-import com.opensymphony.xwork2.ActionContext;
-import com.opensymphony.xwork2.ActionInvocation;
-import com.opensymphony.xwork2.ActionProxy;
-import com.opensymphony.xwork2.util.ValueStack;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.when;
 
+public class VelocityResultTest extends XWorkJUnit4TestCase {
```
kusalk, line 42: Converted this to JUnit4 and Mockito

plugins/velocity/src/test/java/org/apache/struts2/views/velocity/VelocityManagerTest.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.util.Properties;
+
+import static org.apache.struts2.views.velocity.VelocityManager.KEY_VELOCITY_STRUTS_CONTEXT;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
+
+public class VelocityManagerTest extends StrutsJUnit4TestCase {
```
kusalk, line 42: Doesn't cover everything but it's a start :)

lukaszlenart (approved):
LGTM 👍

PR #732, opened 2023-08-18
WW-5334 Enable usage of junit-plugin inside velocity-plugin

kusalk:
WW-5334
--
I extracted `StrutsPortletTestCase` into its own module from `struts-junit-plugin` to resolve the previous cyclic dependency (junit -> portlet -> velocity -> junit) which was preventing me from using `struts-junit-plugin` to write unit tests for `struts-velocity-plugin`.

See commit messages for other assorted clean up.

PR #733, opened 2023-08-20
WW-5334 Extract ConventionJUnit4Test into correct module

kusalk:
WW-5334
--
Remove last plugin dependency in JUnit plugin. Followup of #732.


lukaszlenart (approved):

lukaszlenart:
LGTM 👍

PR #735, opened 2023-08-21
WW-5336 Deprecate OgnlTool

kusalk:
WW-5336
--
I couldn't see any value in retaining OgnlTool as its own class so I've deprecated it and merged the method into StrutsUtil.

### DIFF
--- a/core/src/main/java/org/apache/struts2/util/StrutsUtil.java
+++ b/core/src/main/java/org/apache/struts2/util/StrutsUtil.java
@@ -18,14 +18,16 @@
  */
 package org.apache.struts2.util;
 
+import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.ObjectFactory;
+import com.opensymphony.xwork2.ognl.OgnlUtil;
 import com.opensymphony.xwork2.util.ClassLoaderUtil;
 import com.opensymphony.xwork2.util.TextParseUtil;
 import com.opensymphony.xwork2.util.ValueStack;
+import ognl.OgnlException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.struts2.StrutsException;
-import org.apache.struts2.views.jsp.ui.OgnlTool;
 import org.apache.struts2.views.util.UrlHelper;
 
 import javax.servlet.RequestDispatcher;
@@ -39,7 +41,16 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Map;
+
+import static java.text.MessageFormat.format;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
 
 /**
  * Struts base utility class, for use in Velocity and Freemarker templates
@@ -50,13 +61,12 @@ public class StrutsUtil {
 
     protected HttpServletRequest request;
     protected HttpServletResponse response;
-    protected Map<String, Class> classes = new Hashtable<>();
-    protected OgnlTool ognl;
+    protected Map<String, Class<?>> classes = new Hashtable<>();
     protected OgnlUtil ognl;
     protected ValueStack stack;
 
-    private UrlHelper urlHelper;
-    private ObjectFactory objectFactory;
+    private final UrlHelper urlHelper;
+    private final ObjectFactory objectFactory;
 
     public StrutsUtil(ValueStack stack, HttpServletRequest request, HttpServletResponse response) {
         this.stack = stack;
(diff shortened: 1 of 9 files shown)


kusalk:
Await merge of #731 first

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -50,32 +61,30 @@ public class StrutsUtil {
 
     protected HttpServletRequest request;
     protected HttpServletResponse response;
-    protected Map<String, Class> classes = new Hashtable<>();
-    protected OgnlTool ognl;
+    protected Map<String, Class<?>> classes = new HashMap<>();
```
kusalk, line 64: This class is constructed per request thread so I don't believe we need concurrency handling here

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -50,32 +61,30 @@ public class StrutsUtil {
...
-        String name = aName.toString();
-        Class c = classes.get(name);
-
-        if (c == null) {
-            c = ClassLoaderUtil.loadClass(name, StrutsUtil.class);
-            classes.put(name, c);
+    public Object bean(Object name) throws Exception {
+        String className = name.toString();
+        Class<?> clazz = classes.get(className);
+        if (clazz == null) {
```
kusalk, line 83: Using #computeIfAbsent here would change the exception handling semantics so I've forgone it

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -124,6 +123,17 @@ public Object findValue(String expression, String className) throws ClassNotFoun
         return stack.findValue(expression, Class.forName(className));
     }
 
+    public Object findValue(String expr, Object context) {
```
kusalk, line 126: Copied this here from OgnlTool

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -124,6 +123,17 @@ public Object findValue(String expression, String className) throws ClassNotFoun
...
+        try {
+            return ognl.getValue(expr, ActionContext.getContext().getContextMap(), context);
+        } catch (OgnlException e) {
+            if (e.getReason() instanceof SecurityException) {
+                LOG.error(format("Could not evaluate this expression due to security constraints: [{0}]", expr), e);
```
kusalk, line 131: Fixed broken logging

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -156,71 +166,64 @@ public String translateVariables(String expression) {
...
-                selectedItems = Arrays.asList((Object[]) i);
-            } else if (i instanceof Collection) {
-                selectedItems = (Collection) i;
-            } else {
-                // treat it is a single item
-                selectedItems = new ArrayList();
-                selectedItems.add(i);
-            }
-        }
+    public List<ListEntry> makeSelectList(String selectedList, String list, String listKey, String listValue) {
```
kusalk, line 169: Fixed cognitive complexity and readability

core/src/main/resources/struts-beans.xml
```diff
@@ -198,7 +198,7 @@
     <bean type="ognl.MethodAccessor" name="com.opensymphony.xwork2.util.CompoundRoot"
           class="com.opensymphony.xwork2.ognl.accessor.CompoundRootAccessor"/>
 
-    <bean class="org.apache.struts2.views.jsp.ui.OgnlTool"/>
+    <bean class="org.apache.struts2.views.jsp.ui.OgnlTool"/> <!-- Deprecated since 6.3.0 -->
```
sepe81, line 201: Maybe it's better to put the comment before the bean?

core/src/test/java/org/apache/struts2/util/StrutsUtilTest.java
```diff
@@ -145,28 +147,51 @@ public void testGetContextMethod() {
...
+    public void testMakeSelectList() {
+        String[] selectedList = new String[]{"Car", "Airplane", "Bus"};
+        List<String> list = Arrays.asList("Lorry", "Car", "Helicopter");
+
+        stack.getContext().put("mySelectedList", selectedList);
+        stack.getContext().put("myList", list);
+
+        List<ListEntry> listMade = strutsUtil.makeSelectList("#mySelectedList", "#myList", null, null);
+
+        assertEquals(listMade.size(), 3);
```
sepe81, line 159: for assertEquals, the expected value should be the first argument

sepe81 (changes requested):

core/src/main/resources/struts-beans.xml
kusalk: Amended - I tend to favour the same line as sometimes comments can become detached from their intended line in highly concurrent teams

core/src/test/java/org/apache/struts2/util/StrutsUtilTest.java
kusalk: Seems all the assertions in this class are mis-ordered, I've amended the ones I've touched :)

lukaszlenart (approved):
LGTM 👍

sepe81 (approved):

lukaszlenart:
Hope this is the last change and I can prepare another RC or even test build and call for vote :)

PR #736, opened 2023-08-21
WW-5337 Make SecurityMemberAccess exclusion checking more performant

kusalk:
WW-5337
--
See comments below


core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -218,14 +220,13 @@ protected void setDevModeExcludedPackageNamePatterns(String commaDelimitedPackag
...
-        Set<String> packagePatterns = TextParseUtil.commaDelimitedStringToSet(commaDelimitedPackagePatterns);
-        Set<Pattern> packageNamePatterns = new HashSet<>();
-
-        for (String pattern : packagePatterns) {
-            packageNamePatterns.add(Pattern.compile(pattern));
+        try {
+            return commaDelimitedStringToSet(commaDelimitedPackagePatterns)
+                    .stream().map(Pattern::compile).collect(toSet());
+        } catch (PatternSyntaxException e) {
+            throw new ConfigurationException(
```
kusalk, line 227: Ensure we throw ConfigurationException when RegEx patterns provided are invalid

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -261,7 +262,8 @@ public void setDevModeExcludedPackageExemptClasses(String commaDelimitedClasses)
     }
 
     private Set<String> parseExcludedPackageNames(String commaDelimitedPackageNames) {
-        Set<String> parsedSet = TextParseUtil.commaDelimitedStringToSet(commaDelimitedPackageNames);
+        Set<String> parsedSet = commaDelimitedStringToSet(commaDelimitedPackageNames)
+                .stream().map(s -> strip(s, ".")).collect(toSet());
```
kusalk, line 266: I think it's unintuitive to require a period at the of each excluded package. Currently if the period is missed, the exclusion will act like the regex `com\.package.*` which likely isn't what users configuring their application intend.

By stripping the period here, we retain backwards compatibility for excluded package lists that still have trailing periods.

However, we do drop support for wildcard package exclusions where applications genuinely desired the previous behaviour when a trailing period wasn't included. They will need to move such exclusions to the regex exclusion list instead.

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -237,33 +242,27 @@ protected String toPackageName(Class<?> clazz) {
...
+            String parentPackage = String.join(".", packageParts.subList(0, i + 1));
+            if (excludedPackageNames.contains(parentPackage)) {
                 return true;
             }
         }
         return false;
     }
 
     protected boolean isClassExcluded(Class<?> clazz) {
-        if (clazz == Object.class || (clazz == Class.class && !allowStaticFieldAccess)) {
```
kusalk, line 259: Moved this logic to #useExcludedClasses to make this method more performant

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -237,33 +242,27 @@ protected String toPackageName(Class<?> clazz) {
...
-            return true;
-        }
-        return excludedClasses.stream().anyMatch(clazz::isAssignableFrom);
+        return excludedClasses.contains(clazz);
     }
 
     protected boolean isExcludedPackageExempt(Class<?> clazz) {
-        return excludedPackageExemptClasses.stream().anyMatch(clazz::equals);
+        return excludedPackageExemptClasses.contains(clazz);
```
kusalk, line 265: Constant time operation

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -237,33 +242,27 @@ protected String toPackageName(Class<?> clazz) {
...
         }
         return false;
     }
 
     protected boolean isClassExcluded(Class<?> clazz) {
-        if (clazz == Object.class || (clazz == Class.class && !allowStaticFieldAccess)) {
-            return true;
-        }
-        return excludedClasses.stream().anyMatch(clazz::isAssignableFrom);
+        return excludedClasses.contains(clazz);
```
kusalk, line 261: Constant time operation BUT it changes the semantics, more on this in my comment on `SecurityMemberAccessTest`

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -237,33 +242,27 @@ protected String toPackageName(Class<?> clazz) {
...
+        return excludedPackageNamePatterns.stream().anyMatch(pattern -> pattern.matcher(packageName).matches());
     }
 
     protected boolean isExcludedPackageNames(Class<?> clazz) {
-        String suffixedPackageName = toPackageName(clazz) + ".";
-        for (String excludedPackageName : excludedPackageNames) {
-            if (suffixedPackageName.startsWith(excludedPackageName)) {
+        String packageName = toPackageName(clazz);
+        List<String> packageParts = Arrays.asList(packageName.split("\\."));
+        for (int i = 0; i < packageParts.size(); i++) {
```
kusalk, line 251: Operations proportional to number of package parts of class being checked. Previously it was proportional to number of excluded packages in configuration.

Note that we still retain the behaviour of sub-packages also being excluded.

core/src/main/resources/struts-excluded-classes.xml
```diff
@@ -55,52 +55,52 @@
...
-                org.objectweb.asm.,
-                org.springframework.context.,
-                com.opensymphony.xwork2.inject.,
-                com.opensymphony.xwork2.ognl.,
-                com.opensymphony.xwork2.security.,
-                com.opensymphony.xwork2.util.,
-                org.apache.tomcat.,
-                org.apache.catalina.core.,
-                org.wildfly.extension.undertow.deployment."/>
+                ognl,
```
kusalk, line 61: Whilst trailing periods are still compatible, I've updated this list to abide by the new convention

core/src/test/java/com/opensymphony/xwork2/ognl/SecurityMemberAccessTest.java
```diff
@@ -178,23 +178,6 @@ public void testMiddleOfInheritanceExclusion3() throws Exception {
         assertTrue("barLogic() from BarInterface isn't accessible!!!", accessible);
     }
 
-    @Test
-    public void testMiddleOfInheritanceExclusion4() throws Exception {
```
kusalk, line 182: Superinterfaces and superclasses of excluded classes are no longer also excluded. I'm not too sure of a real world example where this behaviour might actually be useful/sensible. Or where you wouldn't just add to the exclusion list the desired superinterfaces/superclasses instead?

If this behaviour is genuinely useful, it would be better if we moved such interfaces/classes to a separate exclusion list to keep the regular excluded class list performant.

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -328,11 +327,16 @@ public void useAcceptProperties(Set<Pattern> acceptedProperties) {
...
     @Deprecated
     public void setExcludedClasses(Set<Class<?>> excludedClasses) {
-        this.excludedClasses = excludedClasses;
+        useExcludedClasses(excludedClasses);
     }
 
     public void useExcludedClasses(Set<Class<?>> excludedClasses) {
-        this.excludedClasses = excludedClasses;
+        Set<Class<?>> newExcludedClasses = new HashSet<>(excludedClasses);
+        newExcludedClasses.add(Object.class);
```
kusalk, line 335: Moved this logic here from #isClassExcluded for performance reasons

core/src/test/java/com/opensymphony/xwork2/ognl/SecurityMemberAccessTest.java
lukaszlenart: I cannot recall the exact use case, yet we can try to relax this behaviour hopping it was just my paranoid mind ;)

lukaszlenart (approved):
LGTM 👍

PR #737, opened 2023-08-22
Drops duplicated dependency

lukaszlenart:
Closes #722


sepe81 (approved):
LGTM, mvn dependency:tree shows just `javax.servlet:javax.servlet-api:jar:3.1.0:provided`

PR #741, opened 2023-08-30
WW-5341 Ensure exclusion list applies to objects from all ClassLoaders

kusalk:
WW-5341
--
In applications where there are multiple classloaders, it may be possible for `SecurityMemberAccess` to obtain a false negative if the classloader used to load the target object was different to the one used to load the exclusion list.

To rectify this, we use String comparison of the class name instead. We still use the default classloader to validate the exclusion list on application start. A future enhancement might be to make the validating classloader configurable.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -111,6 +111,11 @@ public class SecurityMemberAccess implements MemberAccess {
             throw new IllegalArgumentException("Target does not match member!");
         }
 
+        if (disallowProxyMemberAccess && ProxyUtil.isProxyMember(member, target)) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -84,16 +86,17 @@ public Object setup(Map context, Object target, Member member, String propertyNa
 
     @Override
     public void restore(Map context, Object target, Member member, String propertyName, Object state) {
-        if (state != null) {
-            final AccessibleObject accessible = (AccessibleObject) member;
-            final boolean stateBoolean = ((Boolean) state).booleanValue();  // Using twice (avoid unboxing)
```
kusalk, line 89: Not sure what this comment meant

lukaszlenart (approved):
LGTM

PR #742, opened 2023-08-30
WW-5342 Add option to block use of default package

kusalk:
WW-5342
--


lukaszlenart:
Is it still in-progress?

lukaszlenart (approved):
LGTM 👍

PR #743, opened 2023-08-30
WW-5339 Enforce exclusion list in CompoundRootAccessor

kusalk:
WW-5339
--
TBA


kusalk:
Having done some further testing, I don't believe this mitigation to be necessary. There isn't anything inherently dangerous about just loading the class. Class members are still subject to `SecurityMemberAccess` checks and static fields and methods can be blocked with the following options.
```
struts.ognl.allowStaticFieldAccess=false
struts.ognl.allowStaticMethodAccess=false
```

PR #744, opened 2023-08-30
WW-5334 Fix empty chained context names in VelocityManager

kusalk:
WW-5334
--
Following #731, there are excessive log warnings arising from this. `Unable to instantiate chained VelocityContext`


kusalk:
I think we should merge this for `6.3.0`

lukaszlenart (approved):

PR #747, opened 2023-08-31
WW-5340 Introducing OGNL Guard

kusalk:
WW-5340
--
This serves as an optional, additional layer of protection to `SecurityMemberAccess`. `OgnlGuard` can validate both the raw and parsed OGNL expression. It is implemented as a user-configurable bean.

The default functionality includes the capability to block any expressions which contain specified OGNL AST nodes.


lukaszlenart:
Conflicts

core/src/main/java/com/opensymphony/xwork2/ognl/DefaultOgnlGuard.java
```diff
@@ -0,0 +1,98 @@
...
+import static com.opensymphony.xwork2.util.TextParseUtil.commaDelimitedStringToSet;
+import static java.util.Collections.emptySet;
+import static java.util.Collections.unmodifiableSet;
+
+/**
+ * The default implementation of {@link OgnlGuard}.
+ *
+ * @since 6.4.0
+ */
+public class DefaultOgnlGuard implements OgnlGuard {
```
lukaszlenart, line 40: I prefer to use `StrutsOgnlGuard` - default means different things for different ppl ;)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlGuard.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Guards all expressions parsed by Struts Core. It is evaluated by {@link OgnlUtil} immediately after parsing any
+ * expression.
+ *
+ * @since 6.4.0
+ */
+public interface OgnlGuard {
+
+    String GUARD_BLOCKED = "_ognl_guard_blocked";
```
lukaszlenart, line 32: Wouldn't be better to name this `EXPRESSION_BLOCKED`? This sounds like the guard itself has been blocked.

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -234,6 +234,8 @@ public final class StrutsConstants {
     /** The name of the parameter to determine whether static field access will be allowed in OGNL expressions or not */
     public static final String STRUTS_ALLOW_STATIC_FIELD_ACCESS = "struts.ognl.allowStaticFieldAccess";
 
+    public static final String STRUTS_OGNL_GUARD = "struts.ognlGuard";
```
lukaszlenart, line 237: What about `struts.ognl.guard` to keep everything in the same _namespace_?

core/src/main/resources/struts-beans.xml
```diff
@@ -166,6 +166,8 @@
           class="com.opensymphony.xwork2.validator.DefaultValidatorFileParser"/>
 
     <bean class="com.opensymphony.xwork2.ognl.OgnlUtil"/>
+    <bean type="com.opensymphony.xwork2.ognl.OgnlGuard" name="struts"
+          class="com.opensymphony.xwork2.ognl.StrutsOgnlGuard" />
```
lukaszlenart, line 170: Just missed that, I prefer to use `org.apache.struts2.ognl` namespace as `com.opensymphony.xwork2` should be removed at some point (in 7.x)

lukaszlenart (commented):
Sorry for review in steps, got distracted :)

core/src/main/java/org/apache/struts2/StrutsConstants.java
kusalk: Hmm, the other options that are in the `ognl.` namespace are configuration options for how OGNL executes. This on the other hand is an extension point for the bean/interface named `OgnlGuard`.

Up to you, I'm not fussed either way - let me know and I'll rename it :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlGuard.java
kusalk: Yeah agreed

core/src/main/java/org/apache/struts2/StrutsConstants.java
lukaszlenart: Ah... right :) We need to figure out a better naming at some point for the extension points, having that said I'm good :)

lukaszlenart:
Is it ready for review?

kusalk:
Yep that should be everything addressed

lukaszlenart (approved):
Nice 💪 LGTM 👍

PR #755, opened 2023-09-18
Split SonarCloud into separate action

kusalk:
Pull requests from forked repos cannot run the Sonar scan due to restricted secret access. To avoid a situation where we miss a failing test on JDK17, I've split the Sonar scan into its own job. It is also less confusing for contributors.

I've additionally replaced the `cache` step with the preset cache configuration in the `setup-java` step.


lukaszlenart (approved):
Nice :) LGTM 👍

PR #758, opened 2023-10-04
[WW-5338] Removes deprecated OgnlTool

lukaszlenart:
Closes [WW-5338](https://issues.apache.org/jira/browse/WW-5338)


core/src/main/java/org/apache/struts2/util/StrutsUtil.java
```diff
@@ -125,12 +120,10 @@ public Object findValue(String expression, String className) throws ClassNotFoun
 
     public Object findValue(String expr, Object context) {
         try {
-            return ognl.getValue(expr, ActionContext.getContext().getContextMap(), context);
-        } catch (OgnlException e) {
-            if (e.getReason() instanceof SecurityException) {
-                LOG.error(format("Could not evaluate this expression due to security constraints: [{0}]", expr), e);
-            }
-            return null;
+            stack.push(context);
```
kusalk, line 123: Should be outside the try block so that we don't pop unless successfully pushed

kusalk (commented):
We can also delete `org.apache.struts2.views.util.ContextUtil#OGNL`

core/src/main/java/org/apache/struts2/util/StrutsUtil.java
lukaszlenart: fixed

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
```diff
@@ -56,8 +56,7 @@
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.joining;
-import static java.util.stream.Collectors.toList;
-import static org.apache.struts2.views.util.ContextUtil.OGNL;
+import static java.util.stream.Collectors.toList;;
```
kusalk, line 59: Double semicolon 😁

kusalk (approved):
LGTM

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManager.java
lukaszlenart: 😵‍💫

PR #760, opened 2023-10-05
WW-5340 Mild refactor StrutsOgnlGuard for easier subclassing

kusalk:
WW-5340
--
Quick follow-up to https://github.com/apache/struts/pull/747
See comments

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
@@ -142,26 +142,22 @@ public class JakartaMultiPartRequest extends AbstractMultiPartRequest {
             }
 
             long size = item.getSize();
-            if (size == 0) {
-                values.add(StringUtils.EMPTY);
-            } else if (size > maxStringLength) {
+            if (size > maxStringLength) {
+                LOG.debug("Form field {} of size {} bytes exceeds limit of {}.", item.getFieldName(), size, maxStringLength);
                 String errorKey = "struts.messages.upload.error.parameter.too.long";
                 LocalizedMessage localizedMessage = new LocalizedMessage(this.getClass(), errorKey, null,
-                    new Object[]{item.getFieldName(), maxStringLength, size});
-
+                        new Object[]{item.getFieldName(), maxStringLength, size});
                 if (!errors.contains(localizedMessage)) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
```diff
@@ -142,26 +142,22 @@ protected void processNormalFormField(FileItem item, String charset) throws Unsu
             }
 
             long size = item.getSize();
-            if (size == 0) {
-                values.add(StringUtils.EMPTY);
-            } else if (size > maxStringLength) {
+            if (size > maxStringLength) {
+                LOG.debug("Form field {} of size {} bytes exceeds limit of {}.", item.getFieldName(), size, maxStringLength);
```
kusalk, line 146: Just added this line, no other functional changes here

core/src/main/java/org/apache/struts2/ognl/StrutsOgnlGuard.java
```diff
@@ -71,28 +71,38 @@ public boolean isRawExpressionBlocked(String expr) {
...
+        for (int i = 0; i < node.jjtGetNumChildren(); i++) {
+            if (recurseNodes(node.jjtGetChild(i))) {
+                return true;
+            }
         }
-        return recurseExcludedNodeType((Node) tree);
+        return false;
+    }
+
+    protected boolean checkNode(Node node) {
```
kusalk, line 96: I separated the recursion logic from the node checking logic so that subclasses don't need to unnecessarily duplicate that code when overriding.

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
kusalk: I think logging the field name at the debug level is fine in terms of security?

lukaszlenart: I assume so, yet this is a user controlled value so maybe it needs escaping

kusalk: Not sure a field name can even contain newline characters but anyway SonarCloud is happy now

lukaszlenart (approved):

PR #763, opened 2023-10-09
WW-5349 Remove Struts core dependency on OGNL VarRefs

kusalk:
WW-5349
--
For applications that don't use components such as Action and Set, or templates which leverage OGNL variable references, `#attr.templateDir` and `#attr.theme` are the only 2 dependencies on this OGNL capability.

By removing this functionality, applications can block `ognl.ASTVarRef` using `struts.ognl.excludedNodeTypes` for strengthened security. This is a commonly used gadget by attackers.

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/UIBean.java
+++ b/core/src/main/java/org/apache/struts2/components/UIBean.java
@@ -601,12 +601,6 @@ public abstract class UIBean extends Component {
             result = findString(this.templateDir);
         }
 
-        // If templateDir is not explicitly given,
-        // try to find attribute which states the dir set to use
-        if (StringUtils.isBlank(result)) {
-            result = stack.findString("#attr.templateDir");
-        }
-
         // Default template set
         if (StringUtils.isBlank(result)) {
             result = defaultTemplateDir;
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Wouldn't make sense to exclude `ognl.ASTVarRef` by default?

kusalk:
I think many applications make use of the VarRef functionality (including any application that uses Action or Set Struts components) so no it wouldn't make sense to block by default.

But by removing this dependency in UIBean, for applications that don't use this functionality, they can block it completely.

I've got a list of other nodes that we probably can block by default (which I'm still compiling) - I think we can introduce that in 7.0.

lukaszlenart:
> I've got a list of other nodes that we probably can block by default (which I'm still compiling) - I think we can introduce that in 7.0.

Nice, could you register a ticket to not forget about that?

is it still a draft?

kusalk:
Yeah still experimenting a bit. But essentially we should castrate OGNL as much as possible without hindering popular features.

Here is the card for 7.0: WW-5353

lukaszlenart (approved):

lukaszlenart:
This change broke this functionality https://struts.apache.org/tag-developers/selecting-template-directory and https://issues.apache.org/jira/browse/WW-5516

PR #766, opened 2023-10-13
WW-5355 Use LRU cache by default

kusalk:
WW-5355
--
I see very limited advantage to using the default cache over the LRU one? The eviction process is faster but at the cost of having to repopulate the whole cache from scratch!
Keen to hear other thoughts or anything I may have missed

Perhaps in the future we can move to a more advanced algorithm such as the ones provided by [Caffeine](https://github.com/ben-manes/caffeine/wiki/Efficiency), although I'm not sure how effective they'd be given the cache can be easily attacked.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlDefaultCache.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlDefaultCache.java
@@ -30,10 +30,10 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class OgnlDefaultCache<Key, Value> implements OgnlCache<Key, Value> {
 
     private final ConcurrentHashMap<Key, Value> ognlCache;
-    private final AtomicInteger cacheEvictionLimit = new AtomicInteger(25000);
+    private final AtomicInteger cacheEvictionLimit = new AtomicInteger();
 
     public OgnlDefaultCache(int evictionLimit, int initialCapacity, float loadFactor) {
-        this.cacheEvictionLimit.set(evictionLimit);
+        cacheEvictionLimit.set(evictionLimit);
         ognlCache = new ConcurrentHashMap<>(initialCapacity, loadFactor);
     }
 
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlLRUCache.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlLRUCache.java
@@ -36,15 +36,15 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class OgnlLRUCache<Key, Value> implements OgnlCache<Key, Value> {
 
     private final Map<Key, Value> ognlLRUCache;
-    private final AtomicInteger cacheEvictionLimit = new AtomicInteger(2500);
+    private final AtomicInteger cacheEvictionLimit = new AtomicInteger();
 
     public OgnlLRUCache(int evictionLimit, int initialCapacity, float loadFactor) {
-        this.cacheEvictionLimit.set(evictionLimit);
+        cacheEvictionLimit.set(evictionLimit);
         // Access-order mode selected (order mode true in LinkedHashMap constructor).
         ognlLRUCache = Collections.synchronizedMap (new LinkedHashMap<Key, Value>(initialCapacity, loadFactor, true) {
             @Override
             protected boolean removeEldestEntry(Map.Entry<Key,Value> eldest) {
-                return (this.size() > cacheEvictionLimit.get());
+                return size() > cacheEvictionLimit.get();
             }
         });
     }
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -673,9 +673,15 @@ public class OgnlUtil {
      *                   note if exclusions AND inclusions are supplied and not null nothing will get copied.
      * @param editable the class (or interface) to restrict property setting to
      */
-    public void copy(final Object from, final Object to, final Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions, Class<?> editable) {
+    public void copy(final Object from,
+                     final Object to,
+                     final Map<String, Object> context,
+                     Collection<String> exclusions,
+                     Collection<String> inclusions,
+                     Class<?> editable) {
         if (from == null || to == null) {
-            LOG.warn("Attempting to copy from or to a null source. This is illegal and is bein skipped. This may be due to an error in an OGNL expression, action chaining, or some other event.");
+            LOG.warn(
+                    "Skipping attempt to copy from, or to, a null source.", new RuntimeException());
             return;
         }
 
@@ -689,8 +695,7 @@ public class OgnlUtil {
             fromPds = getPropertyDescriptors(from);
             if (editable != null) {
                 toPds = getPropertyDescriptors(editable);
-            }
-            else {
+            } else {
                 toPds = getPropertyDescriptors(to);
             }
         } catch (IntrospectionException e) {
@@ -705,29 +710,31 @@ public class OgnlUtil {
         }
 
         for (PropertyDescriptor fromPd : fromPds) {
-            if (fromPd.getReadMethod() != null) {
-                boolean copy = true;
-                if (exclusions != null && exclusions.contains(fromPd.getName())) {
-                    copy = false;
-                } else if (inclusions != null && !inclusions.contains(fromPd.getName())) {
-                    copy = false;
-                }
-
-                if (copy) {
-                    PropertyDescriptor toPd = toPdHash.get(fromPd.getName());
-                    if ((toPd != null) && (toPd.getWriteMethod() != null)) {
-                        try {
-                            Object value = ognlGet(fromPd.getName(), contextFrom, from, null, context, this::checkEnableEvalExpression);
-                            ognlSet(fromPd.getName(), contextTo, to, value, context);
-                        } catch (OgnlException e) {
-                            LOG.debug("Got OGNL exception", e);
-                        }
-                    }
-
-                }
-
+            if (fromPd.getReadMethod() == null) {
+                continue;
             }
 
+            if (exclusions != null && exclusions.contains(fromPd.getName()) ||
+                    inclusions != null && !inclusions.contains(fromPd.getName())) {
+                continue;
+            }
+
+            PropertyDescriptor toPd = toPdHash.get(fromPd.getName());
+            if (toPd == null || toPd.getWriteMethod() == null) {
+                continue;
+            }
+
+            try {
+                Object value = ognlGet(fromPd.getName(),
+                        contextFrom,
+                        from,
+                        null,
+                        context,
+                        this::checkEnableEvalExpression);
+                ognlSet(fromPd.getName(), contextTo, to, value, context);
+            } catch (OgnlException e) {
+                LOG.debug("Got OGNL exception", e);
+            }
         }
     }
 
@@ -746,7 +753,7 @@ public class OgnlUtil {
     }
 
     /**
-     * Get's the java beans property descriptors for the given source.
+     * Gets the java beans property descriptors for the given source.
      *
      * @param source the source object.
      * @return property descriptors.
--- a/core/src/main/resources/org/apache/struts2/default.properties
+++ b/core/src/main/resources/org/apache/struts2/default.properties
@@ -243,12 +243,12 @@ struts.ognl.enableExpressionCache=true
 ### For expressionCacheLRUMode true, the limit will ensure the cache does not exceed
 ### that size, dropping the oldest (least-recently-used) expressions to add new ones.
 ### NOTE: If not set, the default is 25000, which may be excessive.
-# struts.ognl.expressionCacheMaxSize=1000
+struts.ognl.expressionCacheMaxSize=10000
 
 ### Indicates if the OGNL expressionCache should use LRU mode.
 ### NOTE: When true, make sure to set the expressionCacheMaxSize to a reasonable value
 ###       for your application.  Otherwise the default limit will never (practically) be reached.
-# struts.ognl.expressionCacheLRUMode=false
+struts.ognl.expressionCacheLRUMode=true
 
 ### Specify a limit to the number of entries in the OGNL beanInfoCache.
 ### For the standard beanInfoCache mode, when the limit is exceeded the entire cache's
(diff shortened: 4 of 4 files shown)


kusalk:
@JCgH4164838Gh792C124B5

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -705,29 +710,31 @@ public void copy(final Object from, final Object to, final Map<String, Object> c
         }
 
         for (PropertyDescriptor fromPd : fromPds) {
```
kusalk, line 712: Nothing to see here, just made this code more digestible

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlDefaultCache.java
```diff
@@ -30,10 +30,10 @@
 public class OgnlDefaultCache<Key, Value> implements OgnlCache<Key, Value> {
 
     private final ConcurrentHashMap<Key, Value> ognlCache;
-    private final AtomicInteger cacheEvictionLimit = new AtomicInteger(25000);
+    private final AtomicInteger cacheEvictionLimit = new AtomicInteger();
```
kusalk, line 33: These init values are inconsequential as far as I can tell as they are immediately overridden by the constructor

ben-manes:
fwiw, this would change from a concurrent cache to a synchronized one. It looks like the original code replaced an unbounded concurrent map. The performance difference may be unacceptable of explain the concern without benchmarks/profiling.

Caffeine is very resilient to attacks, like hash flooding and being scan resilient. An independent analysis might be of interest, [An evaluation of cache management policies under workloads with malicious requests](https://ieeexplore.ieee.org/abstract/document/8247467/).

kusalk:
Good call out @ben-manes, I completely glossed over the change from concurrent map to synchronised map (and its performance implications).

And thank you for that analysis on resistance to attacks.

In that case I may just go ahead and integrate Caffeine as it seems rather straightforward anyway.

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -243,25 +243,25 @@ struts.ognl.enableExpressionCache=true
...
 ### that size, dropping the oldest (least-recently-used) expressions to add new ones.
 ### NOTE: If not set, the default is 25000, which may be excessive.
-# struts.ognl.expressionCacheMaxSize=1000
+struts.ognl.expressionCacheMaxSize=10000
 
 ### Indicates if the OGNL expressionCache should use LRU mode.
 ### NOTE: When true, make sure to set the expressionCacheMaxSize to a reasonable value
 ###       for your application.  Otherwise the default limit will never (practically) be reached.
-# struts.ognl.expressionCacheLRUMode=false
+struts.ognl.expressionCacheLRUMode=true
```
JCgH4164838Gh792C124B5, line 251: @ben-manes already pointed out potential implications in making the LRU mode cache active by default (switching from a concurrent to synchronized cache).  It might be safer to keep LRU mode caching off by default, at least until someone can confirm if performance of the LRU mode cache is acceptable under load (benchmarking).

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -243,25 +243,25 @@ struts.ognl.enableExpressionCache=true
...
 
 ### Specify a limit to the number of entries in the OGNL beanInfoCache.
 ### For the standard beanInfoCache mode, when the limit is exceeded the entire cache's
 ### content will be cleared (can help prevent memory leaks).
 ### For beanInfoCacheLRUMode true, the limit will ensure the cache does not exceed
 ### that size, dropping the oldest (least-recently-used) expressions to add new ones.
 ### NOTE: If not set, the default is 25000, which may be excessive.
-# struts.ognl.beanInfoCacheMaxSize=1000
+struts.ognl.beanInfoCacheMaxSize=5000
```
JCgH4164838Gh792C124B5, line 259: Above comment for numeric defaults applies here as well.

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -243,25 +243,25 @@ struts.ognl.enableExpressionCache=true
...
 
 ### Indicates if the OGNL beanInfoCache should use LRU mode.
 ### NOTE: When true, make sure to set the beanInfoCacheMaxSize to a reasonable value
 ###       for your application.  Otherwise the default limit will never (practically) be reached.
-# struts.ognl.beanInfoCacheLRUMode=false
+struts.ognl.beanInfoCacheLRUMode=true
```
JCgH4164838Gh792C124B5, line 264: Above comment on LRU mode defaults applies here as well.

JCgH4164838Gh792C124B5 (commented):
Hi @kusalk.  Thanks for reaching out to invite comments on the potential cache implementation changes.  😄 

Switching to LRU mode by default might be a bit risky performance-wise, until someone benchmarks performance with it on/off.

In terms of potentially integrating Caffeine to handle caching, that sounds interesting.  Maybe making it another option (with a wrapper implementing the OgnlCache interface) would be one way to do so ?

In that way, I think that might keep the library dependency as optional, depending on what cache implementation was chosen via configuration (even if a Caffeine-based cache was to become the default cache implementation for the framework).  Others might have different thoughts on the matter ...

Regards.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlDefaultCache.java
kusalk: The overriding is a fair point - I'll move the initialisation of the whole field into the constructor

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -673,9 +673,15 @@ public void copy(final Object from, final Object to, final Map<String, Object> c
...
-    public void copy(final Object from, final Object to, final Map<String, Object> context, Collection<String> exclusions, Collection<String> inclusions, Class<?> editable) {
+    public void copy(final Object from,
+                     final Object to,
+                     final Map<String, Object> context,
+                     Collection<String> exclusions,
+                     Collection<String> inclusions,
+                     Class<?> editable) {
         if (from == null || to == null) {
-            LOG.warn("Attempting to copy from or to a null source. This is illegal and is bein skipped. This may be due to an error in an OGNL expression, action chaining, or some other event.");
+            LOG.warn(
```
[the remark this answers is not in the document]
kusalk: Currently, this method is only called by `ChainingInterceptor` within Struts, and from what I can tell, these parameters will never be null. Although it is also a public method. Calling this method with null values is improper usage and the stack trace will provide better context for such an issue to be resolved. The previous warning message wasn't very helpful.

core/src/main/resources/org/apache/struts2/default.properties
```diff
@@ -243,25 +243,25 @@ struts.ognl.enableExpressionCache=true
 ### For expressionCacheLRUMode true, the limit will ensure the cache does not exceed
 ### that size, dropping the oldest (least-recently-used) expressions to add new ones.
 ### NOTE: If not set, the default is 25000, which may be excessive.
-# struts.ognl.expressionCacheMaxSize=1000
+struts.ognl.expressionCacheMaxSize=10000
```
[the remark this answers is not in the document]
kusalk: Yeah I agree that it would be better to have the defaults in the code and properties file be consistent (as well as the accompanying comments in the properties file). I'll amend this.

kusalk:
Thank you for the well considered review @JCgH4164838Gh792C124B5 

Yep I no longer intend to make LRU (backed by a synchronised LinkedHashMap) the default given the performance implications. And yes I intend to keep the cache implementation configurable, even with the implementation of Caffeine.

I think choosing a default cache size will always be a challenge as it is subject to multiple application-specific factors so the best we can do here is to encourage developers to adjust the cache size as needed.

kusalk:
I found that previously, `struts.ognl.expressionCacheLRUMode=true` was not reliable due to setter injection not guaranteeing the order of operations. I've thus refactored the code to use constructor injection.

core/src/main/resources/org/apache/struts2/default.properties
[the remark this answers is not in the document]
kusalk: I ended up defining the defaults in the properties file and removing defaults from the code - except for required values when bootstrapping the container. (And except for a deprecated constructor which is awaiting deletion)

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -125,6 +127,24 @@
  */
 public class DefaultConfiguration implements Configuration {
 
+    public static final Map<String, Object> BOOTSTRAP_CONSTANTS;
```
kusalk, line 130: Extracted bootstrap constants into a public static final field for reuse by the default Struts configuration provider and any test code as needed

core/src/main/java/com/opensymphony/xwork2/ognl/DefaultOgnlCacheFactory.java
```diff
@@ -15,52 +15,86 @@
...
+ * <p>Currently used for Expression cache and BeanInfo cache creation.</p>
  *
- * @param <Key> The type for the cache key entries
+ * @param <Key>   The type for the cache key entries
  * @param <Value> The type for the cache value entries
  */
 public class DefaultOgnlCacheFactory<Key, Value> implements OgnlCacheFactory<Key, Value> {
 
-    private final AtomicBoolean useLRUCache = new AtomicBoolean(false);
-    private final AtomicInteger cacheMaxSize = new AtomicInteger(25000);
```
kusalk, line 33: Making this factory class thread-safe seemed unnecessary

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlLRUCache.java
```diff
@@ -81,7 +81,9 @@ public int getEvictionLimit() {
 
     @Override
     public void setEvictionLimit(int cacheEvictionLimit) {
+        if (cacheEvictionLimit < size()) {
```
kusalk, line 84: Fixed this bug

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -143,12 +143,18 @@ protected void setEnableExpressionCache(String cache) {
         enableExpressionCache = BooleanUtils.toBoolean(cache);
     }
 
-    @Inject(value = StrutsConstants.STRUTS_OGNL_EXPRESSION_CACHE_MAXSIZE, required = false)
```
kusalk, line 146: Doesn't make much sense to inject this both in the factory and here. Plus, Caffeine doesn't support changing the max size after construction (and neither does LRU really), so we'd have to rebuild the cache to continue supporting this method. I can't see much use for it anyway so I've just gone ahead and deprecated it.

kusalk:
I've taken care to maintain full API compatibility given that these classes are extension points and thus public API

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
ben-manes:
```java
cache.policy().eviction().ifPresent(eviction -> {
  eviction.setMaximum(2 * eviction.getMaximum());
});
```

The [Policy](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Policy.html) api contains various methods for pragmatic needs, while keeping the core apis simple and focused on the common needs.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCaffeineCache.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void put(K key, V value) {
+        cache.put(key, value);
+    }
+
+    @Override
+    public void putIfAbsent(K key, V value) {
+        if (cache.getIfPresent(key) == null) {
+            cache.put(key, value);
+        }
```
ben-manes, line 60: `cache.asMap().putIfAbsent(key, value)`

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCaffeineCache.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public int size() {
+        return Math.toIntExact(cache.estimatedSize());
```
ben-manes, line 65: `cache.asMap().size()`

kusalk:
Thanks @ben-manes, amended!

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCacheFactory.java
```diff
@@ -19,12 +19,52 @@
...
+    default boolean getUseLRUCache() {
+        return CacheType.SYNC_LINKED_LRU.equals(getDefaultCacheType());
+    }
+
+    CacheType getDefaultCacheType();
+
+    enum CacheType {
+        CONCURRENT_BASIC,
+        SYNC_LINKED_LRU,
+        CAFFEINE_WTLFU
```
lukaszlenart, line 68: Are these different implementations of cache layer? Shouldn't this go through the extension point?

lukaszlenart (commented):
Could you adjust a ticket's description because right now I'm a bit confused what's the scope of the ticket.

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCacheFactory.java
kusalk: I was wondering the same thing - I was only maintaining the design introduced in #528.

I simply replaced `struts.ognl.expressionCacheLRUMode=true|false` with `struts.ognl.expressionCacheType=concurrent_basic|sync_linked_lru|caffeine_wtlfu`

Let me look into refactoring this, it will likely be a breaking change though.

kusalk: Yeah given the breaking nature, I'm more inclined to take on this refactor as part of 7.0, all good for me to create a follow-up card for it?

kusalk:
@lukaszlenart I've updated the PR description :)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCacheFactory.java
lukaszlenart: Thanks! That would make more sense and then we can remove all the properties and leave that to a given implementation. Please create the ticket :)

lukaszlenart (approved):

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlCacheFactory.java
kusalk: [WW-5356](https://issues.apache.org/jira/browse/WW-5356)

PR #773, opened 2023-10-23
Improved the StrutsUrlDecoder so that charset retrieval is performed only once.

mygreen:
- StutsUrlDecoder has performance problems when run from multiple threads, because [Charset#availavleCharts(...)](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/nio/charset/Charset.java#L569-L588) called `StrutsUrlDecoder#getCharset(...)` is synchronized blocked.
  - Root causes is synchronized blocked at [AbstractCharsetProvider#charsets()](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/jdk.charsets/share/classes/sun/nio/cs/ext/AbstractCharsetProvider.java#L174-L177) .
  - StutsUrlDecoder is used to evaluate custom tag [<s:url>](https://struts.apache.org/tag-developers/url-tag), it will have performance problems if it receives too many requests.
- Therefore, `Charset#availavleCharts(...)` to be executed only once during the StrutsUrlDecoder initialization.

### DIFF
--- a/core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
+++ b/core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
@@ -33,7 +33,7 @@ public class StrutsUrlDecoder implements UrlDecoder {
 
     private static final Logger LOG = LogManager.getLogger(StrutsUrlDecoder.class);
 
-    private static final Collection<Charset> AVAILAVLE_CHARSETS = Charset.availableCharsets().values();
+    private static final Collection<Charset> AVAILABLE_CHARSETS = Charset.availableCharsets().values();
 
     private String encoding = "UTF-8";
 
--- a/core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
+++ b/core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
@@ -110,7 +110,7 @@ public class StrutsUrlDecoder implements UrlDecoder {
     }
 
     private Charset getCharset(String encoding) throws UnsupportedEncodingException {
-        for (Charset charset : AVAILAVLE_CHARSETS) {
+        for (Charset charset : AVAILABLE_CHARSETS) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
```diff
@@ -27,11 +27,14 @@
 
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
+import java.util.Collection;
 
 public class StrutsUrlDecoder implements UrlDecoder {
 
     private static final Logger LOG = LogManager.getLogger(StrutsUrlDecoder.class);
 
+    private static final Collection<Charset> AVAILAVLE_CHARSETS = Charset.availableCharsets().values();
```
sepe81, line 36:
```suggestion
    private static final Collection<Charset> AVAILABLE_CHARSETS = Charset.availableCharsets().values();
```

core/src/main/java/org/apache/struts2/url/StrutsUrlDecoder.java
```diff
@@ -107,7 +110,7 @@ private byte convertHexDigit(byte b) {
     }
 
     private Charset getCharset(String encoding) throws UnsupportedEncodingException {
-        for (Charset charset : Charset.availableCharsets().values()) {
+        for (Charset charset : AVAILAVLE_CHARSETS) {
```
sepe81, line 113:
```suggestion
        for (Charset charset : AVAILABLE_CHARSETS) {
```

sepe81 (changes requested):
LGTM otherwise

lukaszlenart (approved):
👏

lukaszlenart:
Would you mind opening a [JIRA](https://issues.apache.org/jira/projects/WW) ticket and link it here?

mygreen:
The fllowing JIRA tikcet was created.

https://issues.apache.org/jira/browse/WW-5359

PR #779, opened 2023-11-04
[WW-5333] Refactors AttributeMap

lukaszlenart:
Closes [WW-5333](https://issues.apache.org/jira/browse/WW-5333)

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/Set.java
+++ b/core/src/main/java/org/apache/struts2/components/Set.java
@@ -20,6 +20,7 @@ package org.apache.struts2.components;
 
 import java.io.Writer;
 
+import org.apache.struts2.dispatcher.DispatcherConstants;
 import org.apache.struts2.views.annotations.StrutsTag;
 import org.apache.struts2.views.annotations.StrutsTagAttribute;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/struts2/dispatcher/DispatcherScope.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.dispatcher;
+
+public interface DispatcherScope {
```
kusalk, line 21: Enum instead?

kusalk (approved):

core/src/main/java/org/apache/struts2/dispatcher/DispatcherScope.java
lukaszlenart: I started with enum but then I noticed this is used as string only, let me re-think this approach

lukaszlenart: Enum doesn't make sense, imho. I renamed the class to clearly express that these are simple constants.

PR #780, opened 2023-11-05
WW-5350 Refactor SecurityMemberAccess

kusalk:
WW-5350
--
See comments.

I didn't deprecate methods for which I've changed the signature as from what I can tell `SecurityMemberAccess` is not considered public API (no reasonable situation in which a Struts applications would need to override or instantiate the class themselves, nor is it a defined extensible bean) (WW-5343 aims to make it a configurable bean and thus public API).

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -26,6 +26,7 @@ import org.apache.logging.log4j.Logger;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Member;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
     public boolean isAccessible(Map context, Object target, Member member, String propertyName) {
         LOG.debug("Checking access for [target: {}, member: {}, property: {}]", target, member, propertyName);
 
-        final int memberModifiers = member.getModifiers();
-        final Class<?> memberClass = member.getDeclaringClass();
-        // target can be null in case of accessing static fields, since OGNL 3.2.8
-        final Class<?> targetClass = Modifier.isStatic(memberModifiers) ? memberClass : target.getClass();
```
kusalk, line 110: Wanted to be a bit more deliberate about how we handle the arguments in case OGNL behaviour changes

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
-        if (!memberClass.isAssignableFrom(targetClass)) {
-            throw new IllegalArgumentException("Target does not match member!");
+        if (target instanceof Class) { // Target may be of type Class for static members
+            if (!member.getDeclaringClass().equals(target)) {
+                throw new IllegalArgumentException("Target class does not match static member!");
```
kusalk, line 110: Throw an exception here because if this ceases to be the case, it means the following logic may need changes

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
+            }
+            target = null;
+        } else {
+            if (target != null && !member.getDeclaringClass().isAssignableFrom(target.getClass())) {
+                throw new IllegalArgumentException("Target does not match member!");
+            }
         }
 
-        if (disallowProxyMemberAccess && ProxyUtil.isProxyMember(member, target)) {
```
kusalk, line 115: Extract this into its own protected method for consistency

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
             return false;
         }
 
-        if (!checkPublicMemberAccess(memberModifiers)) {
+        if (!checkPublicMemberAccess(member)) {
             LOG.warn("Access to non-public [{}] is blocked!", member);
             return false;
         }
 
-        if (!checkStaticFieldAccess(member, memberModifiers)) {
```
kusalk, line 125: Removed redundant argument

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
         }
 
-        if (disallowProxyMemberAccess && ProxyUtil.isProxyMember(member, target)) {
-            LOG.warn("Access to proxy is blocked! Target class [{}] of target [{}], member [{}]", targetClass, target, member);
+        if (!checkProxyMemberAccess(target, member)) {
+            LOG.warn("Access to proxy is blocked! Member class [{}] of target [{}], member [{}]", member.getDeclaringClass(), target, member);
             return false;
         }
 
-        if (!checkPublicMemberAccess(memberModifiers)) {
```
kusalk, line 120: Tried to standardise around just passing the object and/or member to these protected methods

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
         }
 
-        if (isPackageExcluded(targetClass)) {
+        return true;
+    }
+
+    /**
+     * @return {@code true} if member access is allowed
+     */
+    protected boolean checkExclusionList(Object target, Member member) {
```
kusalk, line 157: Extracted all the exclusion list checks into this method

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
             return false;
         }
 
-        if (targetClass != memberClass && isClassExcluded(targetClass)) {
-            LOG.warn("Target class [{}] of target [{}] is excluded!", targetClass, target);
+        if (!checkExclusionList(target, member)) {
             return false;
         }
 
-        if (disallowDefaultPackageAccess) {
```
kusalk, line 151: Extracted this into its own method

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
             return false;
         }
 
-        if (!checkStaticFieldAccess(member, memberModifiers)) {
+        if (!checkStaticFieldAccess(member)) {
             LOG.warn("Access to static field [{}] is blocked!", member);
             return false;
         }
 
-        // it needs to be before calling #checkStaticMethodAccess()
```
kusalk, line 130: This is essentially an exemption to `checkStaticMethodAccess` so I moved it within that method - we shouldn't be relying on method call order here

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
-        return false;
+    /**
+     * @return {@code true} if member access is allowed
+     */
+    protected boolean checkEnumAccess(Member member) {
+        return member.getDeclaringClass().isEnum()
+                && isStatic(member)
+                && member instanceof Method
+                && member.getName().equals("values")
+                && ((Method) member).getParameterCount() == 0;
```
kusalk, line 265: Closed a minor hole where another static `values` method could also be exempted

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
-     * @param memberModifiers
-     * @return
+     * @return {@code true} if member access is allowed
      */
-    protected boolean checkStaticFieldAccess(Member member, int memberModifiers) {
-        if (Modifier.isStatic(memberModifiers) && member instanceof Field) {
-            return allowStaticFieldAccess;
-        } else {
+    protected boolean checkStaticFieldAccess(Member member) {
+        if (allowStaticFieldAccess) {
```
kusalk, line 242: Bypass computation if static field access is permitted

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,164 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
-        final Class<?> memberClass = member.getDeclaringClass();
-        // target can be null in case of accessing static fields, since OGNL 3.2.8
-        final Class<?> targetClass = Modifier.isStatic(memberModifiers) ? memberClass : target.getClass();
-        if (!memberClass.isAssignableFrom(targetClass)) {
-            throw new IllegalArgumentException("Target does not match member!");
+        if (target instanceof Class) { // Target may be of type Class for static members
+            if (!member.getDeclaringClass().equals(target)) {
+                throw new IllegalArgumentException("Target class does not match static member!");
+            }
+            target = null;
```
kusalk, line 112: Set to null as there is no more useful information to extract here, and it simplifies the checks/logic to follow

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -104,126 +105,168 @@ public void restore(Map context, Object target, Member member, String propertyNa
...
 
-        final int memberModifiers = member.getModifiers();
-        final Class<?> memberClass = member.getDeclaringClass();
-        // target can be null in case of accessing static fields, since OGNL 3.2.8
-        final Class<?> targetClass = Modifier.isStatic(memberModifiers) ? memberClass : target.getClass();
-        if (!memberClass.isAssignableFrom(targetClass)) {
-            throw new IllegalArgumentException("Target does not match member!");
+        if (target != null) {
+            // Special case: Target is a Class object but not Class.class
+            if (Class.class.equals(target.getClass()) && !Class.class.equals(target)) {
```
kusalk, line 110: Validate and throw exceptions if arguments are not what we expect as the following logic depends on these assumptions to be accurate

lukaszlenart (approved):
I'm good, thanks for all the changes! And as you said `SecurityMemberAccess` shouldn't be used by users directly so maybe marking it `final` would be a good idea?

kusalk:
@lukaszlenart I'm going to make it extensible as part of WW-5343 so that applications can add even stronger checks if they so desire - so will leave it not `final`

PR #785, opened 2023-11-12
[WW-5335] Prepares for Java 17 & Struts 7.x

lukaszlenart:
Closes [WW-5335](https://issues.apache.org/jira/browse/WW-5335)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -17,7 +17,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "master" ]
+    branches:
+      - master
+      - release/struts-7-0-x
   pull_request:
 
 permissions:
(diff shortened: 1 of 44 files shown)


.github/workflows/codeql.yml
kusalk, line 1: Need to add `java-version` to these job(s)

pom.xml
```diff
@@ -51,7 +51,7 @@
         <connection>scm:git:https://gitbox.apache.org/repos/asf/struts.git</connection>
         <developerConnection>scm:git:https://gitbox.apache.org/repos/asf/struts.git</developerConnection>
         <url>https://github.com/apache/struts/</url>
-        <tag>STRUTS_6_3_0_1</tag>
+        <tag>release/struts-7-0-x</tag>
```
kusalk, line 54: Convention here is usually `HEAD`

kusalk (changes requested):

.github/workflows/codeql.yml
lukaszlenart: like this?

kusalk: I think so - let's see if the build goes green

kusalk (approved):

sepe81 (approved):
LGTM

lukaszlenart:
No exact plans but first 2.5.x must be EOL which will happen in two months so you can expect the first Struts 7.x around that date

PR #789, opened 2023-11-13
WW-5363 Velocity: read chained contexts before ValueStack

kusalk:
WW-5363
--
I think it's probably a fair assumption that in most applications that the OgnlValueStack lookup will be the most expensive context lookup. Given that, it is probably better we do that last. I've refactored this class slightly to make the lookup order easily overridable.


plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
```diff
@@ -54,38 +55,55 @@ public StrutsVelocityContext(ValueStack stack) {
...
             if (val != null) {
                 return val;
             }
-            val = stack.getContext().get(key);
+        }
+        return null;
+    }
+
+    protected List<Function<String, Object>> contextGetterList() {
+        return Arrays.asList(super::internalGet, this::chainedContextGet, this::stackGet, this::stackContextGet);
```
kusalk, line 80: Happy to revert this back to the original order if you think that's a more sensible default

lukaszlenart (approved):

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
lukaszlenart: I think this is fine, we will see if this will produce some backward incompatibility

PR #778, opened 2023-10-30
[WW-5141] Support for JakartaEE

jdyer1:
WW-5141
[DRAFT - for evaluation only]

I am trying to help begin the effort to support jakarta ee 9+.  This PR represents a "least amount of work" path to the goal.  The code all compiles, but tests do not yet.  There are some BIG TODO's remaining also.   I would like to know if this is the correct path and what is thought to be the next steps completing this ticket.

This is a huge PR!  To help any reviewer to navigate, here is a summary of the changes.

1. modify references in java source:  "javax.">"jakarta."

2. pom dependency changes:
- Java version: 1.8 > 11
- javax.servlet-api > jakarta.servlet-api
- jsp-api > jakarta.servlet.jsp-api
- validation-api > jakarta.validation-api
- javax.el > jakarta.el
- cdi-api > jakarta.enterprise.cdi-api
- jstl > jakarta.servlet.jsp.jstl-api
- persistence-api > jakarta.persistence-api
- hibernate-validator.version: 6.2.4.Final > 8.0.1.Final
- spring.platformVersion: 6.2.4.Final > 8.0.1.Final
- tomcat-* version: 8.5.53 > 10.1.15
- sitemesh to g:org.sitemesh a:sitemesh v:3.2.0-M2  (TODO:  Maybe not neccessary???)
- spring-test version set to 4.3.0.RELEASE for plugins/cdi/pom.xml only.  See comment in pom.xml

3. Code changes

3a./plugins/embeddedjsp/../Generator.java
- change to: "out.printil("    extends jakarta..."

3b. plugins/embeddedjsp/../PageContextImpl.java
- JspException has no "getRootCause()".  change to "getCause()".\

3c. plugins/embeddedjsp/../JspCServletContext.java
- Add no-op methods for those added to interface

3d. /plugins/embeddedjsp/../JspServletWrapper.java
- Remove special logic for "instanceof SingleThreadModel" (removed from api)

3e. /plugins/json/../StrutsMockHttpServletRequest.java
- "com.mockobjects.servlet.MockHttpServletRequest" > "org.springframework.mock.web.MockHttpServletRequest"

3f. /plugins/json/../StrutsMockHttpSession.java
- "com.mockobjects.servlet.MockHttpServletRequest" > "org.springframework.mock.web.MockHttpServletRequest"

3g. /core/../RequestUtils.java
- request.getAttrubite("javax.servlet > ..jakarta.servlet

3h. /core/../Include.java
[...]

PR #798, opened 2023-11-21
Feature/wn 5141 removals only

jdyer1:
@kusalk  As you requested, this is a spin-off from PR  #778.  My hope is this can be the first chunk of the larger project to get the Jakarta changes all merged to the 7.X release branch!


lukaszlenart (approved):
❤️

plugins/pom.xml
kusalk, line 1: You deleted all these modules without actually deleting the files

PR #799, opened 2023-11-23
Delete missed files and deprecated plugins

kusalk:
#798 deleted some deprecated plugin modules but not the associated files and assembly entries.
I've deleted those missed files in this PR.
I've also deleted the remaining deprecated plugins.


lukaszlenart (approved):
Thanks for paying some much attention ❤️

PR #800, opened 2023-11-23
WW-5364 Automatically populate OGNL allowlist

kusalk:
WW-5364
--
We automatically populate the allowlist from interceptors, actions and result types defined in XML configuration.
This will allow most applications to turn on the OGNL allowlist with no or limited additional configuration.
We can then consider turning this security option on by default in Struts 7.0.

Documentation: https://github.com/apache/struts-site/pull/213


kusalk:
WIP: Need to add more unit test coverage for this one

apps/showcase/src/main/resources/struts.xml
```diff
@@ -34,6 +34,19 @@
     <constant name="struts.custom.i18n.resources" value="globalMessages" />
     <constant name="struts.action.extension" value="action,," />
 
+    <constant name="struts.allowlist.enable" value="true" />
```
kusalk, line 37: It works!

core/src/main/java/com/opensymphony/xwork2/XWorkJUnit4TestCase.java
kusalk, line 1: Moved this here so that we can utilise it for tests in core

core/src/main/resources/struts-beans.xml
```diff
@@ -169,6 +169,7 @@
     <bean name="struts" class="com.opensymphony.xwork2.ognl.SecurityMemberAccess" scope="prototype"/>
     <bean type="org.apache.struts2.ognl.OgnlGuard" name="struts"
           class="org.apache.struts2.ognl.StrutsOgnlGuard"/>
+    <bean class="org.apache.struts2.ognl.ProviderAllowlist"/>
```
kusalk, line 172: No extension point - can't see any use for one

core/src/test/java/com/opensymphony/xwork2/config/providers/ConfigurationProviderOgnlAllowlistTest.java
```diff
@@ -0,0 +1,114 @@
...
+                Class.forName("com.opensymphony.xwork2.LocaleProvider"),
+                Class.forName("java.io.Serializable"),
+                Class.forName("com.opensymphony.xwork2.mock.MockResult"),
+                Class.forName("com.opensymphony.xwork2.interceptor.ConditionalInterceptor"),
+                Class.forName("com.opensymphony.xwork2.ActionSupport"),
+                Class.forName("com.opensymphony.xwork2.ActionChainResult"),
+                Class.forName("com.opensymphony.xwork2.TextProvider"),
+                Class.forName("org.apache.struts2.interceptor.NoOpInterceptor"),
+                Class.forName("com.opensymphony.xwork2.interceptor.Interceptor"),
+                Class.forName("java.lang.Object"),
```
kusalk, line 58: Even though `java.lang.Object` is added to the allowlist as it is a superclass, the exclusion list always takes precedence and so this class will not be allowed in practice

plugins/junit/src/main/java/org/apache/struts2/junit/XWorkJUnit4TestCase.java
```diff
@@ -18,74 +18,5 @@
...
-                    @Override
-                    public Class type() {
-                        return impl.getClass();
-                    }
-                }, Scope.SINGLETON);
-            }
-        });
-    }
-
+public abstract class XWorkJUnit4TestCase extends com.opensymphony.xwork2.XWorkJUnit4TestCase {
```
kusalk, line 21: Extended the moved class for compat

core/src/main/resources/struts-default.xml
```diff
@@ -44,8 +44,6 @@
 
         <interceptors>
             <interceptor name="alias" class="com.opensymphony.xwork2.interceptor.AliasInterceptor"/>
-            <interceptor name="autowiring"
```
kusalk, line 47: This class isn't provided by core

lukaszlenart (approved):

PR #806, opened 2023-12-05
WW-5339 Add option to block custom OGNL maps

kusalk:
WW-5339
--
Having consulted with a security researcher - there is still some risk here if there exists a custom map implementation whose `java.util.Map` methods pose some risk. This is because `ognl.MapPropertyAccessor#getProperty` does not consult the `MemberAccess` policy for the inherent `java.util.Map` methods.

Given Struts does not rely on custom OGNL Map implementations, we should add an option to block this capability.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
@@ -21,7 +21,6 @@ package com.opensymphony.xwork2.ognl;
 import com.opensymphony.xwork2.conversion.impl.XWorkConverter;
 import com.opensymphony.xwork2.inject.Container;
 import com.opensymphony.xwork2.inject.Inject;
-import com.opensymphony.xwork2.ognl.accessor.CompoundRootAccessor;
 import com.opensymphony.xwork2.util.CompoundRoot;
 import com.opensymphony.xwork2.util.reflection.ReflectionException;
 import ognl.ClassResolver;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/OgnlUtil.java
```diff
@@ -856,10 +855,12 @@ protected Map<String, Object> createDefaultContext(Object root) {
         return createDefaultContext(root, null);
     }
 
-    protected Map<String, Object> createDefaultContext(Object root, ClassResolver classResolver) {
-        ClassResolver resolver = classResolver;
+    protected Map<String, Object> createDefaultContext(Object root, ClassResolver resolver) {
         if (resolver == null) {
-            resolver = container.getInstance(CompoundRootAccessor.class);
```
kusalk, line 862: This didn't work :(

lukaszlenart (approved):
Nice, could you document another security option ? Thanks!

kusalk:
@lukaszlenart
Done :)
https://github.com/apache/struts-site/pull/215

lukaszlenart:
Just tested the Showcase app and it doesn't look good ;-)
```
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:205) - Declaring class [interface java.util.Map$Entry] of member type [public abstract java.lang.Object java.util.Map$Entry.getKey()] is not allowlisted!
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:163) - Access to non-public [final java.lang.Object java.util.HashMap$Node.key] is blocked!
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:163) - Access to non-public [protected java.lang.String org.apache.struts2.components.UIBean.key] is blocked!
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:205) - Declaring class [interface java.util.Map$Entry] of member type [public abstract java.lang.Object java.util.Map$Entry.getValue()] is not allowlisted!
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:163) - Access to non-public [java.lang.Object java.util.HashMap$Node.value] is blocked!
[WARN ] ognl.SecurityMemberAccess (SecurityMemberAccess.java:163) - Access to non-public [protected java.lang.String org.apache.struts2.components.UIBean.value] is blocked!
...
```

PR #807, opened 2023-12-10
[WW-5370] Makes HttpParameters case-insensitive

lukaszlenart:
Closes [WW-5370](https://issues.apache.org/jira/browse/WW-5370)

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
@@ -29,7 +29,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 
 @SuppressWarnings("unchecked")
-public class HttpParameters implements Map<String, Parameter>, Cloneable {
+public class HttpParameters implements Map<String, Parameter> {
 
     final private Map<String, Parameter> parameters;
 
@@ -37,6 +37,7 @@ public class HttpParameters implements Map<String, Parameter>, Cloneable {
         this.parameters = parameters;
     }
 
+    @SuppressWarnings("rawtypes")
     public static Builder create(Map requestParameterMap) {
         return new Builder(requestParameterMap);
     }
@@ -47,7 +48,7 @@ public class HttpParameters implements Map<String, Parameter>, Cloneable {
 
     public HttpParameters remove(Set<String> paramsToRemove) {
         for (String paramName : paramsToRemove) {
-            parameters.remove(paramName);
+            parameters.entrySet().removeIf(p -> p.getKey().equalsIgnoreCase(paramName));
         }
         return this;
     }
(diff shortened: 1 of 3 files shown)


yasserzamani (approved):
LGTM, thanks!

P.S. I'm not sure how much it's hard to implement but you might know better: Looks like that same functionality can be achieved with less code by replacing

final private Map<String, Parameter> parameters;

with

final private Map<String, Parameter> parameters = new TreeMap<String, Parameter>(String.CASE_INSENSITIVE_ORDER);

Or using Apache Commons [CaseInsensitiveMap](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/CaseInsensitiveMap.html).

kusalk:
Yeah good call - would be better to change the underlying map to be one of the off-the-shelf implementations. The Apache Commons one probably makes more sense.

lukaszlenart:
Thanks for the ideas! I decided to use `TreeMap` as proposed by Yasser to avoid introducing another dependency (Commons Collection is used only in tests)

core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
```diff
@@ -29,25 +29,27 @@
...
 
     public static Builder create() {
-        return new Builder(new HashMap<>());
+        return new Builder(new TreeMap<>(String.CASE_INSENSITIVE_ORDER));
     }
 
     public HttpParameters remove(Set<String> paramsToRemove) {
         for (String paramName : paramsToRemove) {
-            parameters.remove(paramName);
+            parameters.entrySet().removeIf(p -> p.getKey().equalsIgnoreCase(paramName));
```
kusalk, line 52: Do we still need this custom comparison?

core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
```diff
@@ -73,7 +78,14 @@ private Map<String, String[]> toMap() {
...
     }
 
+    /**
+     * Appends all the parameters by overriding any existing params in a case-insensitive manner
+     *
+     * @param newParams A new params to append
+     * @return a current instance of {@link HttpParameters}
+     */
     public HttpParameters appendAll(Map<String, Parameter> newParams) {
+        remove(newParams.keySet());
```
kusalk, line 88: Probably don't need this anymore either right?

core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
```diff
@@ -100,7 +112,7 @@ public boolean containsValue(Object value) {
 
     @Override
     public Parameter get(Object key) {
-        if (parameters.containsKey(key)) {
+        if (key != null && contains(String.valueOf(key))) {
```
kusalk, line 115: Not sure this part makes sense, checking for the existence of String#valueOf then retrieving using the object?

Do you mean something like this:
```
@Override
public Parameter get(Object key) {
    if (key == null) {
        return new Parameter.Empty("null");
    }
    Parameter val = parameters.get(key.toString());
    return val != null ? val : new Parameter.Empty(key.toString());
}
```

core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
lukaszlenart: Right, removed

core/src/main/java/org/apache/struts2/dispatcher/HttpParameters.java
lukaszlenart: Simplified to return `null` - no idea what was the reasoning behind :D

lukaszlenart: I will use your code and I will add missing equals&hashCode

kusalk (approved):

yasserzamani (approved):
Thank you Lukasz! 🙇

PR #808, opened 2023-12-11
[WW-5371] Modern upload

lukaszlenart:
This PR implements a new upload mechanism which uses a dedicated callback interface to pass info about uploaded files.

Closes [WW-5371](https://issues.apache.org/jira/browse/WW-5371)


yasserzamani (approved):
Great and smart work, thanks!

PR #812, opened 2023-12-17
[WW-5362] Removes type attribute out of <s:script/> tag

lukaszlenart:
Closes [WW-5362](https://issues.apache.org/jira/browse/WW-5362)

### DIFF
--- a/assembly/src/main/resources/template.vm
+++ b/assembly/src/main/resources/template.vm
@@ -48,7 +48,7 @@ under the License.
         SyntaxHighlighter.all();
     </script>
 #end
-    <script language="javascript">
+    <script>
         var hide = null;
         var show = null;
         var children = null;
--- a/assembly/src/main/resources/template.vm
+++ b/assembly/src/main/resources/template.vm
@@ -1,6 +1,6 @@
 #set ($stylebase = "https://struts.apache.org/css")
 <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
-<!-- 
+<!--
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
 distributed with this work for additional information
@@ -16,7 +16,7 @@ software distributed under the License is distributed on an
 "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
-under the License. 
+under the License.
 -->
 <html>
 <head>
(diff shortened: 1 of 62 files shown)


lukaszlenart:
FYI @sepe81 @gregh3269

gregh3269:
Seems now I get
SEVERE: Exception sending context initialized event to listener instance of class [org.apache.struts2.tiles.StrutsTilesListener]
java.lang.NoClassDefFoundError: org/apache/commons/digester3/Rule
Caused by: java.lang.ClassNotFoundException: org.apache.commons.digester3.Rule
Deps want commons-digester-2.1.jar think this is v3?

lukaszlenart:
What app have you used? This change shouldn't affect Digester

gregh3269:
My app.

lukaszlenart:
This is related to [WW-5347](https://issues.apache.org/jira/browse/WW-5347) - and you are using 6.3.0.1 version as far I see, changes in this PR are unrelated to your problem.

gregh3269:
OK, why would the jars be missing for

.m2/repository/org/apache/struts/struts2-tiles-plugin/6.4.0-SNAPSHOT
.m2/repository/org/apache/struts/struts2-javatemplates-plugin/6.4.0-SNAPSHOT

whilst 
.m2/repository/org/apache/struts/struts2-core/6.4.0-SNAPSHOT are present.

lukaszlenart:
@gregh3269 can we discuss this in the dev@ email list?

gregh3269:
OK, mailing list.

...although I got it working and it seems good.

assembly/src/main/resources/template.vm
```diff
@@ -43,12 +43,12 @@ under the License.
...
     #end
 
-    <script type="text/javascript">
+    <script>
         SyntaxHighlighter.defaults['toolbar'] = false;
         SyntaxHighlighter.all();
     </script>
 #end
-    <script type="text/javascript" language="javascript">
+    <script language="javascript">
```
sepe81, line 51:
```suggestion
    <script>
```

Should the language attribute also be removed?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#deprecated_attributes states:

>Like the type attribute, this attribute identifies the scripting language in use. Unlike the type attribute, however, this attribute's possible values were never standardized. The type attribute should be used instead.

lukaszlenart: Good catch, thanks!

lukaszlenart:
Are we good here?

core/src/site/confluence/ww-template-autoexport.html
```diff
@@ -49,7 +49,7 @@
         color:                 #666;
       }
     </style>
-    <script type="text/javascript" language="javascript">
+    <script language="javascript">
```
sepe81, line 52:
```suggestion
    <script>
```

one more place

sepe81 (approved):
LGTM after cleanup for [ww-template-autoexport.html](https://github.com/apache/struts/pull/812/files#diff-ff99f59be56cc691ab5126524d28682baa370e99604b4413584e3cde264483d9)

PR #814, opened 2023-12-17
[WW-5373] Update JavaDoc CspReportAction.java 

assachs:
WW-5373 
change comment to match 204 http status and change parameter name statusCode to status.


lukaszlenart:
Thanks a lot for the PR! I also added a small note about the report action to the docs https://github.com/apache/struts-site/pull/218

PR #819, opened 2023-12-20
Fix CVE-2023-50164 in 2-3-x

henrikplate:
Parameters related to file uploads, esp. `<xxx>FileName`, will be added only after existing parameters with identical names (case-insensitive) have been removed.

Since the 2-5-x class `HttpParameters` is not available yet, the fix is done very locally in class `FileUploadInterceptor`, i.e. other potential parameter pollutions are not covered (e.g. during the later execution of other interceptors that also update the parameters).

The fix has been tested using a [vulnerable test application](https://github.com/henrikplate/struts-demo).

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/FileUploadInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/FileUploadInterceptor.java
@@ -46,6 +46,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


henrikplate:
Apart from assessing whether the fix works, I guess the other important question is whether it is worthwhile to produce a version 2.3.38 for the EOL release branch 2-3-x. The severity of this CVE could be one motivation, however, other severe vulnerabilities would remain unaddressed...

lukaszlenart (approved):
The Apache Struts can accept the change but we won't issue a new release. Struts 2.3.x branch is EOL and shouldn't be used at any cause

PR #821, opened 2023-12-27
WW-5378 Add option to NOT fallback to context lookup when finding value on OgnlValueStack

kusalk:
WW-5378
--
The Struts OGNL context is powerful but also a security nightmare - this option allows applications to disable its access completely when used alongside OGNLGuard with `ognl.ASTVarRef,ognl.ASTThisVarRef` configured.


kusalk:
SonarCloud is comparing to the 7.0 branch which isn't right

lukaszlenart (approved):

lukaszlenart:
Could you document `struts.ognl.valueStackFallbackToContext` or at least add it to the JIRA ticket and how to use it? In other cases user will have to investigate the PR to discover the constant.

kusalk:
@lukaszlenart For sure, will throw up a struts-site PR with this option and some further context on the Struts context

kusalk:
@lukaszlenart Here we are :) https://github.com/apache/struts-site/pull/222

PR #822, opened 2023-12-27
WW-5379 Implement alternative mechanism for Velocity directives to obtain ValueStack

kusalk:
WW-5379
--
This affords applications the ability to not include the ValueStack as `$stack` in the Velocity context and reduce risk of SSTI escalation.

### DIFF
--- a/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
+++ b/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
@@ -77,10 +77,10 @@ public class StrutsVelocityContext extends VelocityContext {
     }
 
     protected List<Function<String, Object>> contextGetterList() {
-        return Arrays.asList(this::superGet, this::chainedContextGet, this::stackGet);
+        return Arrays.asList(this::superInternalGet, this::chainedContextGet, this::stackGet);
     }
 
-    protected Object superGet(String key) {
+    protected Object superInternalGet(String key) {
         return super.internalGet(key);
     }
 
@@ -96,7 +96,7 @@ public class StrutsVelocityContext extends VelocityContext {
             return null;
         }
         for (VelocityContext chainedContext : chainedContexts) {
-            Object val = chainedContext.internalGet(key);
+            Object val = chainedContext.get(key);
             if (val != null) {
                 return val;
             }
--- a/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
+++ b/plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
@@ -103,4 +103,8 @@ public class StrutsVelocityContext extends VelocityContext {
         }
         return null;
     }
+
+    public ValueStack getValueStack() {
+        return stack;
+    }
 }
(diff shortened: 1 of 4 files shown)


kusalk:
(Again, SonarCloud is using the wrong base)

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
```diff
@@ -96,11 +96,15 @@ protected Object chainedContextGet(String key) {
             return null;
         }
         for (VelocityContext chainedContext : chainedContexts) {
-            Object val = chainedContext.internalGet(key);
+            Object val = chainedContext.get(key);
```
kusalk, line 99: Was technically a bug as we were failing to look at chained contexts within our chained contexts

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/components/AbstractDirective.java
```diff
@@ -57,8 +59,11 @@ public int getType() {
     protected abstract Component getBean(ValueStack stack, HttpServletRequest req, HttpServletResponse res);
 
     public boolean render(InternalContextAdapter ctx, Writer writer, Node node) throws IOException, ResourceNotFoundException, ParseErrorException, MethodInvocationException {
-        // get the bean
-        ValueStack stack = (ValueStack) ctx.get("stack");
+        ValueStack stack = extractValueStack(ctx);
+        if (stack == null) {
```
kusalk, line 63: This shouldn't occur but left it for backwards compatibility in case there is a scenario I didn't foresee

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/StrutsVelocityContext.java
```diff
@@ -96,11 +96,15 @@ protected Object chainedContextGet(String key) {
...
             if (val != null) {
                 return val;
             }
         }
         return null;
     }
+
+    public ValueStack getValueStack() {
+        return stack;
+    }
```
lukaszlenart, line 109: Do we need to have this public? Also this requires to cast to `StrutsVelocityContext`, wouldn't be better to have a marking interface `ValueStackAware` or `ValueStackProvider` to expose the ValueStack?

kusalk: I think it has to be but I wonder if Directives can simply obtain the ValueStack from the ActionContext? It's not clear to me if the ValueStack on the ActionContext changes between the time of Velocity context creation and directive rendering. I recall this was how the ValueStack was obtained in WebWork 2.1 but I presume it was changed for a reason?

But also, there's no change in terms of security as the stack was already exposed on the `StrutsVelocityContext` instance using `internalGet("stack")` or `get("stack")`.

And yep I can definitely use a marker interface to allow more flexibility in the Velocity context implementation used by applications.

lukaszlenart: It shouldn't change as far I understand, yet I'm not sure if this is true :) Let's leave it as is, at some point I will solve this puzzle ;-)

lukaszlenart (approved):

PR #824, opened 2023-12-29
WW-5381 Introduce extension point for CompoundRootAccessor

kusalk:
WW-5381
--
Fairly straightforward


core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -50,9 +50,11 @@ protected void setXWorkConverter(XWorkConverter converter) {
         this.xworkConverter = converter;
     }
 
-    @Inject(value = "com.opensymphony.xwork2.util.CompoundRoot")
-    protected void setClassResolver(ClassResolver classResolver) {
-        this.compoundRootAccessor = (RootAccessor) classResolver;
+    @Inject
+    protected void setCompoundRootAccessor(RootAccessor compoundRootAccessor) {
+        this.compoundRootAccessor = compoundRootAccessor;
+        OgnlRuntime.setPropertyAccessor(CompoundRoot.class, compoundRootAccessor);
```
kusalk, line 56: Unified the previously 3 separate instances of CompoundRootAccessor

kusalk:
(SonarCloud report is comparing against wrong base branch)

core/src/main/resources/struts-beans.xml
```diff
@@ -174,11 +174,9 @@
     <bean type="com.opensymphony.xwork2.util.TextParser" name="struts"
           class="com.opensymphony.xwork2.util.OgnlTextParser" scope="singleton"/>
 
-    <bean type="ognl.ClassResolver" name="com.opensymphony.xwork2.util.CompoundRoot"
+    <bean type="com.opensymphony.xwork2.ognl.accessor.RootAccessor" name="struts"
           class="com.opensymphony.xwork2.ognl.accessor.CompoundRootAccessor"/>
```
lukaszlenart, line 178: Could this be a breaking change? Maybe somehow mention that `ClassResolver` shouldn't be used anymore and implement `CompoundRoot` instead?

lukaszlenart (approved):
Looks good, just one small comment

core/src/main/resources/struts-beans.xml
kusalk: Yep very true - looks like our bundled version of Guice is quite old and doesn't support binding beans to multiple interfaces. Although that might not work here either since the Bean name has also changed.

I can create a 2nd bean using the old interface and name to prevent injection failures but obviously this bean won't be the same one used by the `OgnlValueStackFactory` - it might be fine depending on what these classes intend to do with the `CompoundRootAccessor` bean, and if they really need it to share state with the bean injected into `OgnlValueStackFactory`

kusalk: Done - reinstated the old beans for injection compatibility with any custom beans written for Struts 6.3 or earlier

PR #825, opened 2023-12-29
WW-5381 Introduce extension point for MethodAccessor

kusalk:
WW-5381
--
Fairly straight forward again but see my comment below


core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
```diff
@@ -87,12 +92,6 @@ protected void setContainer(Container container) throws ClassNotFoundException {
             OgnlRuntime.setPropertyAccessor(cls, container.getInstance(PropertyAccessor.class, name));
         }
 
-        names = container.getInstanceNames(MethodAccessor.class);
```
kusalk, line 90: Now there is one drawback here - applications lose the ability to specify custom method accessors for different types. They will now instead need to use the extension point and do the delegation within this "global" method accessor. I'm not sure there's a way to retain this capability and make it compatible with the current extension point system.

kusalk:
(SonarCloud report is comparing against wrong base branch)

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
lukaszlenart: Hmm... so maybe just mark this functionality as `@deprecated` or remove - once someone starts complain we can re-think how to put back such functionality.

lukaszlenart (approved):

core/src/main/java/com/opensymphony/xwork2/ognl/OgnlValueStackFactory.java
kusalk: Hmm actually just had another think and we should be able to retain backwards compatibility, the code will be a little ugly though - let me give it a shot

kusalk: I think we are good now :) Hopefully the code makes sense, added some JavaDoc

lukaszlenart: Looks good to me, thanks!

lukaszlenart (approved):

PR #826, opened 2023-12-30
WW-5382 Fix stale injections in Dispatcher

kusalk:
WW-5382
--
I understand most applications do not reload configuration after startup - but for applications that do, this is a critical bug as it means stale configuration persists after a reload. This extends to the `SecurityMemberAccess` policy which is now container injected.


kusalk:
Need to test this further in a staging environment and will improve test coverage to prevent similar regressions from reoccurring

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -988,18 +1031,7 @@ protected boolean isMultipartRequest(HttpServletRequest request) {
      * @return a multi part request object
      */
     protected MultiPartRequest getMultiPartRequest() {
-        MultiPartRequest mpr = null;
```
kusalk, line 991: Remnants from before the bean aliasing functionality existed I presume?

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -1068,22 +1100,17 @@ public ConfigurationManager getConfigurationManager() {
...
-                ContainerHolder.store(container);
-                return container;
+        if (ContainerHolder.get() == null) {
+            try {
+                ContainerHolder.store(getConfigurationManager().getConfiguration().getContainer());
+            } catch (NullPointerException e) {
+                throw new IllegalStateException("ConfigurationManager and/or Configuration should not be null", e);
             }
         }
+        if (injectedContainer != ContainerHolder.get()) {
```
kusalk, line 1110: This is the core change of this PR - we compare the retrieved container instance against the one we last injected this with, and if it has changed, reinject

core/src/main/java/org/apache/struts2/dispatcher/MockDispatcher.java
kusalk, line 1: Deleted this class in favour of partial mocks using Mockito spy - still not ideal but a step in the right direction

core/src/test/java/org/apache/struts2/StrutsInternalTestCase.java
```diff
@@ -38,22 +40,22 @@ public abstract class StrutsInternalTestCase extends XWorkTestCase {
     /**
      * Sets up the configuration settings, XWork configuration, and
      * message resources
-     * 
+     *
      * @throws java.lang.Exception
      */
     @Override
     protected void setUp() throws Exception {
-        super.setUp();
```
kusalk, line 46: These super calls were just wasted computation as everything was replaced right after

core/src/main/java/org/apache/struts2/util/StrutsTestCaseHelper.java
```diff
@@ -52,8 +50,16 @@ public static Dispatcher initDispatcher(ServletContext ctx, Map<String,String> p
         return du;
     }
 
-    public static void tearDown() throws Exception {
-        Dispatcher.setInstance(null);
+    public static void tearDown(Dispatcher dispatcher) {
+        if (dispatcher != null && dispatcher.getConfigurationManager() != null) {
```
kusalk, line 54: Moved this here for better reuse

core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
```diff
@@ -18,46 +18,65 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
 /**
  * Test case for Dispatcher.
  */
-public class DispatcherTest extends StrutsInternalTestCase {
+public class DispatcherTest extends StrutsJUnit4InternalTestCase {
```
kusalk, line 77: Upgraded to JUnit4 and replaced older mocks with Mockito

core/src/test/java/org/apache/struts2/dispatcher/DispatcherTest.java
```diff
@@ -421,196 +397,201 @@ public void testServiceActionCreatesNewProxyIfDifferentMapping() throws Exceptio
...
+        when(request.getLocale()).thenThrow(new IllegalStateException("Test theoretical state preventing HTTP Request Locale access"));
+        ActionContext context = ActionContext.of(createTestContextMap(dispatcher, request, response));
 
         // Then
         assertEquals(Locale.getDefault(), context.getLocale());  // Expect the system default value when Mock request access fails.
-        mock.verify();
+    }
+
+    @Test
+    public void dispatcherReinjectedAfterReload() {
```
kusalk, line 569: New test for reinjection functionality

plugins/portlet/src/main/java/org/apache/struts2/portlet/dispatcher/Jsr168Dispatcher.java
kusalk, line 1: (Changed line-endings to `LF` for this file)

plugins/sitemesh/src/main/java/org/apache/struts2/sitemesh/OldDecorator2NewStrutsDecorator.java
kusalk, line 1: Same deal with line-endings

kusalk:
I've now tested this fix with the Confluence DC test suite and did some additional manual sanity testing and all seems to be as expected now. Following a `ConfigurationManager#reload`, all constants, including the exclusion list and allowlist, are now refreshed as per the latest ConfigurationProviders.

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -325,12 +325,8 @@ public Class<? extends Configuration> type() {
     }
 
     protected ActionContext setContext(Container cont) {
-        ActionContext context = ActionContext.getContext();
-        if (context == null) {
```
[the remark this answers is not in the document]
lukaszlenart: To be honest I would love to re-organise the whole proces. I assume the `ActionContext` is only needed when serving an action, it should be created just before and destroyed just after.

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -192,6 +194,11 @@ public class Dispatcher {
      * Store ConfigurationManager instance, set on init.
      */
     protected ConfigurationManager configurationManager;
+    private ObjectFactory objectFactory;
```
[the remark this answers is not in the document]
lukaszlenart: Maybe they should be moved into constructor?

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
lukaszlenart: Hm... good question, no idea :)

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -988,18 +1031,7 @@ protected boolean isMultipartRequest(HttpServletRequest request) {
...
-        for (String multiName : multiNames) {
-            if (multiName.equals(multipartHandlerName)) {
-                mpr = getContainer().getInstance(MultiPartRequest.class, multiName);
-            }
-        }
-        if (mpr == null) {
-            mpr = getContainer().getInstance(MultiPartRequest.class);
-        }
-        return mpr;
+        return getContainer().getInstance(MultiPartRequest.class);
```
lukaszlenart, line 1034: Shouldn't it be an instance field as other?

lukaszlenart (commented):
Looks ok, left a few comments

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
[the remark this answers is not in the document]
kusalk: Yeah that would be ideal - if that were the case we'd be able to remove the need for the bootstrap context too it looks like

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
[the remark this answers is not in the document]
kusalk: Since the Dispatcher itself is not container-managed, we can't use constructor injection for it

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
kusalk: `MultiPartRequest` is prototype scope and we need a unique instance per request so injecting it once wouldn't work here

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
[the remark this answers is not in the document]
lukaszlenart: It doesn't have to be, just using `container.inject(Dispatcher.class)` will create a new instance and injects all the dependencies. Anyway this would be a bit out of scope for this PR.

lukaszlenart (approved):

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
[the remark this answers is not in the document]
kusalk: Yep would require a major refactor, I've got some ideas but not a high priority right now, as long as everything works

PR #828, opened 2024-01-02
WW-5364 Add String.class to system allowlist

kusalk:
WW-5364
--
This class is often used in validation logic and is not dangerous so makes sense to add to the system allowlist.


sepe81 (commented):
Maybe insert after line 69 for better readability?

kusalk:
@sepe81 True let me alphabetise this list :)

sepe81 (approved):

lukaszlenart (approved):

PR #830, opened 2024-01-03
[WW-5383] Updates RegEx to excludes JARs by default

lukaszlenart:
Closes [WW-5383](https://issues.apache.org/jira/browse/WW-5383)

### DIFF
--- a/plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
+++ b/plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
@@ -24,14 +24,23 @@ import com.opensymphony.xwork2.FileManagerFactory;
 import com.opensymphony.xwork2.ObjectFactory;
 import com.opensymphony.xwork2.config.Configuration;
 import com.opensymphony.xwork2.config.ConfigurationException;
-import com.opensymphony.xwork2.config.entities.*;
+import com.opensymphony.xwork2.config.entities.ActionConfig;
+import com.opensymphony.xwork2.config.entities.ExceptionMappingConfig;
+import com.opensymphony.xwork2.config.entities.InterceptorMapping;
+import com.opensymphony.xwork2.config.entities.PackageConfig;
+import com.opensymphony.xwork2.config.entities.ResultConfig;
 import com.opensymphony.xwork2.inject.Container;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.util.AnnotationUtils;
 import com.opensymphony.xwork2.util.TextParseUtil;
 import com.opensymphony.xwork2.util.WildcardHelper;
 import com.opensymphony.xwork2.util.classloader.ReloadingClassLoader;
-import com.opensymphony.xwork2.util.finder.*;
+import com.opensymphony.xwork2.util.finder.ClassFinder;
+import com.opensymphony.xwork2.util.finder.ClassFinderFactory;
+import com.opensymphony.xwork2.util.finder.ClassLoaderInterface;
+import com.opensymphony.xwork2.util.finder.ClassLoaderInterfaceDelegate;
+import com.opensymphony.xwork2.util.finder.Test;
+import com.opensymphony.xwork2.util.finder.UrlSet;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
```diff
@@ -673,17 +697,17 @@ protected void buildConfiguration(Set<Class> classes) {
...
                 Map<String, List<Action>> map = getActionAnnotations(actionClass);
                 Set<String> actionNames = new HashSet<>();
                 boolean hasDefaultMethod = ReflectionTools.containsMethod(actionClass, DEFAULT_METHOD);
                 if (!map.containsKey(DEFAULT_METHOD)
-                        && hasDefaultMethod
-                        && actionAnnotation == null && actionsAnnotation == null
-                        && (alwaysMapExecute || map.isEmpty())) {
+                    && hasDefaultMethod
+                    && actionAnnotation == null && actionsAnnotation == null
+                    && (alwaysMapExecute || map.isEmpty())) {
```
kusalk, line 710: My personal preference is to use double indents for continuations - it avoids situations like these where the guard blends in with the following code

kusalk (approved):

plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
```diff
@@ -24,69 +24,102 @@
...
+    private static final String DEFAULT_METHOD = "execute";
+
     private static final boolean EXTRACT_BASE_INTERFACES = true;
 
+    /**
+     * Pattern to match the whole path with sub-path as on JDK9+ getClassLoader().getResources("")
+     * can return also a sub-path like "!/META-INF/versions/..."
+     */
+    private static final String EXCLUDE_ALL_JARS_PATTERN = ".*?\\.jar(!/|/)?(.*)?";
+    private static final String DEFAULT_SPLIT_PATTERN = "\\s*,\\s*";
```
sepe81, line 93: looks good without the original `[,]` 👍

plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
lukaszlenart: I'm using standard IDEA settings, yet it would be good to introduce a common IDE independent mechanism to format code.

sepe81: IMHO Eclipse does this by default, IntelliJ not.

We could introduce `.editorconfig` settings. The following setting does the job:

```
[*.java]
continuation_indent_size = 8
```

plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
```diff
@@ -438,30 +471,26 @@ private UrlSet buildUrlSet(List<URL> resourceUrls) throws IOException {
...
+            // This happens in GAE since the sandbox contains no java.home directory
+            LOG.warn("Could not exclude JAVA_HOME, is this a sandbox jvm?");
         }
         urlSet = urlSet.excludePaths(System.getProperty("sun.boot.class.path", ""));
         urlSet = urlSet.exclude(".*/JavaVM.framework/.*");
 
         if (includeJars == null) {
-            urlSet = urlSet.exclude(".*?\\.jar(!/|/)?");
+            LOG.debug("\"{}\" is not defined, excluding all JAR files!", ConventionConstants.CONVENTION_ACTION_INCLUDE_JARS);
+            urlSet = urlSet.exclude(EXCLUDE_ALL_JARS_PATTERN);
```
sepe81, line 491: I guess the extra `(.*)?` from `EXCLUDE_ALL_JARS_PATTERN = ".*?\\.jar(!/|/)?(.*)?";` is intentional, isn't it?

plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
kusalk: Just checked this as I was curious and doesn't seem to be the case (at least on the version I'm using)
image

lukaszlenart: Oh, right! Just checked my settings and it was 4 for continuation, reverted to default one :)

plugins/convention/src/main/java/org/apache/struts2/convention/PackageBasedActionConfigBuilder.java
lukaszlenart: Yes, since JDK9 we can get either `jar:file:/.../org/codehaus/plexus/plexus-utils/3.4.1/plexus-utils-3.4.1.jar!/META-INF/versions/10/` or just `jar:file:/.../org/codehaus/plexus/plexus-utils/3.4.1/plexus-utils-3.4.1.jar!/`

kusalk (approved):

sepe81 (approved):

PR #831, opened 2024-01-03
WW-5352 Refactor ParametersInterceptor

kusalk:
WW-5352
--
General clean up in preparation for #832

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/security/AcceptedPatternsChecker.java
+++ b/core/src/main/java/com/opensymphony/xwork2/security/AcceptedPatternsChecker.java
@@ -32,37 +32,37 @@ public interface AcceptedPatternsChecker {
      * @param value to check
      * @return object containing result of matched pattern and pattern itself
      */
-    public IsAccepted isAccepted(String value);
+    IsAccepted isAccepted(String value);
 
     /**
      * Sets excluded patterns during runtime
      *
      * @param commaDelimitedPatterns comma delimited string with patterns
      */
-    public void setAcceptedPatterns(String commaDelimitedPatterns);
+    void setAcceptedPatterns(String commaDelimitedPatterns);
 
     /**
      * Set excluded patterns during runtime
      *
      * @param patterns array of additional excluded patterns
      */
-    public void setAcceptedPatterns(String[] patterns);
+    void setAcceptedPatterns(String[] patterns);
 
     /**
      * Sets excluded patterns during runtime
      *
      * @param patterns set of additional patterns
      */
-    public void setAcceptedPatterns(Set<String> patterns);
+    void setAcceptedPatterns(Set<String> patterns);
 
     /**
      * Allow access list of all defined excluded patterns
      *
      * @return set of excluded patterns
      */
-    public Set<Pattern> getAcceptedPatterns();
+    Set<Pattern> getAcceptedPatterns();
 
-    public final static class IsAccepted {
+    final class IsAccepted {
 
         private final boolean accepted;
         private final String acceptedPattern;
--- a/core/src/main/java/org/apache/struts2/dispatcher/Parameter.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/Parameter.java
@@ -18,11 +18,11 @@
  */
 package org.apache.struts2.dispatcher;
 
-import java.util.Objects;
-
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+
+import java.util.Objects;
 
 public interface Parameter {
 
@@ -58,7 +58,7 @@ public interface Parameter {
         @Override
         public String getValue() {
             String[] values = toStringArray();
-            return (values != null && values.length > 0) ? values[0] : null;
+            return values.length > 0 ? values[0] : null;
         }
 
         private String[] toStringArray() {
@@ -124,7 +124,7 @@ public interface Parameter {
 
     class Empty implements Parameter {
 
-        private String name;
+        private final String name;
 
         public Empty(String name) {
             this.name = name;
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/opensymphony/xwork2/security/AcceptedPatternsChecker.java
```diff
@@ -32,37 +32,37 @@ public interface AcceptedPatternsChecker {
      * @param value to check
      * @return object containing result of matched pattern and pattern itself
      */
-    public IsAccepted isAccepted(String value);
```
kusalk, line 35: Redundant in interfaces

core/src/main/java/org/apache/struts2/dispatcher/Parameter.java
```diff
@@ -58,7 +58,7 @@ public String getName() {
         @Override
         public String getValue() {
             String[] values = toStringArray();
-            return (values != null && values.length > 0) ? values[0] : null;
```
kusalk, line 61: `#toStringArray` never returns a null value

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -132,138 +133,149 @@ static private int countOGNLCharacters(String s) {
     @Override
     public String doIntercept(ActionInvocation invocation) throws Exception {
```
kusalk, line 134: Used early returns to reduce nesting and improve readability

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -132,138 +133,149 @@ static private int countOGNLCharacters(String s) {
...
      * @param newParams The parameter map to apply
-     *                  <p>
-     *                  In this class this is a no-op, since the parameters were fetched from the same location.
-     *                  In subclasses both retrieveParameters() and addParametersToContext() should be overridden.
-     *                  </p>
      */
     protected void addParametersToContext(ActionContext ac, Map<String, ?> newParams) {
     }
 
     protected void setParameters(final Object action, ValueStack stack, HttpParameters parameters) {
```
kusalk, line 185: Extracted a bunch of helper methods to make this method much more readable

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -287,13 +302,11 @@ protected boolean isAcceptableParameter(String name, Object action) {
      * @return true if parameter is accepted
      */
     protected boolean isAcceptableParameterValue(Parameter param, Object action) {
-        ParameterValueAware parameterValueAware = (action instanceof ParameterValueAware) ? (ParameterValueAware) action : null;
-        boolean acceptableParamValue = (parameterValueAware == null || parameterValueAware.acceptableParameterValue(param.getValue()));
-        if (hasParamValuesToExclude() || hasParamValuesToAccept()) {
-            // Additional validations to process
-            acceptableParamValue &= acceptableValue(param.getName(), param.getValue());
```
kusalk, line 294: `&=` doesn't short-circuit

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -338,18 +344,17 @@ protected boolean acceptableName(String name) {
             return false;
         }
         boolean accepted = isWithinLengthLimit(name) && !isExcluded(name) && isAccepted(name);
-        if (devMode && accepted) { // notify only when in devMode
```
kusalk, line 341: Cleaned out comments for behaviours which are fairly obvious from code

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -132,138 +133,149 @@ static private int countOGNLCharacters(String s) {
...
-            ReflectionContextState.setDenyMethodExecution(context, true);
-            ReflectionContextState.setReportingConversionErrors(context, true);
-
-            //keep locale from original context
             newStack.getActionContext().withLocale(stack.getActionContext().getLocale()).withValueStack(stack);
         }
+        return newStack;
+    }
+
+    protected void setMemberAccessProperties(ValueStack stack) {
```
lukaszlenart, line 216: I would avoid using _setter/getter_ pattern if this isn't a setter _per se_ - this also reduces a risk of using such method in expression evaluation. What about using `applyMemberAccessProperties()` instead?

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -132,138 +133,149 @@ static private int countOGNLCharacters(String s) {
...
+            return HttpParameters.create().withComparator(getOrderedComparator()).withParent(parameters).build();
+        } else {
+            return HttpParameters.create().withParent(parameters).build();
+        }
+    }
 
-        for (Map.Entry<String, Parameter> entry : acceptableParameters.entrySet()) {
-            String name = entry.getKey();
-            Parameter value = entry.getValue();
+    protected void setParametersOnStack(ValueStack stack, Map<String, Parameter> parameters, Object action) {
```
lukaszlenart, line 254: `applyParametersOnStack`?

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
kusalk: I'll rename all of these :) They are `protected` so shouldn't be accessible from expressions but I agree with the premise

lukaszlenart (approved):

PR #791, opened 2023-11-14
WW-5343 Make SecurityMemberAccess an extensible bean

kusalk:
WW-5343
--
This will allow applications to easily extend the capabilities of `SecurityMemberAccess`. It additionally allows us to significantly simplify the configuration loading.

PR #781, opened 2023-11-05
WW-5350 Implement OGNL Allowlist capability

kusalk:
WW-5350
--
Implementation for strict OGNL allowlist feature. It is up to the application to determine which classes/packages need to be allowlisted. The exclusion list will still take precedence (classes on the exclusion list cannot be allowlisted).

I hope to clean this implementation up and both `OgnlUtil` and `SecurityMemberAccess` up as part of WW-5343.

PR #832, opened 2024-01-03
WW-5352 Introducing the StrutsParameter annotation

kusalk:
WW-5352
--
The final piece of the Struts 6.4 security overhaul. When this capability is enabled alongside the OGNL allowlist, security is greatly heightened with no additional complex configuration.

In addition to assisting OGNL allowlist configuration, the primary benefit of this new annotation is that it prevents inexperienced Struts developers from inadvertently introducing parameter injection points - as they often do not realise any public members on an Action class are parameter injectable. All parameter injection points will need to be explicitly annotated.

The OGNL allowlist capability can already be enabled using `struts.allowlist.enable`. It was introduced (#781) alongside the configuration options `struts.allowlist.classes` and `struts.allowlist.packageNames` using which you can define the allowlist. However, as one might have expected, the configuration of an exhaustive allowlist was a massively cumbersome task.

I later introduced #800 which was able to populate the allowlist on startup from all XML configuration provided to Struts. Whilst this significantly reduced the barrier to enabling the allowlist, with most applications only requiring that their model or DTO packages be allowlisted, I felt we could do better.

I had already planned to rework, strengthen and contribute an existing Struts interceptor capability used in Confluence - introduced in this PR as the `@StrutsParameter` annotation. As mentioned, its primary purpose is to prevent the unintentional introduction of parameter injection points. However, I now saw the opportunity to integrate this seamlessly with the OGNL allowlist capability.

Following a crucial contribution (#791) to make the `SecurityMemberAccess` policy more versatile, in this PR I was able to implement the capability to mutate the allowlist during runtime on a per-request basis.

The only code change required by applications is to annotate all Action class members intended for parameter injection with this new annotation. This can be scripted for large codebases by scanning for public members on classes which implement the Action interface.

This new annotation requirement is enabled using the option `struts.parameters.requireAnnotations`. I've added an additional option, `struts.parameters.requireAnnotations.transitionMode`, that applications can use to, as the name suggests, transition into this feature. When this mode is enabled, only 'nested' parameters, i.e. those represented by public getters on Action classes, will require annotations. This means public setters will still be exposed for parameter injection.

What's significant here is that in this mode, the auto-allowlisting capability is not degraded in any way, so it proves a useful transitioning option for applications that wish to enable the OGNL allowlist as soon as possible. These applications can then enable the full annotation capability once they have worked through annotating all their parameters and are ready.

Even with this annotation capability enabled, there will of course still be edge cases where applications will need to manually allowlist certain classes. One such edge case is when a POJO within a POJO needs to be parameter injected. The 'Showcase' application now requires zero manual allowlist configuration following the addition of the annotations.

I believe this to be a major step forward in Struts' security posture. Atlassian will be shipping both the OGNL allowlist and the StrutsParameter annotation capabilities (along with all other recently contributed security options) in Confluence DC 8.8.

I welcome all comments and feedback and encourage everyone to test both the allowlist and annotation capability with their own applications.

### DIFF
--- a/apps/showcase/src/main/java/org/apache/struts2/showcase/action/ParamsAnnotationAction.java
+++ b/apps/showcase/src/main/java/org/apache/struts2/showcase/action/ParamsAnnotationAction.java
@@ -1,0 +1,133 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.showcase.action;
+
+import com.opensymphony.xwork2.ActionSupport;
+import org.apache.struts2.interceptor.parameter.StrutsParameter;
+import org.apache.struts2.showcase.model.MyDto;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+
+/**
+ * This class supports {@link com.atlassian.confluence.stateless.webdriver.selenium3.security.StrutsParametersTest}
+ * which prevents critical security regressions. Do NOT modify without understanding the motivation behind the tests and
+ * the implications of any changes.
+ */
+public class ParamsAnnotationAction extends ActionSupport {
+
+    @StrutsParameter
+    public String varToPrint;
+
+    public String publicField = "no";
+
+    @StrutsParameter
+    public String publicFieldAnnotated = "no";
+
+    private String privateField = "no";
+
+    public int[] publicArray = new int[]{0};
+
+    @StrutsParameter(depth = 1)
+    public int[] publicArrayAnnotated = new int[]{0};
+
+    public List<String> publicList = new ArrayList<>(singletonList("no"));
+
+    @StrutsParameter(depth = 1)
+    public List<String> publicListAnnotated = new ArrayList<>(singletonList("no"));
+
+    private List<String> privateList = new ArrayList<>(singletonList("no"));
+
+    public Map<String, String> publicMap = new HashMap<>(singletonMap("key", "no"));
+
+    @StrutsParameter(depth = 1)
+    public Map<String, String> publicMapAnnotated = new HashMap<>(singletonMap("key", "no"));
+
+    public MyDto publicMyDto = new MyDto();
+
+    @StrutsParameter(depth = 2)
+    public MyDto publicMyDtoAnnotated = new MyDto();
+
+    @StrutsParameter(depth = 1)
+    public MyDto publicMyDtoAnnotatedDepthOne = new MyDto();
+
+    private MyDto privateMyDto = new MyDto();
+
+    public void setPrivateFieldMethod(String privateField) {
+        this.privateField = privateField;
+    }
+
+    @StrutsParameter
+    public void setPrivateFieldMethodAnnotated(String privateField) {
+        this.privateField = privateField;
+    }
+
+    public List<String> getPrivateListMethod() {
+        return privateList;
+    }
+
+    @StrutsParameter(depth = 1)
+    public List<String> getPrivateListMethodAnnotated() {
+        return privateList;
+    }
+
+    public MyDto getUnsafeMethodMyDto() {
+        return privateMyDto;
+    }
+
+    @StrutsParameter(depth = 2)
+    public MyDto getSafeMethodMyDto() {
+        return privateMyDto;
+    }
+
+    @StrutsParameter(depth = 1)
+    public MyDto getSafeMethodMyDtoDepthOne() {
+        return privateMyDto;
+    }
+
+    public String renderVarToPrint() throws ReflectiveOperationException {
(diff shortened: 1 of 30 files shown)


kusalk:
I'm going to write some more acceptance tests before merging this but it's now ready for review :)

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -295,13 +340,150 @@ protected void notifyDeveloperParameterException(Object action, String property,
...
+    protected void allowlistFieldIfParameterized(Field field) {
+        allowlistParameterizedTypeArg(field.getGenericType());
+    }
+
+    /**
+     * Annotation retrieval logic. Can be overridden to support extending annotations or some other form of annotation
+     * inheritance.
+     */
+    protected StrutsParameter getParameterAnnotation(AnnotatedElement element) {
+        return element.getAnnotation(StrutsParameter.class);
```
lukaszlenart, line 475: There is `AnnotationUtils#isAnnotatedBy()` - wouldn't be better to use it here or instead of `getParameterAnnotation()`?

lukaszlenart (commented):
Astonish work, nice 👏

kusalk:
Acceptance tests done :) Struts docs PR coming up

(but also I'm not sure it makes sense to include the showcase code for unit test coverage - I wrote that action specifically for acceptance tests so doesn't make a lot of sense to add unit tests for it IMO)

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
kusalk: Doesn't look that utility method provides any additional value here - `#getParameterAnnotation` exists purely for overriding purposes. Applications can extend this default interceptor and add support for their own annotations or so on.

apps/showcase/src/main/java/org/apache/struts2/showcase/action/ParamsAnnotationAction.java
```diff
@@ -0,0 +1,133 @@
...
+import static java.util.Collections.singletonMap;
+
+/**
+ * This class supports {@link com.atlassian.confluence.stateless.webdriver.selenium3.security.StrutsParametersTest}
+ * which prevents critical security regressions. Do NOT modify without understanding the motivation behind the tests and
+ * the implications of any changes.
+ */
+public class ParamsAnnotationAction extends ActionSupport {
+
+    @StrutsParameter
```
lukaszlenart, line 42: Oh nice, I was going to ask about that, so this mechanism supports injecting incoming values directly into fields without a need to define setter, is that right? What about getter to fetch the value, is it still needed?

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -395,6 +396,7 @@ public static ContainerBuilder bootstrapFactories(ContainerBuilder builder) {
                 .factory(SecurityMemberAccess.class, Scope.PROTOTYPE)
                 .factory(OgnlGuard.class, StrutsOgnlGuard.class, Scope.SINGLETON)
                 .factory(ProviderAllowlist.class, Scope.SINGLETON)
+                .factory(ThreadAllowlist.class, Scope.SINGLETON)
```
lukaszlenart, line 399: Wouldn't be better to use `PROTOTYPE` scope and do not use ThreadLocal instead?

core/src/main/java/org/apache/struts2/ognl/ThreadAllowlist.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Allows any bean to allowlist a class for use in OGNL expressions, for the current thread only. The allowlist can be
+ * cleared once any desired OGNL expressions have been evaluated.
+ *
+ * @since 6.4.0
+ */
+public class ThreadAllowlist {
+
+    private final ThreadLocal<Set<Class<?>>> allowlist = new ThreadLocal<>();
```
lukaszlenart, line 35: Why do you use ThreadLocal? Wouldn't be better to use one global cache for the whole app?

lukaszlenart (commented):
A few questions to clarify my doubts

core/src/main/java/org/apache/struts2/ognl/ThreadAllowlist.java
kusalk: Yeah this is a good question - the reason I went with this design is because by isolating the allowlists on a per-request basis, it should make the application even more secure. Request A will not be able to access a class that might only be required by request B, and vice-versa. Note that we clear the allowlist at the end of every request. The less options attackers have to craft a payload, the better. Atlassian have seen some very creative payloads where our internal classes are chained together in unconventional ways.

There is also no performance benefit by making this a global cache as we don't currently cache the annotation lookups either. I don't expect the overhead of checking annotations per-request to be significant, but this is something we can reevaluate in the future.

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
kusalk: This bean was introduced as a way to mutate the allowlist read by `SecurityMemberAccess` - if it were `prototype` scope, `ParametersInterceptor` and `SecurityMemberAccess` would not share the same instance of `ThreadAllowlist` and thus this wouldn't be possible.

lukaszlenart:
And please fill the request in the Google [program](https://struts.apache.org/submitting-patches.html#googles-patch-reward-program) - you can donate money to charity if you want, your work requires recognition :)

kusalk:
I appreciate that, I will look into it

And please do ask as many clarifying questions as you need :)

apps/showcase/src/main/java/org/apache/struts2/showcase/action/ParamsAnnotationAction.java
lukaszlenart: Ah.. I missed the `public` scope, I was hopping on some _magic_ :D Anyway, thanks for the detailed explanation.

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
lukaszlenart: 👍

lukaszlenart (approved):
No more doubts, LGTM 👍  (I would give a bit of time to other reviewers)

PR #838, opened 2024-01-08
Reduces log level to debug to reduce noise in the logs

lukaszlenart:
See an [example PR run](https://github.com/apache/struts/actions/runs/7442031148/job/20244935590?pr=836) to notice enormous number of the log entries as this is happening for each unit test.


lukaszlenart:
//cc: @sepe81

kusalk (approved):

PR #839, opened 2024-01-08
WW-5385 Decompose StrutsVelocityContext for better re-usability

kusalk:
WW-5385
--
Decomposes `StrutsVelocityContext` into:

* `CompositeContext` - combines multiple contexts into 1, can only be used as an inner or "chained" context

* `ChainedVelocityContext` - extends the standard `VelocityContext` to correct `#getKeys` when an inner context is used
* `DirectiveVelocityContext` - is additionally capable of rendering Struts directives
* `StrutsVelocityContext` - is additionally capable of looking up items in the `ValueStack`

The `DirectiveVelocityContext` specifically, enables applications greater flexibility in which capabilities they wish to leverage. `ChainedVelocityContext` and `CompositeContext` are useful utility contexts that can be used for general Velocity purposes.


kusalk:
SonarCloud note: There are indeed 6 levels of inheritance here but 3 of them are within the Velocity library - plus each of our context classes have a distinct purpose/use

PR #840, opened 2024-01-08
WW-5379 Add one more method to provide Velocity directives with ValueStack

kusalk:
WW-5379
--


kusalk:
Nevermind with this one too

PR #847, opened 2024-01-18
[WW-5369] Re-define minimal library set

lukaszlenart:
This PR adds missing `caffeine-x.x.x.jar` &  `javassist-x.x.x.jar`

[WW-5369](https://issues.apache.org/jira/browse/WW-5369)

### DIFF
--- a/assembly/src/main/assembly/min-lib.xml
+++ b/assembly/src/main/assembly/min-lib.xml
@@ -41,6 +41,8 @@
         <include>ognl:ognl</include>
         <include>commons-fileupload:commons-fileupload</include>
         <include>org.apache.commons:commons-io</include>
+        <include>com.github.ben-manes.caffeine:caffeine</include>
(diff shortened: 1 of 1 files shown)


kusalk (approved):
Caffeine is required for the default configuration but technically optional

lukaszlenart:
The min-lib is used by ppl who do not use Maven, so caffeine is required to start the app in such a case.

kusalk:
If their `struts.properties` has
```
struts.ognl.expressionCacheType=basic
struts.ognl.beanInfoCacheType=basic
```
then Caffeine won't be required - but given the default configuration requires it, probably best to include it

lukaszlenart:
We can either use `basic` and then make Caffeine `optional` in `pom.xml` or distribute it with the minimal set. Most of the uses rarely change the default settings.

kusalk:
Yeah makes sense - best to distribute it

PR #861, opened 2024-01-24
[WW-5388] Uses the latest JakartaEE FileUpload Servlet 6 package

lukaszlenart:
Closes [WW-5388](https://issues.apache.org/jira/browse/WW-5388)

### DIFF
--- a/core/src/main/java/org/apache/struts2/StrutsConstants.java
+++ b/core/src/main/java/org/apache/struts2/StrutsConstants.java
@@ -140,21 +140,28 @@ public final class StrutsConstants {
     public static final String STRUTS_UI_ESCAPE_HTML_BODY = "struts.ui.escapeHtmlBody";
 
     /** The maximum size of a multipart request (file upload) */
-    public static final String STRUTS_MULTIPART_MAXSIZE = "struts.multipart.maxSize";
+    public static final String STRUTS_MULTIPART_MAX_SIZE = "struts.multipart.maxSize";
+
+    /** The maximum size of all uploaded files.
+       Used only with {@link org.apache.struts2.dispatcher.multipart.JakartaStreamMultiPartRequest} */
+    public static final String STRUTS_MULTIPART_MAX_SIZE_OF_FILES = "struts.multipart.maxSizeOfFiles";
 
     /** The maximum number of files allowed in a multipart request */
-    public static final String STRUTS_MULTIPART_MAXFILES = "struts.multipart.maxFiles";
+    public static final String STRUTS_MULTIPART_MAX_FILES = "struts.multipart.maxFiles";
 
     /** The maximum length of a string parameter in a multipart request. */
     public static final String STRUTS_MULTIPART_MAX_STRING_LENGTH = "struts.multipart.maxStringLength";
 
     /** The maximum size per file in a multipart request */
-    public static final String STRUTS_MULTIPART_MAXFILESIZE = "struts.multipart.maxFileSize";
-    /** The directory to use for storing uploaded files */
-    public static final String STRUTS_MULTIPART_SAVEDIR = "struts.multipart.saveDir";
+    public static final String STRUTS_MULTIPART_MAX_FILE_SIZE = "struts.multipart.maxFileSize";
 
-    /** Declares the buffer size to be used during streaming multipart content to disk. Used only with {@link org.apache.struts2.dispatcher.multipart.JakartaStreamMultiPartRequest} */
-    public static final String STRUTS_MULTIPART_BUFFERSIZE = "struts.multipart.bufferSize";
+    /** The directory to use for storing uploaded files */
+    public static final String STRUTS_MULTIPART_SAVE_DIR = "struts.multipart.saveDir";
+
+    /** Declares the buffer size to be used during streaming multipart content to disk.
+     * Used only with {@link org.apache.struts2.dispatcher.multipart.JakartaStreamMultiPartRequest}
+     */
+    public static final String STRUTS_MULTIPART_BUFFER_SIZE = "struts.multipart.bufferSize";
 
     /**
      * The org.apache.struts2.dispatcher.multipart.MultiPartRequest parser implementation
--- a/core/src/main/java/org/apache/struts2/action/UploadedFilesAware.java
+++ b/core/src/main/java/org/apache/struts2/action/UploadedFilesAware.java
@@ -20,6 +20,7 @@ package org.apache.struts2.action;
 
 import org.apache.struts2.dispatcher.multipart.UploadedFile;
 
+import java.io.File;
 import java.util.List;
 
 /**
@@ -35,6 +36,6 @@ public interface UploadedFilesAware {
      *
      * @param uploadedFiles a list of {@link UploadedFile}, cannot be null. It can be empty.
      */
-    void withUploadedFiles(List<UploadedFile> uploadedFiles);
+    void withUploadedFiles(List<UploadedFile<File>> uploadedFiles);
 
 }
--- a/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
@@ -20,12 +20,12 @@ package org.apache.struts2.dispatcher.multipart;
 
 import com.opensymphony.xwork2.LocaleProviderFactory;
 import com.opensymphony.xwork2.inject.Inject;
+import jakarta.servlet.http.HttpServletRequest;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.struts2.StrutsConstants;
 import org.apache.struts2.dispatcher.LocalizedMessage;
 
-import jakarta.servlet.http.HttpServletRequest;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
@@ -125,7 +125,7 @@ public abstract class AbstractMultiPartRequest implements MultiPartRequest {
 
     /**
      * @param request Inspect the servlet request and set the locale if one wasn't provided by
-     * the Struts2 framework.
+     *                the Struts2 framework.
      */
     protected void setLocale(HttpServletRequest request) {
         if (defaultLocale == null) {
@@ -136,7 +136,7 @@ public abstract class AbstractMultiPartRequest implements MultiPartRequest {
     /**
      * Build error message.
      *
-     * @param e the Throwable/Exception
+     * @param e    the Throwable/Exception
      * @param args arguments
      * @return error message
      */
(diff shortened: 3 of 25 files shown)


core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -171,4 +171,8 @@ protected String getCanonicalName(final String originalFileName) {
         return fileName;
     }
 
+    protected String sanitizeNewlines(String before) {
+        return before.replaceAll("[\n\r]", "_");
```
sepe81, line 175:
```suggestion
        return before.replaceAll("\R", "_");
```

maybe use this as more concise way to represent line breaks?

>Any Unicode linebreak sequence, is equivalent to \u000D\u000A|[\u000A\u000B\u000C\u000D\u0085\u2028\u2029]

burtonrhodes:
@lukaszlenart  Looks good. The current version of this PR works well in my initial testing (dev) environment.

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
```diff
@@ -21,375 +21,116 @@
...
-                FileUploadSizeException ex = (FileUploadSizeException) e;
-                errorMessage = buildErrorMessage(e, new Object[]{
-                        ex.getPermitted(), ex.getActualSize()
-                });
-            } else if (e instanceof FileUploadContentTypeException) {
-                FileUploadContentTypeException ex = (FileUploadContentTypeException) e;
-                errorMessage = buildErrorMessage(e, new Object[]{
-                        ex.getContentType()
-                });
+        for (DiskFileItem item : servletFileUpload.parseRequest(request)) {
```
burtonrhodes, line 52: @lukaszlenart Quick thought on the maxFiles setting: I still think it makes more sense that maxFiles would apply to only file fields and not all form fields in the request. However it appears that this logic is actually baked into the JakartaServletFileUpload class itself when calling `servletFileUpload.parseRequest()`. Wonder why they coded it this way??

Anyway, if we ever wanted to change this, you could set the maxFiles to unlimited in the servletFileUpload object, and then perform the files count check in the `if (it.isFormFIeld()...` logic below.  Just a thought.

burtonrhodes: In thinking about it more, that might be a security issue, as the serletFileUpload object would upload everything first (something that the setting is trying to prevent), only to perform the check after the processing is done.  Ok, never mind. ;-)

lukaszlenart: I assume so, this can be changed with a proper implementation of Streaming API which can then calculate the max size of uploaded files internally in Struts ... hm or maybe even right now, let me think this through 🤔

lukaszlenart: @burtonrhodes introduced a new constant `maxSizeOfFiles` which controls how many files you can upload based on their size - this works with `JakartaStreamMultiPartRequest` only.

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaStreamMultiPartRequest.java
```diff
@@ -52,309 +51,159 @@
...
-            }
-        } catch (IOException e) {
-            if (file != null) {
-                try {
-                    file.delete();
-                } catch (SecurityException se) {
-                    LOG.warn("Failed to delete '{}' due to security exception above.", file.getName(), se);
-                }
-            }
+        Long currentFilesSize = actualSizeOfUploadedFiles();
```
kusalk, line 202: This seems expensive to compute from scratch every time a file field is encountered, especially when the application might not have even configured a `maxSizeOfFiles` limit!

kusalk: Tbh I'm not convinced this option adds much value in the first place - I think `maxSize` serves this need well enough. I'm trying to think in what situation would an application need to configure `maxSizeOfFiles` instead of or in addition to `maxSize`. They want to accepts lots of big form fields but want to limit the amount of data written to disk? I guess it's possible

core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -142,18 +142,25 @@ public final class StrutsConstants {
     /** The maximum size of a multipart request (file upload) */
     public static final String STRUTS_MULTIPART_MAXSIZE = "struts.multipart.maxSize";
 
+    /** The maximum size of all uploaded files.
+       Used only with {@link org.apache.struts2.dispatcher.multipart.JakartaStreamMultiPartRequest} */
+    public static final String STRUTS_MULTIPART_MAXSIZE_OF_FILES = "struts.multipart.maxSizeOfFiles";
```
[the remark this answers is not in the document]
lukaszlenart: IDEA doesn't complain about `MAXSIZE` that's why I left it, but you are right :)

core/src/main/java/org/apache/struts2/action/UploadedFilesAware.java
```diff
@@ -27,14 +27,14 @@
  * The {@link org.apache.struts2.interceptor.ActionFileUploadInterceptor} will use the interface
  * to notify action about the multiple uploaded files.
  */
-public interface UploadedFilesAware {
+public interface UploadedFilesAware<T> {
```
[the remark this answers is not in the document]
kusalk: Oh okay - yeah I just don't know if it makes sense for a marker interface to have a generic type - might just need 2 marker interfaces instead

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaStreamMultiPartRequest.java
lukaszlenart: > This seems expensive to compute from scratch every time a file field is encountered, especially when the application might not have even configured a `maxSizeOfFiles` limit!

Fixed!

core/src/main/java/org/apache/struts2/action/UploadedFilesAware.java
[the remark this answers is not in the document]
lukaszlenart: Reverted

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaStreamMultiPartRequest.java
```diff
@@ -52,309 +51,159 @@
...
-            }
-        } catch (IOException e) {
-            if (file != null) {
-                try {
-                    file.delete();
-                } catch (SecurityException se) {
-                    LOG.warn("Failed to delete '{}' due to security exception above.", file.getName(), se);
-                }
-            }
+        Long currentFilesSize = maxSizeOfFiles != null ? actualSizeOfUploadedFiles() : null;
```
kusalk, line 202: You can just inline this right?
`if (maxSizeOfFiles != null && actualSizeOfUploadedFiles() + file.length() >= maxSizeOfFiles) {`

kusalk (approved):
Lot of refactoring here so tricky to follow all the changes, but looks good at a glance

burtonrhodes:
Just FYI, I implemented this PR on a small subset of my server cluster and have been receiving the following ambiguous errors in some of the upload requests.  I can't duplicate it in any testing environment, but curious if you might have a clue as to why Struts is creating an actionError with no value for the file.  For this implementation, I am **_not_** using the stream parser

_Error printed by the code below_

```
Unidentified ActionError present for file upload: [Error uploading: {0}!, Error uploading: {0}!]
```

_Code that quickly checks for any file upload errors_

```
if (hasActionErrors()) {
    List<String> uploadErrors = (List<String>) getActionErrors();
    log.error("Unidentified ActionError present for file upload: " + uploadErrors);
}
```

_A redacted version of the action_


```
public class UploadLocalFileJsonAction extends ActionSupport implements Validateable, UploadedFilesAware<UploadedFile<File>> {

    private FileAttachmentUpload fileAttachmentUpload;

    @Override
    public void withUploadedFiles(List<UploadedFile<UploadedFile<File>>> uploadedFiles) {
        if (!uploadedFiles.isEmpty()) {
            fileAttachmentUpload = UploadedFileHelper.createFromUploadedFile(uploadedFiles.get(0));
        }
    }

    public void validate() {
        // First check to see if there are any errors from the file upload
        if (hasActionErrors()) {
            List<String> uploadErrors = (List<String>) getActionErrors();
            log.error("Unidentified ActionError present for file upload: " + uploadErrors);
        }
    }

    public String execute() throws Exception {
        // Omitted for brevity
    }
}
```

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaStreamMultiPartRequest.java
lukaszlenart: Not really as it is used inside the `if` clause in `exceedsMaxSizeOfFiles`

lukaszlenart:
@burtonrhodes I tried to play with the Showcase app and right now there is a problem uploading large files - I get blank page, yet the app is using the new Action Upload Interceptor

core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaStreamMultiPartRequest.java
kusalk: Missed that, all good :)

lukaszlenart:
@burtonrhodes did you have a chance to re-test the latest changes? Everything works fine when testing with the Showcase app yet it uses the new action upload interceptor, but still the same MultiPartParser is used behind the scene. I even added a dedicated integration test to cover max files constraint.

burtonrhodes:
@lukaszlenart yes, all looks good on my end.

lukaszlenart:
LGTM 💯

core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -171,4 +298,109 @@ protected String getCanonicalName(final String originalFileName) {
...
+    public Enumeration<String> getParameterNames() {
+        return Collections.enumeration(parameters.keySet());
+    }
+
+    /* (non-Javadoc)
+     * @see org.apache.struts2.dispatcher.multipart.MultiPartRequest#getParameterValues(java.lang.String)
+     */
+    public String[] getParameterValues(String name) {
+        return parameters.getOrDefault(name, Collections.emptyList())
+                .toArray(String[]::new);
```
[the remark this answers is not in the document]
lukaszlenart: @criderp please register this bug in [JIRA](https://issues.apache.org/jira/browse/WW), thanks!

PR #862, opened 2024-01-24
Stops running sonar.yml on forks

lukaszlenart:


sepe81 (approved):
Thank you for the quick solution.

PR #865, opened 2024-01-29
WW-5352 Ensure StrutsParameter compatibility with TypeConversion setters

kusalk:
WW-5352
--


lukaszlenart:
This should work without a converter, the `Model` has the default constructor and Struts uses OGNL logic to create and extend lists

kusalk:
So I tried testing this, and it seems without a matching getter, the list will not be populated. It will simply be set to an empty list.

The current `@StrutsParameter` logic only checks for annotations on getters when trying to set a collection or model class. My concern was that it might be possible to do this with only a setter. But if not, we can safely close this PR.

kusalk:
I've just had a read of the following TypeConverter test and it seems it is only possible to set a List without a matching getter when using a serialized format.
`core/src/test/java/com/opensymphony/xwork2/test/annotations/PersonActionTest.java`

So in the earlier example I gave, the request would have to be a custom format as defined in the type converter, e.g.:
`example.action?models=kusal:lukasz`

So I think we are indeed okay to close this PR.

PR #867, opened 2024-02-02
WW-5391 Add interface for VelocityManager extension point

kusalk:
WW-5391
--
Allows applications to provide a completely custom VelocityManager

### DIFF
--- a/plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiConfigurationProvider.java
+++ b/plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiConfigurationProvider.java
@@ -24,15 +24,15 @@ import com.opensymphony.xwork2.ObjectFactory;
 import com.opensymphony.xwork2.config.Configuration;
 import com.opensymphony.xwork2.config.ConfigurationException;
 import com.opensymphony.xwork2.config.PackageProvider;
-import com.opensymphony.xwork2.config.entities.PackageConfig;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.util.finder.ClassLoaderInterface;
-import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.LogManager;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.struts2.osgi.host.OsgiHost;
 import org.apache.struts2.osgi.loaders.VelocityBundleResourceLoader;
 import org.apache.struts2.views.velocity.VelocityManager;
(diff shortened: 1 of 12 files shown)


plugins/osgi/src/main/java/org/apache/struts2/osgi/OsgiConfigurationProvider.java
```diff
@@ -255,16 +254,27 @@ public void setBundleAccessor(BundleAccessor acc) {
     }
 
     @Inject
-    public void setVelocityManager(VelocityManager vm) {
-        LOG.trace("OSGi ConfigurationProvider - setVelocityManager() called - VelocityManager: [{}]", vm);
-
+    public void setVelocityManager(VelocityManagerInterface vmi) {
+        LOG.trace("OSGi ConfigurationProvider - setVelocityManager() called - VelocityManager: [{}]", vmi);
+        if (!(vmi instanceof VelocityManager)) {
```
kusalk, line 259: Didn't bother with a marker interface here as this plugin is deprecated anyway

plugins/portlet/src/main/java/org/apache/struts2/portlet/result/PortletVelocityResult.java
kusalk, line 1: Standardised line endings to `LF`

plugins/velocity/src/main/java/org/apache/struts2/views/velocity/VelocityManagerInterface.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.velocity.context.Context;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * @since 6.4.0
+ */
+public interface VelocityManagerInterface {
```
kusalk, line 32: Maybe in 7.0 we can consider renaming `VelocityManager` to `StrutsVelocityManager` and `VelocityManagerInterface` to `VelocityManager`

lukaszlenart (commented):
Looks good, are you going to address some of the reported issues by Sonar?

kusalk:
@lukaszlenart Sonar results don't make any sense, it's been misconfigured ever since the 7.0 branch was introduced

lukaszlenart:
Should be fixed

kusalk:
@lukaszlenart Thanks! Looks like it's just complaining about deprecations, serialisation and code coverage. I've fixed the serialisation warnings. The code coverage I don't think it makes sense for me to rectify as I've just done a find and replace on injection methods. Adding coverage to those injection methods alone won't add much value, and it's out of scope for me to add coverage to the whole class.

lukaszlenart (approved):

PR #873, opened 2024-02-11
[WW-5388] Fixes a few issues in Servlet 6 file upload usage

lukaszlenart:
- removes duplicated validation messages where both interceptors have been used
- defines a proper error message on a general file upload exception
- removes generic definition in tests

Follow ups #861
Refs [WW-5388](https://issues.apache.org/jira/browse/WW-5388)


lukaszlenart:
//cc: @burtonrhodes

burtonrhodes:
Thanks for all your work here.  Limited testing on my dev machine looks good. I have uploaded this patch to a subset our server cluster and will report back with any issues I see.

lukaszlenart:
Are we good here?

burtonrhodes:
@lukaszlenart Every there seems to working well on my end.  This gets my approval

PR #876, opened 2024-02-14
s:file shows server/file location WW-5396

gregh3269:
After a file upload with the s:file tag it gives details of the webapp location in the value attribute.


lukaszlenart:
Wouldn't be better to simple override the `setValue()` in the `File.java`?
```java
    @Override
    @StrutsTagAttribute(description="Ignored during file upload")
    public void setValue(String value) {
        // ignores provided value
    }
```

gregh3269:
OK, good idea, we can see what it does.  I will amend the PR.

gregh3269:
I ran the FileTest.java and it passed.

lukaszlenart:
Please add re-generated `file-attributes.html` as well

PR #898, opened 2024-03-27
WW-5251 Reinstate deleted interfaces with transparent compat

kusalk:
Partially reverts https://github.com/apache/struts/pull/670 to restore compatibility with deprecated interfaces
Refs [WW-5251](https://issues.apache.org/jira/browse/WW-5251)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.interceptor;
+
+import org.apache.struts2.dispatcher.HttpParameters;
+
+import java.util.Map;
+
+@Deprecated
+public interface ParameterAware extends org.apache.struts2.action.ParametersAware {
+
+    void setParameters(Map map);
+
+    @Override
+    default void withParameters(HttpParameters parameters) {
+        setParameters(parameters);
+    }
+}
--- a/core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
@@ -22,6 +22,8 @@ import org.apache.struts2.dispatcher.HttpParameters;
 
 import java.util.Map;
 
+import static java.util.stream.Collectors.toMap;
+
 @Deprecated
 public interface ParameterAware extends org.apache.struts2.action.ParametersAware {
 
@@ -29,6 +31,6 @@ public interface ParameterAware extends org.apache.struts2.action.ParametersAwar
 
     @Override
     default void withParameters(HttpParameters parameters) {
-        setParameters(parameters);
+        setParameters(parameters.entrySet().stream().collect(toMap(Map.Entry::getKey, e -> e.getValue().getMultipleValues())));
     }
 }
--- a/core/src/main/java/org/apache/struts2/interceptor/ApplicationAware.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ApplicationAware.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.interceptor;
+
+import java.util.Map;
+
+@Deprecated
+public interface ApplicationAware extends org.apache.struts2.action.ApplicationAware {
+
+    void setApplication(Map<String, Object> application);
+
+    @Override
+    default void withApplication(Map<String, Object> application) {
+        setApplication(application);
+    }
+}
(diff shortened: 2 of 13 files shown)


kusalk:
@brianandle If you bundle these classes with your Struts application, it should restore compatibility.

core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Map;
+
+@Deprecated
+public interface ParameterAware extends org.apache.struts2.action.ParametersAware {
+
+    void setParameters(Map map);
+
+    @Override
+    default void withParameters(HttpParameters parameters) {
+        setParameters(parameters);
```
brianandle, line 32: This would need to be something like the following since the setParameters takes a Map:

```default void withParameters(HttpParameters parameters) {
       Map<String, String[]> newParams = new HashMap<String, String[]>();
        for(String key : parameters.keySet()){
            Parameter p = parameters.get(key);
            if(p instanceof Parameter.Request){
            	if(p.isMultiple()) {
            		newParams.put(key, p.getMultipleValues());
            	} else {
            		newParams.put(key, new String[] {p.getValue()});
            	}
            }
        }
       setParameters(newParams);

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: We're also missing the restoration of org.apache.struts2.interceptor.RequestAware which would be something like:

```@Deprecated
public interface RequestAware extends org.apache.struts2.action..ServletRequestAware {

   void setRequest(Map<String,Object> request);

    default void withServletRequest(HttpServletRequest request) {
		setRequest(request.getParameterMap());
	}
}

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: I created https://issues.apache.org/jira/browse/WW-5403 as well which highlights some other classes/methods I found that (probably not just the ones listed) that were removed in a minor when they really, IMHO, should have been removed only in the next major (7.x).

I would worry what other newly deprecated, in 6.x, classes/methods were removed in a minor vs waiting until 7.x. Unless the class/method exposed a significant security risk on it's own.

**Review by brianandle — CHANGES_REQUESTED** [review]

File: core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java (hunk 1 above)
Comment (kusalk) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: As far as I know Struts has never adhered to SemVer. I noticed deprecations and deletions happening in successive minor versions when I first began contributing so I just went with it too. Maybe with 7.0.0 we can start adhering to it strictly if this is what the community/users would prefer?

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: From Struts 6.0.0 announcement page.

Version change

You can be surprised by the version change, previously we have been using Struts 2.5.x versioning schema, but this was a bit misleading. Struts 2 is a different framework than Struts 1 and its versioning is supposed to start with 1.0.0, yet that never happened. With each breaking changes release (like Struts 2.5), we had been only upgrading the MINOR part of the versioning schema. To fix that problem as from Struts 2 ver. 6.0.0 (aka Struts 2.6) we adopt a proper SemVer to avoid such confusion.

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: Given S2-066 was only fixed in 6.3.0.2 and not other breakage points would also seem to enforce that Struts has truly moved to SemVer starting with 6.0

Comment (kusalk) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: SemVer also definitely doesn't have a 4th version component but here we are with version 6.3.0.2 🤷🏾

I think we've probably deleted too much code to be able to reinstate full API compatibility with version 6.0.0 at this point in time. Technically all `protected`+ methods defined in Struts extension point classes are also public API but these have also been aggressively refactored. But we can try reinstate some of the most used APIs as per your Jira card, at least the ones we can verify to not compromise security.

Also, adding a new dependency (Caffeine) isn't a SemVer violation as far as I know.

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: Yup, the 0.3 confused me as well as it deviate from SemVer but I assumed it was just habit. :)

The caffeine was more of a note of what I found as I tried to get our app running on latest. Since 6.4 hasn't been released I didn't have release notes to dig into.

Seems like almost all the breakages I hit were introduced in 6.2.

Maybe your approach icould be used and bundled together as a compatibility plugin/seperate jar?  That might not be possible in all cases but maybe helps the teams goals of deprecation "removal" while abstraction into its own jar.

My largest concern based on past historical Struts usage is I expect a lot of usages sitting on 2.5 after EOL and when a new security issue is released scrambling and hitting similar upgrade issues and likely not being able to upgrade in a timely manor.

I do want to say I deeply appreciate the teams time and effort on the ongoing maintenance and improvements of Struts.

Comment (kusalk) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: Yes that was the approach (bundled compatibility jar) I took when upgrading my Struts application. This technique can be used for the renamed marker interfaces only however. The remaining API changes are easy enough to find and replace in your codebase.

I'll leave it to @lukaszlenart to decide whether we want to restore full API compatibility with Struts 6.0.0. Given you're upgrading from 2.5 and you're the first (as far as I know) to complain, I'm inclined to say no.

However, I'm very much in favour of ensuring there are no more SemVer violations henceforth.

Comment (lukaszlenart) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: Using the four numbers format is an old approach to marking a given release a security hotfix. Anyway, restoring an old behaviour can be dangerous. I'm fine with restoring the old interfaces, yet I would be more careful brining back changes related to ActionContext - this is was a source of lot of security vulnerabilities.

Comment (brianandle) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: That makes sense to not revert all of the ActionContext change. It might make sense to add back getName(), since it was never marked deprecated, marking deprecated and removing in 7.0.0.

I agree that reverting something that would re-expose a security issue wouldn't make sense. If the revert in https://github.com/apache/struts/pull/897 was done for 6.4.0.0 then from our implementation we'd be good to go. I created WW-5403 thinking about the community at large and that it could cause adoption issues.

Comment (kusalk) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java:32]: Let me fix `ParameterAware` and add `RequestAware` to this PR

Comment (kusalk) [reply core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java]: @brianandle I believe this PR should be equivalent to #897 now - let me know if there's any issues

File: core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.struts2.dispatcher.HttpParameters;
+
+import java.util.Map;
+
+import static java.util.stream.Collectors.toMap;
+
+@Deprecated
+public interface ParameterAware extends org.apache.struts2.action.ParametersAware {
+
+    void setParameters(Map map);
```
brianandle, line 30: To match the existing signature, which caused build issues for us, this needs to be:

setParameters(Map<String, String[]> parameters);

brianandle (changes requested):

core/src/main/java/org/apache/struts2/interceptor/ParameterAware.java
kusalk: Ah good catch, not sure how I managed to drop the generic types here

brianandle: I did check the other files in the review and it was just this one.

We were able to build fine and it's going through out automation, what we have anyway, tonight.

lukaszlenart:
Sorry, I merged wrong PR :( Should I revert it?

brianandle:
@lukaszlenart  I believe it's the correct one, this seems like a better approach vs the revert, but the ParameterAware still needs an additional change by @kusalk. Or you could fix the method signature (see comment)

brianandle:
Thanks @kusalk that other PR looks good to me. Commented there too

PR #910, opened 2024-04-08
WW-5406 Ensure Action excluded patterns are reinjected

kusalk:
WW-5406
--


kusalk:
Ignore SonarCloud results, they are incorrect again

lukaszlenart:
> Ignore SonarCloud results, they are incorrect again

Probably because the last scan was for Struts 7 branch, I will fix that

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
```diff
@@ -340,6 +348,27 @@ public void setMultipartValidationRegex(String multipartValidationRegex) {
...
+    private static List<Pattern> buildExcludedPatternsList(String patterns, String separator) {
+        if (patterns == null || patterns.trim().isEmpty()) {
+            return emptyList();
+        }
+        return unmodifiableList(Arrays.stream(patterns.split(separator)).map(String::trim).map(Pattern::compile).collect(toList()));
+    }
+
+    public List<Pattern> getActionExcludedPatterns() {
+        return actionExcludedPatterns;
+    }
```
lukaszlenart, line 370: It's getting crowd'y here, what about moving all the injectable options into `DispatcherOptions`? It doesn't have to happen now, but I can create a ticket to address that later

core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
```diff
@@ -43,6 +43,8 @@ public class StrutsPrepareFilter implements StrutsStatics, Filter {
     protected static final String REQUEST_EXCLUDED_FROM_ACTION_MAPPING = StrutsPrepareFilter.class.getName() + ".REQUEST_EXCLUDED_FROM_ACTION_MAPPING";
 
     protected PrepareOperations prepare;
+
+    @Deprecated
```
lukaszlenart, line 47: Please document why it's deprecated and what to use instead. Also `since` should be added, and feel free to create a ticket in JIRA to remove this deprecated element in Struts 7. [An example deprecation](https://github.com/apache/struts/blob/master/core/src/main/java/org/apache/struts2/interceptor/FileUploadInterceptor.java#L176)

lukaszlenart (changes requested):

core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
kusalk: Yep let me rectify
Regarding a ticket for the deletion - my assumption would be that all deprecated elements are deleted in the next major version. Do we still require individual follow-up issues in the case of a deprecation?

core/src/main/java/org/apache/struts2/dispatcher/Dispatcher.java
kusalk: Yep I think that's a fair call, I'll create an issue for it

kusalk: WW-5410

core/src/main/java/org/apache/struts2/dispatcher/filter/StrutsPrepareFilter.java
lukaszlenart: We can have one general ticket to remove all the deprecated options, just to be sure we won't miss this one at some point (so `since` would be needed)

kusalk: Fixed and created this general issue - WW-5411

kusalk:
Ah wait, just realised I made a breaking change in PrepareOperations.java, let me fix that

kusalk:
Done

lukaszlenart (approved):

PR #911, opened 2024-04-09
WW-5407 Extend SecurityMemberAccess proxy detection to other proxies

jefferyxhy:
WW-5407 
--
**Reason**
Currently `SecurityMemberAccess#isAccessible` return true for a method of a proxy object, which expose the beans at risk of being changed. We need to have the ability to detect proxy object and reject the access if required.

See Jira card above for more details.

&nbsp;
 
**Changes/ Solution**
currently in  `isAccessible -> checkProxyMemberAccess` , it use `disallowProxyMemberAccess && ProxyUtil.isProxyMember(member, target)`  which is not enough, as `isProxyMember` only matches the member **directly** from proxy class, and does not match those ones **original** from the target class.

 &nbsp;
 
So we update the `isAccessible`:
* add one extra checking `checkProxyAccess` before `checkProxyMemberAccess` which is controlled by:
    * `struts.disallowProxyObjectAccess` : an new struts constant to enable or disable this checking (default as `false`)
    * if disallow, then we do the proxy checking against  the target object.
 
 &nbsp;
 
**Result & Impact**
* By default `struts.disallowProxyObjectAccess` as `default`, no difference.
* Set `struts.disallowProxyObjectAccess`  as `true`, access to any member of a proxy object will be rejected, including both proxy member and original member of class. Which means whenever chained parameter `a.b.c.d.x` has one part that is a proxy, we reject the set to the last `x`

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -230,6 +230,15 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
         </dependency>
+
+        <!-- Optional used in com.opensymphony.xwork2.util.ProxyUtil to detect if object is HibernateProxy -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>5.6.15.Final</version>
+            <optional>true</optional>
+        </dependency>
+
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
--- a/core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java
+++ b/core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java
@@ -18,13 +18,20 @@
  */
 package com.opensymphony.xwork2.util;
 
+import com.opensymphony.xwork2.ognl.DefaultOgnlCacheFactory;
+import com.opensymphony.xwork2.ognl.OgnlCache;
+import com.opensymphony.xwork2.ognl.OgnlCacheFactory;
 import org.apache.commons.lang3.reflect.ConstructorUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.commons.lang3.reflect.MethodUtils;
+import org.hibernate.proxy.HibernateProxy;
 
-import java.lang.reflect.*;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Member;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Proxy;
 
 /**
  * <code>ProxyUtil</code>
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java
```diff
@@ -18,13 +18,20 @@
  */
 package com.opensymphony.xwork2.util;
 
+import com.opensymphony.xwork2.ognl.DefaultOgnlCacheFactory;
+import com.opensymphony.xwork2.ognl.OgnlCache;
+import com.opensymphony.xwork2.ognl.OgnlCacheFactory;
 import org.apache.commons.lang3.reflect.ConstructorUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.commons.lang3.reflect.MethodUtils;
+import org.hibernate.proxy.HibernateProxy;
```
lukaszlenart, line 27: This isn't acceptable solution as it means everybody will have to use Hibernate just to satisfy this dependency.

kusalk: The dependency is optional and the check silently skipped :)

core/pom.xml
```diff
@@ -230,6 +230,15 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
         </dependency>
+
+        <!-- Optional used in com.opensymphony.xwork2.util.ProxyUtil to detect if object is HibernateProxy -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>6.4.4.Final</version>
+            <optional>true</optional>
```
kusalk, line 239: @lukaszlenart

core/pom.xml
```diff
@@ -230,6 +230,15 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
         </dependency>
+
+        <!-- Optional used in com.opensymphony.xwork2.util.ProxyUtil to detect if object is HibernateProxy -->
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>6.4.4.Final</version>
```
lukaszlenart, line 238: As far I know this version won't work on JDK8 which is a base version for Struts 6.x

core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java
lukaszlenart: Sorry, missed the "skip" :/

core/pom.xml
kusalk: You are correct - downgraded to Hibernate 5 as this proxy detection code is identical :)

kusalk:
Hi @lukaszlenart 

@jefferyxhy is a colleague of mine at Atlassian who is helping implement some security enhancements.
I believe this one is ready for review now. I've given it a look over and it looks good to me. Let us know if there are any issues otherwise we can prepare an accompanying struts-site PR

core/src/main/java/com/opensymphony/xwork2/util/ProxyUtil.java
```diff
@@ -96,12 +105,41 @@ public static boolean isProxyMember(Member member, Object object) {
...
+    /**
+     * Check whether the given member is a member of a Hibernate proxy.
+     *
+     * @param member the member to check
+     */
+    public static boolean isHibernateProxyMember(Member member) {
+        try {
+            Class<?> clazz = ClassLoaderUtil.loadClass(HIBERNATE_HIBERNATEPROXY_CLASS_NAME, ProxyUtil.class);
+            if (hasMember(clazz, member))
+                return true;
```
lukaszlenart, line 136: Shouldn't be inlined?
```java
return hasMember(clazz, member);
```

jefferyxhy: @lukaszlenart updated. Thanks

lukaszlenart (approved):
Great 👏  LGTM 👍

PR #912, opened 2024-04-09
WW-5408 add option to not fallback to empty namespace when unresolved

jefferyxhy:
WW-5408

**Reason**
There is a case we want to enhance to improve security for struts url mapping: Be able to prevent arbitrary namespace fallback to empty (root) namespace when not match

&nbsp;
 
**Changes/ Solution**
*  introduce new Struts constant ` STRUTS_DISABLE_EMPTY_NAMESPACE_FALLBACK [struts.disableActionConfigFallbackToEmptyNamespace]` to disable fallback to empty namespace when not match
*  default as null which means does not disable (Need to manually set as true in confluence struts.xml)

 &nbsp;
 
**Result & Impact**
* By default `struts.disableActionConfigFallbackToEmptyNamespace` is `null`, no difference.
* Set `struts.disableActionConfigFallbackToEmptyNamespace`  as `true`, not matched namepsace **WILL NOT** fallback to empty namespace anymore, and struts will threw `ConfigurationException`, it relies on the application to handle this exception.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
+++ b/core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
@@ -120,6 +120,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -459,12 +460,12 @@ public class DefaultConfiguration implements Configuration {
         boolean appendNamedParameters = Boolean.parseBoolean(
                 container.getInstance(String.class, StrutsConstants.STRUTS_MATCHER_APPEND_NAMED_PARAMETERS)
         );
-        boolean emptyNamespaceFallbackDisabled = Boolean.parseBoolean(
-                container.getInstance(String.class, StrutsConstants.STRUTS_DISABLE_EMPTY_NAMESPACE_FALLBACK)
+        boolean fallbackToEmptyNamespace = Boolean.parseBoolean(
+                Optional.ofNullable(container.getInstance(String.class, StrutsConstants.STRUTS_ACTION_CONFIG_FALLBACK_TO_EMPTY_NAMESPACE)).orElse("true")
         );
 
         return new RuntimeConfigurationImpl(Collections.unmodifiableMap(namespaceActionConfigs),
-                Collections.unmodifiableMap(namespaceConfigs), matcher, appendNamedParameters, emptyNamespaceFallbackDisabled);
+                Collections.unmodifiableMap(namespaceConfigs), matcher, appendNamedParameters, fallbackToEmptyNamespace);
     }
 
     private void setDefaultResults(Map<String, ResultConfig> results, PackageConfig packageContext) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/struts2/StrutsConstants.java
```diff
@@ -230,6 +230,8 @@ public final class StrutsConstants {
     public static final String STRUTS_XWORKCONVERTER = "struts.xworkConverter";
 
     public static final String STRUTS_ALWAYS_SELECT_FULL_NAMESPACE = "struts.mapper.alwaysSelectFullNamespace";
+    /** Disable fallback to empty namespace when request namespace didn't match any in action configuration */
+    public static final String STRUTS_DISABLE_EMPTY_NAMESPACE_FALLBACK = "struts.disableActionConfigFallbackToEmptyNamespace";
```
kusalk, line 234: @lukaszlenart Any thoughts on what the best name for this constant might be?

lukaszlenart: Wouldn't be better to have `struts.actionConfig.fallbackToEmptyNamespace` with default value set to `true` instead?

kusalk: Sounds reasonable to me - @jefferyxhy could you please update?

jefferyxhy: @kusalk @lukaszlenart updated. Thanks

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -583,11 +590,10 @@ public ActionConfig getActionConfig(String namespace, String name) {
             }
 
             // fail over to empty namespace
-            if (config == null && StringUtils.isNotBlank(namespace)) {
+            if (config == null && StringUtils.isNotBlank(namespace) && ("/".equals(namespace) || fallbackToEmptyNamespace)) {
```
lukaszlenart, line 593: I would extract this logic into a private method to name it, something like
```java
if (config == null && shouldFallbackToEmptyNamespaces(namespace)) {
    ...
```

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
```diff
@@ -459,9 +460,12 @@ protected synchronized RuntimeConfiguration buildRuntimeConfiguration() throws C
         boolean appendNamedParameters = Boolean.parseBoolean(
                 container.getInstance(String.class, StrutsConstants.STRUTS_MATCHER_APPEND_NAMED_PARAMETERS)
         );
+        boolean fallbackToEmptyNamespace = Boolean.parseBoolean(
+                Optional.ofNullable(container.getInstance(String.class, StrutsConstants.STRUTS_ACTION_CONFIG_FALLBACK_TO_EMPTY_NAMESPACE)).orElse("true")
```
lukaszlenart, line 464: Instead of this fallback I would create a new entry into [default.properties](https://github.com/apache/struts/blob/master/core/src/main/resources/org/apache/struts2/default.properties) with value set to `true` and document it - this will help others understand the change

lukaszlenart (commented):
Looks good, two minor issues

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
kusalk: @jefferyxhy and I just discussed this one and one of the drawbacks of putting it in `default.properties` is that it isn't read by unit tests and causes a bunch of test failures, as the unit tests will default to `false`. To get around this we could additionally add the constant to `StrutsDefaultConfigurationProvider` as well as `default.properties`.

I'm personally not too fussed. In the past I've deliberately made constants default to false to try sidestep this issue. Let us know know what you would prefer

lukaszlenart: Understood, yet `default.properties` is used as a documentation by users so I would define the constant there with short description and keep fallback to `true` as above

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
jefferyxhy: Done. Thanks

core/src/main/java/com/opensymphony/xwork2/config/impl/DefaultConfiguration.java
jefferyxhy: Updated. Thanks

lukaszlenart (approved):

PR #913, opened 2024-04-10
WW-5400 Extend default configuration options for the CSP interceptor.

eschulma:
Previously, it was impossible to set global options for the CSP interceptor. The only option was to have every action individually implement CspSettingsAware.

To fix this, we add an interceptor parameter of defaultCspSettingsClassName. Values from this class will be used in the CSP header instead of DefaultCspSettings. Users may define their own custom class which implements CspSettings, and that will be the default for all actions that do not implement the CspSettingsAware interface. It is now possible to create this custom class by simply extending DefaultCspSettings.

I have fixed a spelling error in DefaultCspSettings.java -- cratePolicyFormat renamed to createPolicyFormat.

Closes [WW-5400](https://issues.apache.org/jira/browse/WW-5400)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
@@ -46,6 +46,9 @@ public final class CspInterceptor extends AbstractInterceptor {
     private boolean prependServletContext = true;
     private boolean enforcingMode;
     private String reportUri;
+    private String reportTo;
+
+    private String defaultCspSettingsClassName = DefaultCspSettings.class.getName();
 
     @Override
     public String intercept(ActionInvocation invocation) throws Exception {
@@ -54,8 +57,24 @@ public final class CspInterceptor extends AbstractInterceptor {
             LOG.trace("Using CspSettings provided by the action: {}", action);
             applySettings(invocation, ((CspSettingsAware) action).getCspSettings());
         } else {
-            LOG.trace("Using DefaultCspSettings with action: {}", action);
-            applySettings(invocation, new DefaultCspSettings());
+            LOG.trace("Using {} with action: {}", defaultCspSettingsClassName, action);
+
+            // if the defaultCspSettingsClassName is not a real class, throw an exception
+            try {
+                Class.forName(defaultCspSettingsClassName, false, Thread.currentThread().getContextClassLoader());
+            }
+            catch (ClassNotFoundException e) {
+                throw new IllegalArgumentException("The defaultCspSettingsClassName must be a real class.");
+            }
+
+            // if defaultCspSettingsClassName does not implement CspSettings, throw an exception
+            if (!CspSettings.class.isAssignableFrom(Class.forName(defaultCspSettingsClassName))) {
+                throw new IllegalArgumentException("The defaultCspSettingsClassName must implement CspSettings.");
+            }
+
+            CspSettings cspSettings = (CspSettings) Class.forName(defaultCspSettingsClassName)
+                    .getDeclaredConstructor().newInstance();
+            applySettings(invocation, cspSettings);
         }
         return invocation.invoke();
     }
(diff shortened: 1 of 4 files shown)


eschulma:
@lukaszlenart submitted per your request

eschulma:
Hold off a bit, I need to check something (this is what I get for implementing my own separate solution)

eschulma:
Ok all good.

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
```diff
@@ -124,4 +153,11 @@ public void setPrependServletContext(boolean prependServletContext) {
...
     }
 
-}
+    /**
+     * Sets the class name of the default {@link CspSettings} implementation to use when the action does not
+     * set its own values. If not set, the default is {@link DefaultCspSettings}.
+     */
+    public void setDefaultCspSettingsClassName(String defaultCspSettingsClassName) {
+        this.defaultCspSettingsClassName = defaultCspSettingsClassName;
+    }
```
lukaszlenart, line 162: You can use Struts inject mechanism instead of using raw class and creating the instance by yourself. It's all about defining a `<bean name="customCspSettings" class="..."/>` and then annotating the setter with `@Inject("customCspSettings")`.

I assume you never played with Struts @Inject, so let's leave it as is and I will change that in the next PR.

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
```diff
@@ -124,4 +153,11 @@ public void setPrependServletContext(boolean prependServletContext) {
         this.prependServletContext = prependServletContext;
     }
 
-}
+    /**
+     * Sets the class name of the default {@link CspSettings} implementation to use when the action does not
+     * set its own values. If not set, the default is {@link DefaultCspSettings}.
+     */
```
lukaszlenart, line 159: Please add `@since Struts 6.5.0` [annotation](https://www.oracle.com/pl/technical-resources/articles/java/javadoc-tool.html#@since)

core/src/main/java/org/apache/struts2/interceptor/csp/CspSettings.java
```diff
@@ -56,6 +57,11 @@ public interface CspSettings {
      */
     void setReportUri(String uri);
 
+    /**
+     * Sets the report group where csp violation reports will be sent
+     */
```
lukaszlenart, line 62: Could you add [@since Struts 6.5.0](https://www.oracle.com/pl/technical-resources/articles/java/javadoc-tool.html#@since)?

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
eschulma: Done

core/src/main/java/org/apache/struts2/interceptor/csp/CspSettings.java
eschulma: Done

core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
```diff
@@ -54,8 +57,24 @@ public String intercept(ActionInvocation invocation) throws Exception {
...
+            }
+
+            // if defaultCspSettingsClassName does not implement CspSettings, throw an exception
+            if (!CspSettings.class.isAssignableFrom(Class.forName(defaultCspSettingsClassName))) {
+                throw new IllegalArgumentException("The defaultCspSettingsClassName must implement CspSettings.");
+            }
+
+            CspSettings cspSettings = (CspSettings) Class.forName(defaultCspSettingsClassName)
+                    .getDeclaredConstructor().newInstance();
+            applySettings(invocation, cspSettings);
```
lukaszlenart, line 77: I wonder if we can move this code into `init()` method of the interceptor as right now a new instance is created per each invocation

lukaszlenart (approved):
LGTM and I will try to refactor it afterwards

PR #914, opened 2024-04-11
WW-5409 introduce final attribute to package element which make them unextendable

jefferyxhy:
WW-5409

 &nbsp;

**Reason**
Extending packages is a very useful capability of Struts but there are some quirks, that if a developer is not aware of, can lead to critical vulnerabilities. One such misunderstood quirk is the default-interceptor-ref element.

e.g. a parent package add permission interceptor for its action (say **Action A**), while child package extends parent package will inherit its actions but not interceptor. So if the develop is not aware of this, then Action A is now exposed with permission vuln through child package's namespace.

 &nbsp;
 
**Changes/ Solution**
introduce new `final` attribute on `package` element which will make it is unextendable

  &nbsp;
  
**Result & Impact**
* By default, package `final` attribute is implied as `false`, so no difference.
* Set package `final` attribute explicitly as `true`, will make this package unextendable, so any extends to this package will cause a ConfigurationException to be thrown during application struts config xml load step.
 
 &nbsp;

### DIFF
--- a/core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
+++ b/core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
@@ -54,6 +54,7 @@ public class StrutsXmlConfigurationProvider extends XmlConfigurationProvider {
         put("-//Apache Software Foundation//DTD Struts Configuration 2.3//EN", "struts-2.3.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 2.5//EN", "struts-2.5.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 6.0//EN", "struts-6.0.dtd");
+        put("-//Apache Software Foundation//DTD Struts Configuration 6.4.0//EN", "struts-6.4.0.dtd");
     }});
     private File baseDir = null;
     private final String filename;
--- a/core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
+++ b/core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
@@ -54,7 +54,7 @@ public class StrutsXmlConfigurationProvider extends XmlConfigurationProvider {
         put("-//Apache Software Foundation//DTD Struts Configuration 2.3//EN", "struts-2.3.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 2.5//EN", "struts-2.5.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 6.0//EN", "struts-6.0.dtd");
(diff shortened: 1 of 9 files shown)


core/src/main/resources/struts-6.4.0.dtd
```diff
@@ -0,0 +1,158 @@
...
+-->
+<!-- START SNIPPET: strutsDtd -->
+
+<!--
+   Struts configuration DTD.
+   Use the following DOCTYPE
+
+   <!DOCTYPE struts PUBLIC
+    "-//Apache Software Foundation//DTD Struts Configuration 6.4.0//EN"
+    "https://struts.apache.org/dtds/struts-6.4.0.dtd">
```
jefferyxhy, line 30: @lukaszlenart Let us know if you want this changed to 6.5.0 cc @kusalk

lukaszlenart: Yes, please. Struts 6.4.0 is out (not officially released yet) and this change will be included in 6.5.0

jefferyxhy: updated. Thanks

core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
```diff
@@ -54,6 +54,7 @@ public class StrutsXmlConfigurationProvider extends XmlConfigurationProvider {
         put("-//Apache Software Foundation//DTD Struts Configuration 2.3//EN", "struts-2.3.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 2.5//EN", "struts-2.5.dtd");
         put("-//Apache Software Foundation//DTD Struts Configuration 6.0//EN", "struts-6.0.dtd");
+        put("-//Apache Software Foundation//DTD Struts Configuration 6.5.0//EN", "struts-6.5.0.dtd");
```
lukaszlenart, line 57: I would keep two digits pattern as such changes cannot be introduced with `PATCH` version anyway. So instead of 
```java
put("-//Apache Software Foundation//DTD Struts Configuration 6.5.0//EN", "struts-6.5.0.dtd")
```
use
```java
put("-//Apache Software Foundation//DTD Struts Configuration 6.5//EN", "struts-6.5.dtd")
```
and rename the corresponding file.

---

What do you think? does it make sense?

lukaszlenart (commented):
One concern about versioning DTDs

core/src/main/java/org/apache/struts2/config/StrutsXmlConfigurationProvider.java
jefferyxhy: Yeah. It make sense. I've updated the naming accordingly. Thanks.

lukaszlenart (approved):
LGTM 💯

PR #916, opened 2024-04-18
WW-5418 Forbid Enums and Jasper classes

kusalk:
WW-5418
--


kusalk:
It is indeed - but I think the security benefits should come first in this scenario

lukaszlenart (approved):

PR #919, opened 2024-04-20
WW-5353 Stronger security defaults for 7.0

kusalk:
WW-5353
--


core/src/main/java/org/apache/struts2/util/StrutsTestCaseHelper.java
```diff
@@ -22,23 +22,26 @@
...
 
 /**
  * Generic test setup methods to be used with any unit testing framework.
  */
 public class StrutsTestCaseHelper {
 
     public static Dispatcher initDispatcher(ServletContext ctx, Map<String, String> params) {
-        Dispatcher du = new DispatcherWrapper(ctx, params != null ? params : emptyMap());
+        var finalParams = new HashMap<>(params);
+        finalParams.putIfAbsent(STRUTS_ALLOWLIST_ENABLE, "false");
```
kusalk, line 43: I think it makes sense to keep the allowlist disabled for unit tests as the auto-allowlisting is only effective in production applications

lukaszlenart (approved):

lukaszlenart:
I can roll a new Milestone release during the weekend

PR #920, opened 2024-04-21
[WW-5419] Fixes support for loading Tiles definitions

lukaszlenart:
Closes [WW-5419](https://issues.apache.org/jira/browse/WW-5419)

### DIFF
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesContainerFactory.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesContainerFactory.java
@@ -68,10 +68,7 @@ import javax.el.MapELResolver;
 import javax.el.ResourceBundleELResolver;
 import javax.servlet.jsp.JspFactory;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesContainerFactory.java
```diff
@@ -105,19 +102,8 @@ public class StrutsTilesContainerFactory extends BasicTilesContainerFactory {
...
-     *
-     * @deprecated since Struts 6.4.0, use {@link #TILES_DEFAULT_PATTERNS} instead
      */
-    @Deprecated
-    public static final String TILES_DEFAULT_PATTERN = "/WEB-INF/**/tiles*.xml,classpath*:META-INF/**/tiles*.xml";
-
-    /**
-     * Default pattern to be used to collect Tiles definitions if user didn't configure any
-     */
-    public static final Set<String> TILES_DEFAULT_PATTERNS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
```
kusalk, line 117: Technically you need to deprecate this since it's been released in 6.4.0. But also there's no need to create the set anew each time below. You can just keep this field as is and update the value like so:

```public static final Set<String> TILES_DEFAULT_PATTERNS = TextParseUtil.commaDelimitedStringToSet("*tiles*.xml");```

lukaszlenart: Right, fixed

kusalk (approved):

gregh3269:
Looks ok (for webapp/WEB-INF/*tiles*.xml)

PR #932, opened 2024-05-11
WW-5414 Always call afterInvocation even in case of exception

lukaszlenart:
Close [WW-5414](https://issues.apache.org/jira/browse/WW-5414)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/exec/StrutsBackgroundProcess.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/exec/StrutsBackgroundProcess.java
@@ -20,6 +20,8 @@ package org.apache.struts2.interceptor.exec;
 
 import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.ActionInvocation;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.Serializable;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/struts2/interceptor/exec/StrutsBackgroundProcess.java
```diff
@@ -61,11 +66,17 @@ public BackgroundProcess prepare() {
...
-                    afterInvocation();
                 } catch (Exception e) {
+                    LOG.warn("Exception during invokeActionOnly() execution", e);
                     exception = e;
                 } finally {
-                  done = true;
+                    try {
+                        afterInvocation();
+                    } catch (Exception ex) {
+                        exception = ex;
```
kusalk, line 76: Hmm what if `beforeInvocation()` already threw an exception, do we want to overwrite that?

lukaszlenart: Yeah, I know, after adding additional `afterException` and maybe `beforeException` fields to keep this information

kusalk: If we're only going to record one, the first one is probably more relevant - I'll let you make the call

kusalk (approved):

core/src/main/java/org/apache/struts2/interceptor/exec/StrutsBackgroundProcess.java
lukaszlenart: Right, let's keep it simple, I will set `exception` only if wasn't already set

PR #933, opened 2024-05-12
WW-5415 Fixes accessing public constructors via expression

lukaszlenart:
This PR fixes creating instances via OGNL expressions. The breaking change was introduced in #780

Closes [WW-5415](https://issues.apache.org/jira/browse/WW-5415)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -147,7 +147,7 @@ public class SecurityMemberAccess implements MemberAccess {
         if (target != null) {
             // Special case: Target is a Class object but not Class.class
             if (Class.class.equals(target.getClass()) && !Class.class.equals(target)) {
-                if (!isStatic(member)) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -147,7 +147,7 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
         if (target != null) {
             // Special case: Target is a Class object but not Class.class
             if (Class.class.equals(target.getClass()) && !Class.class.equals(target)) {
-                if (!isStatic(member)) {
+                if (!isStatic(member) && Arrays.stream(((Class<?>) target).getConstructors()).noneMatch(p -> p.getClass().equals(member.getClass()))) {
```
lukaszlenart, line 150: This is just a temporary solution to express where the problem is, @kusalk I count on your comment :)

kusalk: Thanks @lukaszlenart for writing the test case - I did indeed overlook the constructor case, I've pushed a commit to this PR with the appropriate fix :)

kusalk (approved):

PR #956, opened 2024-06-05
WW-5400 Simplifies how CspSettings is created

lukaszlenart:
Small refactoring of how `CspSettings` class is created plus additional tests. See #913 for more details.

Closes [WW-5400](https://issues.apache.org/jira/browse/WW-5400)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
@@ -19,7 +19,9 @@
 package org.apache.struts2.interceptor.csp;
 
 import com.opensymphony.xwork2.ActionInvocation;
+import com.opensymphony.xwork2.config.ConfigurationException;
 import com.opensymphony.xwork2.interceptor.AbstractInterceptor;
+import com.opensymphony.xwork2.util.ClassLoaderUtil;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/struts2/interceptor/csp/CspInterceptor.java
```diff
@@ -57,26 +59,28 @@ public String intercept(ActionInvocation invocation) throws Exception {
...
+        try {
+            cspSettingsClass = ClassLoaderUtil.loadClass(cspSettingsClassName, getClass());
+        } catch (ClassNotFoundException e) {
+            throw new ConfigurationException(String.format("The class %s doesn't exist!", cspSettingsClassName));
+        }
 
-            CspSettings cspSettings = (CspSettings) Class.forName(defaultCspSettingsClassName)
-                    .getDeclaredConstructor().newInstance();
-            applySettings(invocation, cspSettings);
+        if (!CspSettings.class.isAssignableFrom(Class.forName(cspSettingsClassName))) {
```
kusalk, line 78:
```suggestion
        if (!CspSettings.class.isAssignableFrom(cspSettingsClass)) {
```

kusalk (approved):

eschulma:
@lukaszlenart ok, I did figure out that there is another variable, prependServletContext, that can be set to false to prevent this behavior. I guess I question setting the default to true, particularly if someone already put in an absolute URL. But at least things can work.

PR #967, opened 2024-06-17
WW-5428 Allowlist capability should resolve Hibernate proxies when disableProxyObjects is not set

kusalk:
WW-5428
--
The OGNL allowlist capability cannot resolve proxy objects and thus will not work in applications where proxy objects such as Spring beans and Hibernate entities are accessed via OGNL.

Allowing access to such objects via OGNL is not recommended anyway as such access can be used to escalate an SSTI vulnerability.

However, given there seem to be many applications that still rely on accessing Hibernate entities directly, it would still offer much greater security in these cases, if the allowlist were enabled with exemptions for Hibernate proxies, rather than disabled altogether.

As such, in this PR, if an application is configured with `struts.disallowProxyObjectAccess=false`, the allowlist capability will resolve the underlying class of any Hibernate proxies and enforce the allowlist against this class instead.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
@@ -77,16 +77,23 @@ public class SecurityMemberAccess implements MemberAccess {
 
     private final ProviderAllowlist providerAllowlist;
     private final ThreadAllowlist threadAllowlist;
+
     private boolean allowStaticFieldAccess = true;
+
     private Set<Pattern> excludeProperties = emptySet();
     private Set<Pattern> acceptProperties = emptySet();
+
     private Set<String> excludedClasses = unmodifiableSet(new HashSet<>(singletonList(Object.class.getName())));
     private Set<Pattern> excludedPackageNamePatterns = emptySet();
     private Set<String> excludedPackageNames = emptySet();
     private Set<String> excludedPackageExemptClasses = emptySet();
+
+    private boolean isDevMode;
+
     private boolean enforceAllowlistEnabled = false;
     private Set<Class<?>> allowlistClasses = emptySet();
     private Set<String> allowlistPackageNames = emptySet();
+
     private boolean disallowProxyObjectAccess = false;
     private boolean disallowProxyMemberAccess = false;
     private boolean disallowDefaultPackageAccess = false;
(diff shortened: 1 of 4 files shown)


kusalk:
@lukaszlenart I need to add coverage, will try get it done this weekend

plugins/spring/src/test/java/com/opensymphony/xwork2/ognl/SecurityMemberAccessProxyTest.java
kusalk, line 1: No functional changes in here, I upgraded it to JUnit4 and cleaned it up

core/src/test/java/com/opensymphony/xwork2/ognl/SecurityMemberAccessTest.java
```diff
@@ -931,6 +981,15 @@ public void packageInclusion_subclass_both() throws Exception {
     private static String formGetterName(String propertyName) {
         return "get" + propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1);
     }
+
+    @SuppressWarnings("unchecked")
+    private static <T> T mockHibernateProxy(T originalObject, Class<T> proxyInterface) {
```
kusalk, line 986: Mocking a Hibernate proxy isn't perfect - integration/acceptance tests would offer better protection against regressions but I'd prefer not to complicate the tests further by introducing a Hibernate session factory etc

core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -209,10 +209,23 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
...
-        Class<?> memberClass = member.getDeclaringClass();
         if (!enforceAllowlistEnabled) {
             return true;
         }
+
+        if (!disallowProxyObjectAccess && target != null && ProxyUtil.isProxy(target)) {
+            // If `disallowProxyObjectAccess` is not set, allow resolving Hibernate entities to their underlying
+            // classes/members. This allows the allowlist capability to continue working and offer some level of
+            // protection in applications where the developer has accepted the risk of allowing OGNL access to Hibernate
+            // entities. This is preferred to having to disable the allowlist capability entirely.
```
lukaszlenart, line 220: Wouldn't be good to log this info? Maybe even in WARN level if `struts.devMode` is enabled, wdyt?

kusalk: Hmm yeah doesn't hurt to add some logging - will do

kusalk: Done

lukaszlenart (commented):
_approved_ ;)

PR #968, opened 2024-06-18
WW-5310 Fixes broken support for Fragments in <s:url/> tag

lukaszlenart:
Closes [WW-5310](https://issues.apache.org/jira/browse/WW-5310)


rgielen (approved):
I've crosschecked https://datatracker.ietf.org/doc/html/rfc3986#section-2.1 to understand whether it's our business to take care of unencoded '=' in value of k/v pair, but it looks like it SHOULD be encoded, but is not required. Splitting at first '=' only seems to be legit and preferred behaviour.

That said, the fix correctly addresses both the improved handling for unencoded '='  as well as proper dealing with the fragment part 👍

kusalk (approved):

PR #970, opened 2024-06-21
WW-5430 Uses Freemarker native support for JakartaEE instead of manually transforming the artifacts

lukaszlenart:
Closes [WW-5430](https://issues.apache.org/jira/browse/WW-5430)


lukaszlenart:
cc: @rubensgomes (hope the right one :D )

kusalk (approved):

rubensgomes:
Approved.  The root cause of the issue was related to the use of old (legacy) freemarker library based on javax.  I verified that the freemarker dependency libraries were upgraded, and proper code changes were done.

rubensgomes (approved):
Changes look good.

PR #979, opened 2024-07-08
WW-5439 Move Dev Mode security configuration

kusalk:
WW-5439
--
I noticed that after a previous refactor of `SecurityMemberAccess` - the devMode security configuration stopped working. 

To be fair, this capability is probably not too useful anymore from 7.0, given the allowlist is enabled by default and that has no devMode allowances.

Best we still fix it for now and can consider a better mechanism later.


core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java
```diff
@@ -460,4 +473,41 @@ public void useDisallowProxyMemberAccess(String disallowProxyMemberAccess) {
...
+        this.devModeExcludedPackageNames = toNewPackageNamesSet(devModeExcludedPackageNames, commaDelimitedPackageNames);
+    }
+
+    @Inject(value = StrutsConstants.STRUTS_DEV_MODE_EXCLUDED_PACKAGE_EXEMPT_CLASSES, required = false)
+    public void useDevModeExcludedPackageExemptClasses(String commaDelimitedClasses) {
+        this.devModeExcludedPackageExemptClasses = toClassesSet(commaDelimitedClasses);
+    }
+
+    private void useDevModeConfiguration() {
+        if (!isDevMode || isDevModeInit) {
```
kusalk, line 503: The `isDevModeInit` check isn't thread-safe but it doesn't need to be as there's no negative consequence of running this method more than once when DevMode is enabled.

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlUtilTest.java
```diff
@@ -1284,12 +1295,11 @@ public void testAvoidCallingMethodsWithBraces() {
         assertEquals(expected.getMessage(), "Inappropriate OGNL expression: toString()");
     }
 
-    public void testAvoidCallingSomeClasses() {
+    public void testStaticMethodBlocked() {
         Foo foo = new Foo();
 
         Exception expected = null;
         try {
-            ognlUtil.setExcludedClasses(Runtime.class.getName());
```
kusalk, line 1292: The exclusion list isn't checked here as it's already blocked by the static method check

core/src/test/java/com/opensymphony/xwork2/ognl/OgnlUtilTest.java
```diff
@@ -1166,9 +1171,11 @@ public void testAvoidCallingMethodsOnObjectClass() {
     public void testAllowCallingMethodsOnObjectClassInDevModeTrue() {
         Exception expected = null;
         try {
-            ognlUtil.setExcludedClasses(Foo.class.getName());
```
kusalk, line 1169: These methods don't do anything, so we inject the configuration instead

lukaszlenart (approved):

PR #986, opened 2024-07-13
WW-5440 Fix OGNL allowlist compat with Convention plugin

kusalk:
WW-5440
--
* Fixed `struts2-convention-plugin` compatibility with OGNL allowlist, Actions are now auto-allowlisted as expected
* Fixed `struts2-config-browser-plugin` compatibility with `struts.parameters.requireAnnotations=true`
* Fixed Showcase App Hangman Action by adding appropriate classes to OGNL allowlist configuration
* Fixed a number of other Showcase App Actions which required further annotating/allowlisting
* Deprecated `AnnotationParameterFilterInterceptor` which is superseded by `@StrutsParameter` capability

I also went through and did a batch application of `@StrutsParameter` even in unit test Actions in which they may not be strictly required. Did this as a precautionary measure against any other regressions or unintended test behaviour.

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/Allowed.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/Allowed.java
@@ -28,9 +28,11 @@ import java.lang.annotation.Target;
  * a HttpRequest parameter.
  *
  * @author martin.gilday
+ * @deprecated since 6.6.0, use {@link org.apache.struts2.interceptor.parameter.StrutsParameter}.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
+@Deprecated
 public @interface Allowed {
 
 }
--- a/core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/AnnotationParameterFilterInterceptor.java
+++ b/core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/AnnotationParameterFilterInterceptor.java
@@ -23,13 +23,13 @@ import com.opensymphony.xwork2.ActionInvocation;
 import com.opensymphony.xwork2.interceptor.AbstractInterceptor;
 import com.opensymphony.xwork2.interceptor.Interceptor;
 import com.opensymphony.xwork2.interceptor.ParameterFilterInterceptor;
-import com.opensymphony.xwork2.interceptor.ParametersInterceptor;
 import com.opensymphony.xwork2.util.AnnotationUtils;
 import org.apache.struts2.dispatcher.HttpParameters;
+import org.apache.struts2.interceptor.parameter.ParametersInterceptor;
+import org.apache.struts2.interceptor.parameter.StrutsParameter;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 
 /**
@@ -46,7 +46,10 @@ import java.util.List;
  * </p>
  *
  * @author martin.gilday
+ * @deprecated since 6.6.0, integrated into {@link ParametersInterceptor} with {@link StrutsParameter} using
+ * {@code struts.parameters.requireAnnotations=true}
  */
+@Deprecated
 public class AnnotationParameterFilterInterceptor extends AbstractInterceptor {
 
     /* (non-Javadoc)
--- a/core/src/test/java/org/apache/struts2/ExecutionCountTestAction.java
+++ b/core/src/test/java/org/apache/struts2/ExecutionCountTestAction.java
@@ -41,9 +41,10 @@ public class ExecutionCountTestAction extends ActionSupport {
         return executionCount;
     }
 
+    @Override
     public String execute() throws Exception {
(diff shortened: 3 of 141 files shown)


kusalk:
@lukaszlenart Sorry about the massive diff on this one - it should be a bit easier to review commit by commit. Otherwise, let me know and I'll try split it up into smaller PRs

core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/Allowed.java
```diff
@@ -28,9 +28,11 @@
  * a HttpRequest parameter.
  *
  * @author martin.gilday
+ * @deprecated since 6.6.0, use {@link org.apache.struts2.interceptor.parameter.StrutsParameter}.
```
lukaszlenart, line 31: ❤️

core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/AnnotationParameterFilterInterceptor.java
```diff
@@ -46,7 +46,10 @@
  * </p>
  *
  * @author martin.gilday
+ * @deprecated since 6.6.0, integrated into {@link ParametersInterceptor} with {@link StrutsParameter} using
```
lukaszlenart, line 49: Do we have a task to remove this interceptor in Struts 7?

core/src/test/java/org/apache/struts2/ExecutionCountTestAction.java
```diff
@@ -41,9 +41,10 @@ public int getExecutionCount() {
         return executionCount;
     }
 
+    @Override
     public String execute() throws Exception {
         executionCount++;
-        LOG.info("executing ExecutionCountTestAction. Current count is " + executionCount);
+        LOG.info("executing ExecutionCountTestAction. Current count is {}", executionCount);
```
lukaszlenart, line 47: ❤️

plugins/config-browser/src/main/java/org/apache/struts2/config_browser/ShowConfigAction.java
```diff
@@ -98,16 +104,17 @@ public PropertyDescriptor[] getProperties() {
...
         config = configHelper.getActionConfig(namespace, actionName);
-        actionNames = new TreeSet<String>(configHelper.getActionNames(namespace));
+        actionNames = new TreeSet<>(configHelper.getActionNames(namespace));
         try {
             Object action = objectFactory.buildAction(actionName, namespace, config, null);
             properties = reflectionProvider.getPropertyDescriptors(action);
         } catch (Exception e) {
-            LOG.error("Unable to get properties for action " + actionName, e);
-            addActionError("Unable to retrieve action properties: " + e.toString());
+            LOG.error("Unable to get properties for action {}", actionName, e);
```
lukaszlenart, line 116: This won't work, you must use `ParameterizedMessage` like this

```java
LOG.error(new ParameterizedMessage("Unable to get properties for action {}", actionName), e);
```

lukaszlenart:
Thanks a lot for all your work, this is huge! One small thing and I'm good

core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/AnnotationParameterFilterInterceptor.java
kusalk: I've just updated the description of [WW-5411](https://issues.apache.org/jira/browse/WW-5411) - if you'd prefer a separate card I can create one too

plugins/config-browser/src/main/java/org/apache/struts2/config_browser/ShowConfigAction.java
kusalk: So I double checked this because IntelliJ insisted this was correct despite the JavaDoc for the method suggesting it was incorrect. Turns out Log4J 2 will actually log this correctly as it will identify that the last argument is an exception (also tested locally). Here is the relevant Log4J 2 code - https://github.com/apache/logging-log4j2/blob/rel/2.23.1/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java#L128

core/src/main/java/com/opensymphony/xwork2/interceptor/annotations/AnnotationParameterFilterInterceptor.java
lukaszlenart: That's fine, just to not forget about this :)

plugins/config-browser/src/main/java/org/apache/struts2/config_browser/ShowConfigAction.java
lukaszlenart: Ah, nice! I were always using `ParameterizedMessage` in such case, good to know!

lukaszlenart:
_conflict_

kusalk:
Resolved!

lukaszlenart (approved):

PR #1005, opened 2024-07-26
WW-5449 Make Velocity Tools dependency optional for Velocity plugin

kusalk:
WW-5449
--
I suspect many applications that use the Velocity plugin don't need the Velocity Tools dependency. By extracting the Velocity Tools related functionality into a separate class `VelocityTools.java`, we give applications the ability to opt out. This class will only be initialised if `struts.velocity.toolboxlocation` is set.

Velocity Tools is also still needed if using certain SiteMesh or Tiles plugin functionality. This is the case for the Showcase app so I've added Velocity Tools as an explicit dependency in that pom.

I've also refactored `VelocityManager` to be more readable.

### DIFF
--- a/apps/showcase/pom.xml
+++ b/apps/showcase/pom.xml
@@ -87,6 +87,14 @@
             <groupId>org.apache.struts</groupId>
             <artifactId>struts2-velocity-plugin</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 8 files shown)


lukaszlenart (approved):
LGTM 👏

PR #1010, opened 2024-08-03
WW-5454 Log warnings on startup for disabled critical security features

kusalk:
WW-5454
--
This PR adds startup log warnings for disabling the following capabilities:
* OGNL class allowlist
* @StrutsParameter annotation requirement


core/src/main/java/com/opensymphony/xwork2/util/DebugUtils.java
```diff
@@ -37,4 +42,12 @@ public static void notifyDeveloperOfError(Logger log, Object action, String mess
         }
     }
 
+    /**
+     * @since 7.0
+     */
+    public static void logWarningForFirstOccurrence(String key, Logger log, String msg, Object... args) {
```
kusalk, line 48: This method isn't really a `DebugUtils` method but couldn't think of a better home - can create a separate class if you prefer

lukaszlenart (approved):

PR #1068, opened 2024-10-06
Potential mitigation for WW-5466

JCgH4164838Gh792C124B5:
Hello Apache Struts Team.

This is a proposed mitigation for WW-5466.  It is unlikely that many users will encounter it given the default configuration normally sets `struts.multipart.maxStringLength` .

The proposed change is a basic guard, combined with some additional unit tests to confirm that processing works both when the configuration values have not been set, as well as when they have been set.

Please advise if anyone thinks the change should be considered, or needs additional consideration.

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
@@ -142,7 +142,7 @@ public class JakartaMultiPartRequest extends AbstractMultiPartRequest {
             }
 
             long size = item.getSize();
-            if (size > maxStringLength) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
```diff
@@ -142,7 +142,7 @@ protected void processNormalFormField(FileItem item, String charset) throws Unsu
             }
 
             long size = item.getSize();
-            if (size > maxStringLength) {
+            if (maxStringLength != null && size > maxStringLength) {
```
lukaszlenart, line 145: Good catch!

lukaszlenart (approved):

PR #1071, opened 2024-10-11
WW-5468 Fixes @StrutsParameter for ModelDriven

lukaszlenart:
Fixes [WW-5468](https://issues.apache.org/jira/browse/WW-5468)

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/ModelDriven.java
+++ b/core/src/main/java/com/opensymphony/xwork2/ModelDriven.java
@@ -18,6 +18,8 @@
  */
 package com.opensymphony.xwork2;
 
+import org.apache.struts2.interceptor.parameter.StrutsParameter;
+
 /**
  * ModelDriven Actions provide a model object to be pushed onto the ValueStack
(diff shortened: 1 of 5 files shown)


kusalk:
It seems that the `java.beans.PropertyDescriptor` API will always return the getter method on the `ModelDriven` interface rather than the Action (I didn't actually realise this when I wrote this feature). This means it doesn't matter if `getModel` on the actual Action class is annotated or not, it will only ever check the interface method.

To be honest, I don't mind this behaviour - what if we just exempt all model driven Actions from the annotation requirement? The `@StrutsParameter` requirement was designed to protect against random getters/setters on the Action class from being injected. But if an Action is using a dedicated model object (and the Action class is indeed not on the root of the ValueStack) it is already much safer.

Here is my alternative solution for your consideration: f67f593b

lukaszlenart:
I'm fine either way, my PR is just a starting point for discussion. Feel free to open PR and we can discuss further.

kusalk:
I've created #1072 which exempts `ModelDriven` Actions from the annotation requirement entirely - I've also retained your other changes, including the additional debug logging, with some minor adjustments

PR #1072, opened 2024-10-12
WW-5468 Exempt ModelDriven Actions from @StrutsParameter requirement

kusalk:
WW-5468
--
The `@StrutsParameter` requirement was designed to protect against arbitrary getters and setters on the Action class from being invoked by users and/or attackers. However, if an Action is using a dedicated model object alongside the `ModelDrivenInterceptor`  (which ensures the Action is not on the root of the value stack) much of this risk is mitigated. I suggest we exempt this specific scenario from requiring the `@StrutsParameter` annotation.

Closes #1071

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
@@ -20,6 +20,7 @@ package org.apache.struts2.interceptor.parameter;
 
 import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.ActionInvocation;
+import com.opensymphony.xwork2.ModelDriven;
 import com.opensymphony.xwork2.inject.Inject;
 import com.opensymphony.xwork2.interceptor.MethodFilterInterceptor;
 import com.opensymphony.xwork2.security.AcceptedPatternsChecker;
@@ -348,7 +349,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
         }
 
         long paramDepth = name.codePoints().mapToObj(c -> (char) c).filter(NESTING_CHARS::contains).count();
-        if (requireAnnotationsTransitionMode && paramDepth == 0) {
+        if (!isModelDriven(action) && requireAnnotationsTransitionMode && paramDepth == 0) {
             return true;
         }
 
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -348,7 +349,15 @@ protected boolean isParameterAnnotatedAndAllowlist(String name, Object action) {
         }
 
         long paramDepth = name.codePoints().mapToObj(c -> (char) c).filter(NESTING_CHARS::contains).count();
+
+        if (action instanceof ModelDriven<?> && !ActionContext.getContext().getValueStack().peek().equals(action)) {
+            LOG.debug("Model driven Action detected, exempting from @StrutsParameter annotation requirement and OGNL allowlisting model type");
+            // (Exempted by annotation on com.opensymphony.xwork2.ModelDriven#getModel)
+            return hasValidAnnotatedMember("model", action, paramDepth + 1);
```
kusalk, line 356: This conditional block and the added annotation on the `ModelDriven` interface comprise the core fix

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -400,8 +411,10 @@ protected boolean hasValidAnnotatedPropertyDescriptor(Object action, PropertyDes
             }
             return false;
         }
+        LOG.debug("Success: Matching annotated method [{}] found for property [{}] of depth [{}] on Action [{}]",
+                relevantMethod.getName(), propDesc.getName(), paramDepth, action.getClass().getSimpleName());
         if (paramDepth >= 1) {
-            allowlistClass(relevantMethod.getReturnType());
+            allowlistClass(propDesc.getPropertyType());
```
kusalk, line 417: Minor fix here to ensure the specific model type is allowlisted rather than the type declared in the interface (`Object.class`)

core/src/test/java/com/opensymphony/xwork2/ModelDrivenAction.java
```diff
@@ -44,7 +44,6 @@ public String getFoo() {
     /**
      * @return the model to be pushed onto the ValueStack after the Action itself
      */
-    @StrutsParameter(depth = 2)
```
kusalk, line 47: I removed all these annotations on `ModelDriven` Action model getters as they are no longer required (not that they were being detected correctly previously)

plugins/rest/src/test/java/org/apache/struts2/rest/RestActionInvocationTest.java
kusalk, line 1: Replaced tabs with spaces to fix the janky formatting

core/src/test/java/org/apache/struts2/interceptor/parameter/StrutsParameterAnnotationTest.java
```diff
@@ -343,6 +360,14 @@ public Map<String, Pojo> getPublicPojoMapDepthTwo() {
         }
     }
 
-    class Pojo {
+    static class ModelAction implements ModelDriven<Pojo> {
```
lukaszlenart, line 363: There is one more case where `ModelDriven<Object>` is used and then `getModel()` can return conditional model, either a `List<Pojo>` or `Pojo` itself. See [that example](https://github.com/apache/struts-examples/blob/master/rest-angular/src/main/java/org/apache/examples/struts/actions/data/OrderController.java#L71-L80)

lukaszlenart (approved):
LGTM 👍

core/src/test/java/org/apache/struts2/interceptor/parameter/StrutsParameterAnnotationTest.java
kusalk: Ah interesting, let me test this

kusalk: I think in such cases, the app developer should manually OGNL allowlist any required types, I'm not confident of a secure way to auto allowlist in this case

lukaszlenart: Yeah, it can be tricky, but having such a note should be enough 👍

PR #1077, opened 2024-10-16
WW-5472 Removes Struts Sitemesh plugin

lukaszlenart:
WW-5472 
--

PR removes Struts Sitemesh plugin as the plugin uses Sitemesh 2 which isn't supported anymore. It also replaces Struts Sitemesh plugin integration in the Showcase app with pure Sitemesh 3 integration.

~~Depends on [WW-5473](https://issues.apache.org/jira/browse/WW-5473)~~
~~Depends on #1078~~


apps/showcase/src/main/webapp/WEB-INF/decorators/main.jsp
```diff
@@ -61,7 +59,7 @@
     <meta name="description" content="Struts2 Showcase for Apache Struts Project">
     <meta name="author" content="The Apache Software Foundation">
 
-    <title><decorator:title default="Struts2 Showcase"/></title>
+    <title><sitemesh:write property="title"/></title>
```
sepe81, line 62: @lukaszlenart Don't we need a taglib replacement on L:52?

IntelliJ warns `Namespace 'sitemesh' is not bound`.

lukaszlenart: Should be ok based on the doc
https://github.com/sitemesh/sitemesh3?tab=readme-ov-file#creating-a-decorator

sepe81: Thank you for the clarification!

PR #1080, opened 2024-10-17
WW-3714 Deprecate and migrate factory extension points

kusalk:
WW-3714
--
Testing


kusalk:
Hmm this could break compatibility with `bean` elements - abandoning this PR

PR #1086, opened 2024-10-17
WW-3714 Deprecate and repackage common APIs part 7

kusalk:
WW-3714
--
This PR deprecates and provides repackaged replacements for the following APIs, ahead of their permanent migration in Struts 7.0:
- ``


kusalk:
Hmm these specific changes are getting a bit too complex/risky for my liking

PR #1079, opened 2024-10-17
WW-3714 Deprecate and repackage common APIs part 1

kusalk:
WW-3714
--
This PR deprecates and provides repackaged replacements for the following APIs, ahead of their permanent migration in Struts 7.0:
- `com.opensymphony.xwork2.ActionContext`
- `com.opensymphony.xwork2.Action`
- `com.opensymphony.xwork2.Result`
- `com.opensymphony.xwork2.interceptor.Interceptor`
- `com.opensymphony.xwork2.interceptor.ConditionalInterceptor`

Note that these changes alone are not sufficient for defining Interceptors/Results implementing the new interfaces in XML configuration files. Factory support is also necessary which is introduced in #1087.

PR #1087, opened 2024-10-17
WW-3714 Deprecate and repackage common APIs part 2.5

kusalk:
WW-3714
--
Ensure factory support for new Interceptor, Result interfaces introduced in #1079.

When Results and Interfaces are parsed and instantiated from XML configuration they are expected to be of the deprecated `com.opensymphony.xwork2` types. We want to also ensure the new `org.apache.struts2` types are supported. We cannot easily change the factory interface signatures without breaking compatibility for the associated extension points. Thus, we ensure these types are adapted within the default factory implementations instead. Additional helper code was needed to ensure that the marker interfaces `ReflectionExceptionHandler`, `WithLazyParams`, `ParamNameAwareResult` are still respected. This helper code can of course be completely discarded in Struts 7.0.


kusalk:
@lukaszlenart Sorry to bug you but I need this one reviewed and merged first (as the other PRs are stacked on top) 🙂

lukaszlenart (approved):

PR #1105, opened 2024-11-01
WW-5478 Deprecate DefaultResultFactory

kusalk:
WW-5478
--
I can't see any value in maintaining this class as it does the same thing as `StrutsResultFactory` minus support for the marker `ParamNameAwareResult`.


lukaszlenart (approved):
Updated ticket reference, beside that looks good LGTM 👍

kusalk:
Oops, thank you

PR #1109, opened 2024-11-02
WW-3714 Move new Result class into result package

kusalk:
WW-3714
--
Resolves unusual issue compiling JSPs.


lukaszlenart:
I wouldn't do this in the master branch as this is a breaking change. Also I don't have problems with building the master branch, only with release/struts-7-x-x ;-)

lukaszlenart (approved):
Now I noticed that this is a new Result 🤦

PR #1124, opened 2024-11-10
WW-5455 Defines a new plugin to support Jasper Reports 7

lukaszlenart:
WW-5455
--

Jasper Reports 7 break backward compatibility  with previous Jasper Reports, even JRXML file definition has changed.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -174,6 +174,11 @@
 
         <dependency>
             <groupId>org.apache.struts</groupId>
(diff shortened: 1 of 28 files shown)


kusalk (approved):
Not familiar with Jasper Reports to be able to review this in detail

lukaszlenart:
> Not familiar with Jasper Reports to be able to review this in detail

Me too, and that's why I marked it as experimental tbh. Thanks!

PR #1141, opened 2024-12-04
WW-5493 Restores parameters and uses getAttributes()

lukaszlenart:
Instead of using parameters field directly, this PR uses getAttributes() instead to avoid coupling with internal Component state

Fixes [WW-5493](https://issues.apache.org/jira/browse/WW-5493)


sepe81 (approved):
Seems reasonable, plus a few cleanups.

kusalk (approved):

PR #1156, opened 2024-12-23
WW-5501 Exclude malicious names

lukaszlenart:
Hardening file upload logic

### DIFF
--- a/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
+++ b/core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
@@ -180,8 +180,4 @@ public abstract class AbstractMultiPartRequest implements MultiPartRequest {
         return patternsChecker.isAllowed(fileName).isAllowed();
     }
 
-    protected String sanitizeNewlines(String before) {
-        return before.replaceAll("[\n\r]", "_");
-    }
-
 }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -169,4 +176,12 @@ protected String getCanonicalName(final String originalFileName) {
         return fileName;
     }
 
+    protected boolean isAccepted(String fileName) {
+        return patternsChecker.isAllowed(fileName).isAllowed();
+    }
+
+    protected String sanitizeNewlines(String before) {
```
kusalk, line 183: Nit: So I introduced this function in #760 but I realise now we can just use `org.apache.commons.lang3.StringUtils#normalizeSpace` instead

kusalk (approved):

core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
kusalk: Although given this is a protected method, deleting it would be a breaking change so only thing we could really do is replace the implementation with `return StringUtils.normalizeSpace(before);`

lukaszlenart: We can always mark this function as deprecated and use `StringUtils.normalizeSpace()` directly

kusalk: Yep that works too

kusalk (approved):

PR #1159, opened 2024-12-23
WW-5503 Removes unused dependencies

lukaszlenart:
WW-5503


kusalk:
Looks like the `byte-buddy` dependency management may still be required

lukaszlenart:
Yeah, looks like. Trying to use `mvn dependency:analyze` to figure out what can be removed

kusalk (approved):

PR #1174, opened 2025-01-05
WW-5504 Allows to use request instead of session attribute to store nonce

lukaszlenart:
WW-5504


assachs:
I'm not sure if i missed something.

In DefaultCSPSettings, when creating the poplicy format, the nonce is always taken from the session


`       return format(policyFormatBuilder.toString(), **getNonceString(request));**
    }

    protected String getNonceString(HttpServletRequest request) {
        Object nonce = **request.getSession().getAttribute(NONCE_KEY);**
        return Objects.toString(nonce);
    }`

lukaszlenart:
Thanks, fixed! Yet I'm not sure if this is a proper way to do it

PR #1175, opened 2025-01-05
WW-5512 Extends the container to support injecting optional parameters into constructor

lukaszlenart:
WW-5512

### DIFF
--- a/core/src/main/java/com/opensymphony/xwork2/inject/ContainerImpl.java
+++ b/core/src/main/java/com/opensymphony/xwork2/inject/ContainerImpl.java
@@ -419,8 +419,12 @@ class ContainerImpl implements Container {
                 // First time through...
                 constructionContext.startConstruction();
                 try {
-                    final Object[] parameters = getParameters(constructor, context, parameterInjectors);
-                    t = constructor.newInstance(parameters);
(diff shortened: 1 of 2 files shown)


kusalk:
Does this also work for constructors with multiple optional parameters where 1 is available and the other is not?

lukaszlenart:
This is strictly for missing all the parameters and no default constructor exists.

kusalk:
How about the following - both parameters are injected as `null` even though one exists. Is that expected? I guess my concern here is that unintuitive behaviour here could lead to bugs being introduced in the future
```
@Inject(required = false)
public ConstructorCheck(@Inject(value = "constructorCheck.name") String name, @Inject(value = "nonExistingConstant") String name2) {
```

lukaszlenart:
I can add a test case for that, but that should block such a situation

 ```(constructor.getParameterCount() > 0 && parameterInjectors == null)```

constructor has parameters defined but there is nothing to inject.

kusalk:
Yep if you can add a test that verifies the behaviour for my given example hasn't changed (injection exception is thrown) that would make me more comfortable

lukaszlenart:
Done, let me know what do you think

kusalk (approved):
Nice, thank you!

PR #1186, opened 2025-01-17
Adds 6.7.x branch to nigthlies step

lukaszlenart:


sepe81 (changes requested):
It should work without the extra semicolon after "main"

sepe81 (approved):

PR #1190, opened 2025-01-20
WW-5516 Fix TemplateDir and Theme fallback to Request/Session/Application attributes

kusalk:
WW-5516
--
This is a fix for a functionality regression introduced in #763. Instead of re-introducing the dependency on an OGNL context expression, we instead access the required attributes directly through Java instead. This fixes the regression whilst still allowing applications to block OGNL context access (expressions beginning with `#`) using `OgnlGuard` if they desire.

### DIFF
--- a/core/src/test/java/org/apache/struts2/components/UIBeanTagTest.java
+++ b/core/src/test/java/org/apache/struts2/components/UIBeanTagTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/struts2/components/UIBeanTagTest.java
```diff
@@ -0,0 +1,92 @@
...
+            public String getTestDir() {
+                return "testValue";
+            }
+        });
+
+        assertEquals("testValue", bean.getTemplateDir());
+    }
+
+    public void testTemplateDir_attrMapFallback() {
+        bean.setTemplateDir("%{testDir}");
```
lukaszlenart, line 57: Doesn't this test depend on the `testTemplateDir_ognlExpression`?

lukaszlenart (commented):
Looks good, I would just add one more test without using OGNL expression, just set value in a given scope (application or session) and check if it was resolved.

core/src/test/java/org/apache/struts2/components/UIBeanTagTest.java
kusalk: Sorry that was a mistake, fixed!

lukaszlenart (approved):

PR #1200, opened 2025-01-31
WW-5525 Prevent NPE when using new in OGNL

FBibonne:


FBibonne:
Failing tests about the issue pass after the fix. It only stays DynAttribsTest which does not pass (`DynAttribsTest.setUp:69 unable to find struts-tags.tld` error) but it is yet the case for main branch

kusalk:
Hi @FBibonne 

Thanks for reporting this and contributing a PR, including proper test coverage. However, I did have a simpler fix in mind and have copied across your test case - feel free to review here: #1214

cc @lukaszlenart

FBibonne:
Thank you @kusalk for fixing this NPE

Your PR 1214 looks very fine to my point of view

PR #1201, opened 2025-02-01
WW-5525 Fixes NPE when checking if expressions is acceptable

lukaszlenart:
WW-5525

### DIFF
--- a/core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
@@ -160,12 +160,12 @@ public class SecurityMemberAccess implements MemberAccess {
             }
         }
 
-        if (target != null && !checkProxyObjectAccess(target)) {
+        if (!checkProxyObjectAccess(target)) {
             LOG.warn("Access to proxy is blocked! Target [{}], proxy class [{}]", target, target.getClass().getName());
             return false;
         }
 
-        if (target != null && !checkProxyMemberAccess(target, member)) {
+        if (!checkProxyMemberAccess(target, member)) {
             LOG.warn("Access to proxy is blocked! Member class [{}] of target [{}], member [{}]", member.getDeclaringClass(), target, member);
             return false;
         }
@@ -185,15 +185,15 @@ public class SecurityMemberAccess implements MemberAccess {
             return false;
         }
 
-        if (target != null && !checkDefaultPackageAccess(target, member)) {
+        if (!checkDefaultPackageAccess(target, member)) {
             return false;
         }
 
-        if (target != null && !checkExclusionList(target, member)) {
+        if (!checkExclusionList(target, member)) {
             return false;
         }
 
-        if (target != null && !checkAllowlist(target, member)) {
+        if (!checkAllowlist(target, member)) {
             return false;
         }
 
--- a/core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
+++ b/core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
@@ -160,12 +160,12 @@ public class SecurityMemberAccess implements MemberAccess {
             }
         }
 
-        if (!checkProxyObjectAccess(target)) {
+        if (target != null && !checkProxyObjectAccess(target)) {
             LOG.warn("Access to proxy is blocked! Target [{}], proxy class [{}]", target, target.getClass().getName());
             return false;
         }
 
-        if (!checkProxyMemberAccess(target, member)) {
+        if (target != null && !checkProxyMemberAccess(target, member)) {
             LOG.warn("Access to proxy is blocked! Member class [{}] of target [{}], member [{}]", member.getDeclaringClass(), target, member);
             return false;
         }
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/struts2/ognl/SecurityMemberAccessTest.java
```diff
@@ -649,6 +649,7 @@ public void testBlockedStaticFieldWhenFlagIsFalse() throws Exception {
     @Test
     public void testBlockedStaticFieldWhenClassIsExcluded() throws Exception {
         // given
+        assignNewSma(false);
```
kusalk, line 652: By adding this, the test doesn't test the behaviour it's meant to check - the point of this test is to ensure the exclusion list is enforced for static fields even when static fields aren't blocked as a whole

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -157,44 +157,44 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
             }
         }
 
-        if (!checkProxyObjectAccess(target)) {
+        if (target != null && !checkProxyObjectAccess(target)) {
```
kusalk, line 163: This is redundant after https://github.com/apache/struts/pull/1214 - if we wanted to keep it as an additional check it might be more consistent to move it inside `#checkProxyObjectAccess`

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -160,12 +160,12 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
...
             LOG.warn("Access to proxy is blocked! Target [{}], proxy class [{}]", target, target.getClass().getName());
             return false;
         }
 
-        if (!checkProxyMemberAccess(target, member)) {
+        if (target != null && !checkProxyMemberAccess(target, member)) {
```
kusalk, line 168: This is redundant after https://github.com/apache/struts/pull/1214 but could also present a bug if we encountered a proxied static member or constructor

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -185,15 +185,15 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
             return false;
         }
 
-        if (!checkDefaultPackageAccess(target, member)) {
+        if (target != null && !checkDefaultPackageAccess(target, member)) {
```
kusalk, line 188: Bug - prevents blocking of default package for static members and constructors

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -185,15 +185,15 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
...
             return false;
         }
 
-        if (!checkExclusionList(target, member)) {
+        if (target != null && !checkExclusionList(target, member)) {
```
kusalk, line 192: Bug - prevents enforcement of exclusion list for static members and constructors

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -185,15 +185,15 @@ public boolean isAccessible(Map context, Object target, Member member, String pr
...
             return false;
         }
 
-        if (!checkAllowlist(target, member)) {
+        if (target != null && !checkAllowlist(target, member)) {
```
kusalk, line 196: Bug - prevents enforcement of allowlist for static members and constructors

kusalk (changes requested):
While I understand the intent behind this PR - to eliminate even the possibility of an NPE - I don't believe we need it anymore

We expect `target` will be null for static members and constructors. However, the `target` being null doesn't necessarily mean that check is inapplicable. Making that assumption could lead to bugs (pointed out in the comments). It should be up to the consuming method (e.g. `#checkDefaultPackageAccess`) to determine how to handle the nullability of `target`. If we wanted to force the consuming method to handle the nullability (and prevent NPEs) a better solution might be to wrap it in an `Optional`.

The new unit tests look good though, so perhaps we just keep those?

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
lukaszlenart: Thanks, removed!

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
lukaszlenart: Thanks, removed!

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
lukaszlenart: Got it, removed!

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
lukaszlenart: Got it, removed!

core/src/test/java/org/apache/struts2/ognl/SecurityMemberAccessTest.java
lukaszlenart: Yeah, now it's clear why this test started to fail after my changes, thanks a lot for pointing this out!

lukaszlenart:
Done, just tests left :)

kusalk (approved):

PR #1203, opened 2025-02-02
WW-5501 Ignores DMI related action field

lukaszlenart:
WW-5501


core/src/main/java/org/apache/struts2/dispatcher/multipart/JakartaMultiPartRequest.java
```diff
@@ -41,6 +43,16 @@ public class JakartaMultiPartRequest extends AbstractMultiPartRequest {
 
     private static final Logger LOG = LogManager.getLogger(JakartaMultiPartRequest.class);
 
+    public JakartaMultiPartRequest() {
+        super(Boolean.FALSE.toString());
+    }
+
+    public JakartaMultiPartRequest(
```
kusalk, line 50: Do we need an `@Inject` on the constructor too - does the container definitely use this constructor?

lukaszlenart: Hm... I've got an error without the default constructor, but you are right and I will double check

lukaszlenart: @gregh3269 tested this and even proposed a fix :)

PR #1218, opened 2025-02-16
WW-5501 Reverts all changes related to WW-5501

lukaszlenart:
WW-5501
After longer discussion on the Dev User List I decided to revert changes related to hardening file upload logic.


core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -180,43 +157,16 @@ public List<LocalizedMessage> getErrors() {
      * @return the canonical name based on the supplied filename
      */
     protected String getCanonicalName(final String originalFileName) {
-        return FilenameUtils.getName(originalFileName);
```
kusalk, line 183: Could keep this but not important

kusalk (approved):

core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
lukaszlenart: Good catch, restored

PR #1225, opened 2025-02-20
WW-5531 Enforce JRE8 compatibility (when compiled on JDK9+)

kusalk:
WW-5531
--

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,15 @@
             </build>
         </profile>
         <profile>
+            <id>jdk9plus</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
+            </properties>
+        </profile>
+        <profile>
             <id>jdk17</id>
             <activation>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -183,6 +183,15 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>jdk9plus</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
```
lukaszlenart, line 192: Shouldn't be 1.8? And [this options](https://github.com/apache/struts/blob/fe5b1977383b05f08eaa049385b0f67eb19b8f8a/pom.xml#L108-L109) are not enough?

kusalk: `8` and `1.8` are equivalent

The options mean slightly different things, the `release` option is required to ensure bytecode compatibility
[Here](https://stackoverflow.com/a/65655542/14093292) is a good overview from StackOverflow

lukaszlenart (approved):

PR #1232, opened 2025-02-24
WW-5532 Upgrade and align various dependencies

kusalk:
WW-5532
--

Dependencies:
- Spring: `6.0.13` -> `6.2.3`
- Hibernate-Validator: `8.0.1.Final` -> `8.0.2.Final`

Adopt BOMs:
- `jakarta.platform:jakarta.jakartaee-bom`
- `org.glassfish.jaxb:jaxb-bom`
- `org.springframework:spring-framework-bom`
- `org.apache.logging.log4j:log4j-bom`
- `com.fasterxml.jackson:jackson-bom`

Test dependencies:
- ByteBuddy: `1.16.1` -> `1.17.1`
- Mockito: `5.8.0` -> `5.15.2`
- AssertJ: `3.26.3` -> `3.27.3`
- TestNG: `7.5.1` -> `7.11.0`

Remove or minimise usages of deprecated test dependencies:
- `org.springframework:spring-test:4.3.0.RELEASE` -> `com.github.h-thurow:simple-jndi`
- EasyMock
- Mockobjects

Removed usages of Java EE 8 dependencies.

### DIFF
--- a/plugins/rest/pom.xml
+++ b/plugins/rest/pom.xml
@@ -58,6 +58,7 @@
         <dependency>
             <groupId>org.apache.juneau</groupId>
             <artifactId>juneau-marshall</artifactId>
+            <version>8.1.3</version>
             <optional>true</optional>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -106,24 +106,23 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.outputTimestamp>2025-02-17T09:41:25Z</project.build.outputTimestamp>
         <maven.compiler.release>17</maven.compiler.release>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
 
         <!-- dependency versions in alphanumeric order -->
         <asm.version>9.7.1</asm.version>
-        <byte-buddy.version>1.16.1</byte-buddy.version>
+        <byte-buddy.version>1.17.1</byte-buddy.version>
         <freemarker.version>2.3.34</freemarker.version>
-        <hibernate-validator.version>8.0.1.Final</hibernate-validator.version>
+        <hibernate-validator.version>8.0.2.Final</hibernate-validator.version>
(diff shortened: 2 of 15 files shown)


plugins/rest/pom.xml
```diff
@@ -58,6 +58,7 @@
         <dependency>
             <groupId>org.apache.juneau</groupId>
             <artifactId>juneau-marshall</artifactId>
+            <version>8.1.3</version>
```
lukaszlenart, line 61: Version is already defined in the parent pom

pom.xml
```diff
@@ -900,52 +806,12 @@
...
-                <groupId>com.fasterxml.jackson.dataformat</groupId>
-                <artifactId>jackson-dataformat-xml</artifactId>
-                <version>${jackson.version}</version>
-            </dependency>
-
-            <dependency>
-                <groupId>org.apache.juneau</groupId>
-                <artifactId>juneau-marshall</artifactId>
-                <version>8.1.3</version>
-            </dependency>
```
lukaszlenart, line 922: Why did you remove this dependency from <dependencyManagement/> section?

kusalk: I moved version management for dependencies that are specific to only 1 plugin and are unlikely to ever be needed elsewhere as I thought it made the root POM slightly neater. I did the same with the Weld dependencies which are specific to the CDI plugin.

But I don't have a strong opinion here and am happy to revert these specific changes, let me know what your preference is 🙂 

It'd be good to be consistent though, so if we decide to keep them in the root POM, I'd like to version manage the other dependencies that are currently only defined in their plugins (e.g. `org.jfree:jfreechart`, `net.sf.jasperreports:jasperreports` and any others).

lukaszlenart: That's fine yet I would like to adopt the same approach as described here [WW-5376](https://issues.apache.org/jira/browse/WW-5376)

lukaszlenart (approved):

pom.xml
kusalk: I've got a follow-up PR addressing that!

PR #1234, opened 2025-02-25
WW-5376 Fix BOM leaking unrelated dependencies

kusalk:
WW-5376
--
There's a few ways of achieving this, the key criteria being that the `struts2-bom` artifact does not inherit any `dependencyManagement` entries.

The solution I've chosen follows this hierarchy:
`struts2-project` -> `struts2-bom` -> `struts2-parent` -> all other modules

This is in contrast to the previous state:
`struts2-parent` -> `struts2-bom` and all other modules


kusalk:
No idea why CodeQL thinks the project structure is invalid - I can't reproduce the error locally either

lukaszlenart (approved):
Do I get it right that users should import `struts2-bom` to get all the needed dependencies?

kusalk:
> Do I get it right that users should import `struts2-bom` to get all the needed dependencies?

That's correct, users should continue using the existing Struts BOM coordinates, no changes are required on their end. All the unrelated dependencies will no longer be included in the BOM after this change is merged.

kusalk:
Going forth, we should NOT put any dependencies in the root POM (`struts2-project` - `/pom.xml`).
We put only Struts artifacts in the BOM POM (`struts2-bom` - `/bom/pom.xml`).
And any other dependencies should be defined in the "parent" POM (`struts2-parent` - `/parent/pom.xml`).

PR #1236, opened 2025-02-27
WW-5534 Simplify ProxyUtil, add OgnlCache#computeIfAbsent

kusalk:
WW-5534
--
These changes prepare us for introducing Spring proxy support for the OGNL allowlist and `@StrutsParameter` capabilities in a follow-up PR. Refer to PR comments for details on changes.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -187,11 +187,17 @@
             <artifactId>commons-text</artifactId>
         </dependency>
 
-        <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is HibernateProxy -->
         <dependency>
+            <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is HibernateProxy -->
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
             <version>5.6.15.Final</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is Spring proxy -->
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-aop</artifactId>
             <optional>true</optional>
         </dependency>
 
--- a/core/src/main/java/org/apache/struts2/interceptor/ChainingInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ChainingInterceptor.java
@@ -20,11 +20,11 @@ package org.apache.struts2.interceptor;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.struts2.result.ActionChainResult;
 import org.apache.struts2.ActionInvocation;
 import org.apache.struts2.StrutsConstants;
 import org.apache.struts2.Unchainable;
 import org.apache.struts2.inject.Inject;
+import org.apache.struts2.result.ActionChainResult;
 import org.apache.struts2.result.Result;
 import org.apache.struts2.util.CompoundRoot;
 import org.apache.struts2.util.ProxyUtil;
@@ -167,17 +167,18 @@ public class ChainingInterceptor extends AbstractInterceptor {
     }
 
     private void copyStack(ActionInvocation invocation, CompoundRoot root) {
-        List list = prepareList(root);
+        List<Object> list = prepareList(root);
         Map<String, Object> ctxMap = invocation.getInvocationContext().getContextMap();
         for (Object object : list) {
-            if (shouldCopy(object)) {
-                Object action = invocation.getAction();
-                Class<?> editable = null;
-                if(ProxyUtil.isProxy(action)) {
-                    editable = ProxyUtil.ultimateTargetClass(action);
-                }
-                reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
+            if (!shouldCopy(object)) {
+                continue;
             }
+            Object action = invocation.getAction();
+            Class<?> editable = null;
+            if (ProxyUtil.isProxy(action)) {
+                editable = ProxyUtil.ultimateTargetClass(action);
+            }
+            reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
         }
     }
 
@@ -204,9 +205,8 @@ public class ChainingInterceptor extends AbstractInterceptor {
         return o != null && !(o instanceof Unchainable);
     }
 
-    @SuppressWarnings("unchecked")
-    private List prepareList(CompoundRoot root) {
-        List list = new ArrayList(root);
+    private List<Object> prepareList(CompoundRoot root) {
+        var list = new ArrayList<>(root);
         list.remove(0);
         Collections.reverse(list);
         return list;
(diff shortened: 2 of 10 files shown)


kusalk:
`org.apache.struts2.util.ProxyUtil` is covered by `org.apache.struts2.spring.SpringProxyUtilTest` but it seems Sonar doesn't recognise this due to it being in a different module.

core/src/main/java/org/apache/struts2/interceptor/ChainingInterceptor.java
```diff
@@ -167,17 +167,18 @@ public String intercept(ActionInvocation invocation) throws Exception {
...
         Map<String, Object> ctxMap = invocation.getInvocationContext().getContextMap();
         for (Object object : list) {
-            if (shouldCopy(object)) {
-                Object action = invocation.getAction();
-                Class<?> editable = null;
-                if(ProxyUtil.isProxy(action)) {
-                    editable = ProxyUtil.ultimateTargetClass(action);
-                }
-                reflectionProvider.copy(object, action, ctxMap, prepareExcludes(), includes, editable);
+            if (!shouldCopy(object)) {
```
kusalk, line 173: Nothing much to see here, just denested this code for slight readability improvement

core/pom.xml
```diff
@@ -187,13 +187,19 @@
...
-        <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is HibernateProxy -->
         <dependency>
+            <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is HibernateProxy -->
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
             <version>5.6.15.Final</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <!-- Optional used in org.apache.struts2.util.ProxyUtil to detect if object is Spring proxy -->
```
kusalk, line 198: I'm using the same optional dependency technique we used for Hibernate proxy resolution - it allows us to avoid reflecting and duplicating code in `ProxyUtil` leading to improved maintainability

core/src/main/java/org/apache/struts2/ognl/OgnlCache.java
```diff
@@ -15,21 +15,37 @@
...
-public interface OgnlCache<Key, Value> {
+public interface OgnlCache<K, V> {
+
+    V get(K key);
 
-    Value get(Key key);
+    /**
+     * @since 7.1
+     */
+    default V computeIfAbsent(K key,
```
kusalk, line 37: Adding a default implementation preserves API compatibility in case some Struts users have defined their own `OgnlCache` implementations using the extension points. It allows them to upgrade to Struts 7.1 without requiring any additional changes

core/src/main/java/org/apache/struts2/ognl/OgnlUtil.java
```diff
@@ -676,13 +675,19 @@ public BeanInfo getBeanInfo(Object from) throws IntrospectionException {
      * @throws IntrospectionException is thrown if an exception occurs during introspection.
      */
     public BeanInfo getBeanInfo(Class<?> clazz) throws IntrospectionException {
-        synchronized (beanInfoCache) {
-            BeanInfo beanInfo = beanInfoCache.get(clazz);
-            if (beanInfo == null) {
-                beanInfo = Introspector.getBeanInfo(clazz, Object.class);
-                beanInfoCache.putIfAbsent(clazz, beanInfo);
+        try {
+            return beanInfoCache.computeIfAbsent(clazz, k -> {
```
kusalk, line 679: By removing the `synchronized` blcok and using `#computeIfAbsent` we uplift performance as now we only lock on the class name rather than whenever we enter this method

core/src/main/java/org/apache/struts2/ognl/OgnlUtil.java
```diff
@@ -676,13 +675,19 @@ public BeanInfo getBeanInfo(Object from) throws IntrospectionException {
...
+                try {
+                    return Introspector.getBeanInfo(k, Object.class);
+                } catch (IntrospectionException e) {
+                    throw new IllegalArgumentException(e);
+                }
+            });
+        } catch (IllegalArgumentException e) {
```
kusalk, line 686: This is rather ugly but it's to ensure backwards compatibility with respect to the exception throwing behaviour. The lambda we pass to `#computeIfAbsent` doesn't allow us to throw checked exceptions

core/src/main/java/org/apache/struts2/util/ProxyUtil.java
```diff
@@ -65,15 +66,18 @@ public class ProxyUtil {
...
-        Class<?> result = null;
-        if (isSpringAopProxy(candidate))
-            result = springUltimateTargetClass(candidate);
-
-        if (result == null) {
-            result = candidate.getClass();
-        }
-
-        return result;
+        return targetClassCache.computeIfAbsent(candidate, k -> {
```
kusalk, line 69: Added caching here too to uplift `SecurityMemberAccess` performance for applications which have chosen to allow Hibernate and Spring proxies

core/src/main/java/org/apache/struts2/util/ProxyUtil.java
```diff
@@ -44,17 +48,14 @@
  *
  */
 public class ProxyUtil {
-    private static final String SPRING_ADVISED_CLASS_NAME = "org.springframework.aop.framework.Advised";
```
kusalk, line 47: Removed these fields which were previously used for reflection based code - we now use the optional Spring dependency with catch clauses

lukaszlenart:
> `org.apache.struts2.util.ProxyUtil` is covered by `org.apache.struts2.spring.SpringProxyUtilTest` but it seems Sonar doesn't recognise this due to it being in a different module.

I think I must put back the code in GH Actions about branches, in theory Sonar should figure it out automatically, but it doesn't :\

core/src/main/java/org/apache/struts2/util/ProxyUtil.java
kusalk: This specific cache does use the object as the key rather than the Class.

> Also, as I understand, action is a request scope, so I'm not sure about what is the benefit of caching it

This is true but the same request-scoped Action can be executed upon in multiple OGNL expressions. Also, Actions aren't the only objects being cached here.

But yes you are right, it means all other referred objects from these 10,000 cache keys will also be prevented from being GC'd.

I think perhaps we should reduce the max size of this specific cache and also change the cache keys to weak references. I will see what I can do here.

The best workaround you could apply today would be to use reflection upon application startup to reinitialise the `targetClassCache` static field with a cache of lower capacity.

kusalk: Ah yes, good point.

One other suggestion I can make is to write some code to set the memory-heavy references on the Action instance to null after the request has completed or neared completion. This will allow those objects to be garbage collected even when the cache is holding a reference to the original Action instance.

Otherwise, you can also consider building a custom/forked Struts JAR from source until a new release is made available.

ahkamel: Thanks, I'll think about building a custom version in the meantime. Do you know what the schedule of the next release?

PR #1237, opened 2025-02-27
WW-5534 Allow @StrutsParameter recognition and OGNL allowlist for Spring proxies

kusalk:
WW-5534
--
This change fixes `@StrutsParameter` detection for Spring proxied Actions. It additionally adds optional support for resolving Spring proxies when enforcing the OGNL allowlist, alongside the existing support for resolving Hibernate proxies.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
@@ -33,12 +33,14 @@ import org.apache.struts2.dispatcher.HttpParameters;
 import org.apache.struts2.dispatcher.Parameter;
 import org.apache.struts2.inject.Inject;
 import org.apache.struts2.interceptor.MethodFilterInterceptor;
+import org.apache.struts2.ognl.OgnlUtil;
 import org.apache.struts2.ognl.ThreadAllowlist;
 import org.apache.struts2.security.AcceptedPatternsChecker;
 import org.apache.struts2.security.DefaultAcceptedPatternsChecker;
 import org.apache.struts2.security.ExcludedPatternsChecker;
 import org.apache.struts2.util.ClearableValueStack;
 import org.apache.struts2.util.MemberAccessValueStack;
+import org.apache.struts2.util.ProxyUtil;
 import org.apache.struts2.util.TextParseUtil;
 import org.apache.struts2.util.ValueStack;
 import org.apache.struts2.util.ValueStackFactory;
@@ -46,7 +48,6 @@ import org.apache.struts2.util.reflection.ReflectionContextState;
 
 import java.beans.BeanInfo;
 import java.beans.IntrospectionException;
-import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Field;
@@ -93,6 +94,7 @@ public class ParametersInterceptor extends MethodFilterInterceptor {
     protected boolean requireAnnotationsTransitionMode = false;
 
     private ValueStackFactory valueStackFactory;
+    private OgnlUtil ognlUtil;
     protected ThreadAllowlist threadAllowlist;
     private ExcludedPatternsChecker excludedPatterns;
     private AcceptedPatternsChecker acceptedPatterns;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -510,9 +519,16 @@ protected StrutsParameter getParameterAnnotation(AnnotatedElement element) {
...
+        if (ProxyUtil.isProxy(action)) {
+            return ProxyUtil.ultimateTargetClass(action);
+        }
+        return action.getClass();
+    }
+
     protected BeanInfo getBeanInfo(Object action) {
         try {
-            return Introspector.getBeanInfo(action.getClass());
+            return ognlUtil.getBeanInfo(ultimateClass(action));
```
kusalk, line 531: When I implemented this class, I forgot `OgnlUtil` already had a cached variant of this capability. We are now using that, and resolving any proxies to ensure annotation detection works as expected.

core/src/test/java/org/apache/struts2/interceptor/parameter/StrutsParameterAnnotationTest.java
kusalk, line 1: I added comments for all the other tests in this class as I realised I didn't name them very well initially

core/src/test/java/org/apache/struts2/interceptor/parameter/StrutsParameterAnnotationTest.java
```diff
@@ -276,6 +384,27 @@ public void publicModelPojo() {
         assertThat(threadAllowlist.getAllowlist()).containsExactlyInAnyOrderElementsOf(getParentClasses(Object.class, Pojo.class));
     }
 
+    /**
+     * Models of ModelDriven actions can be injected without any annotations on the Action, even when the Action is
+     * proxied.
+     */
+    @Test
+    public void publicModelPojo_proxied() {
+        var proxyFactory = new ProxyFactory(new ModelAction());
```
kusalk, line 393: We use the Spring proxy factory to create a CGLIB proxy like would occur when a transactional proxy is applied to a concrete Action class (like in the original bug report)

core/src/main/java/org/apache/struts2/ognl/SecurityMemberAccess.java
```diff
@@ -212,16 +212,17 @@ protected boolean checkAllowlist(Object target, Member member) {
...
+            // underlying classes/members. This allows the allowlist capability to continue working and still offer
             // protection in applications where the developer has accepted the risk of allowing OGNL access to Hibernate
-            // entities. This is preferred to having to disable the allowlist capability entirely.
-            Object newTarget = ProxyUtil.getHibernateProxyTarget(target);
-            if (newTarget != target) {
-                logAllowlistHibernateEntity(target, newTarget);
-                target = newTarget;
-                member = ProxyUtil.resolveTargetMember(member, newTarget);
+            // entities and Spring proxies. This is preferred to having to disable the allowlist capability entirely.
+            Class<?> newTargetClass = ProxyUtil.ultimateTargetClass(target);
```
kusalk, line 222: We replaced `#getHibernateProxyTarget` with `#ultimateTargetClass` which can also resolve Spring proxies. This allows the OGNL allowlist to function in the presence of Spring proxies in applications where `struts.disallowProxyObjectAccess` has been reverted to `false`.

core/src/test/java/org/apache/struts2/ognl/SecurityMemberAccessTest.java
```diff
@@ -967,6 +969,22 @@ public void classInclusion_hibernateProxy_allowProxyObjectAccess() throws Except
...
+        SpringService proxyObject = newSpringService();
+        Method proxyMethod = proxyObject.getClass().getMethod("doSomething");
+
+        sma.useEnforceAllowlistEnabled(Boolean.TRUE.toString());
+        sma.useDisallowProxyObjectAccess(Boolean.FALSE.toString());
+        sma.useAllowlistClasses(SpringServiceImpl.class.getName());
+
+        assertTrue(sma.checkAllowlist(proxyObject, proxyMethod));
+    }
+
```
lukaszlenart, line 987: There is a dedicated `SecurityMemberAccessProxyTest` and maybe it would be good to move proxy related test cases there as well.

kusalk: Done!

PR #1243, opened 2025-03-06
WW-5534 Proper fix ModelDriven parameter injection and allowlisting

kusalk:
WW-5534
--
Unfortunately, the previous fix (#1237) was not reliable as `Introspector#getBeanInfo` would return slightly different things depending on the visibility of the class hierarchy.

I decided to simplify the exemption logic for `ModelDriven` Actions by moving the allowlisting code into `ModelDrivenInterceptor`, and have `ParametersInterceptor` skip the validation logic entirely when a `ModelDriven` Action was detected.

I've additionally introduced a helper method `ThreadAllowlist#allowClassHierarchy`, extracting some repeated code.

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/ModelDrivenInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/ModelDrivenInterceptor.java
@@ -20,6 +20,8 @@ package org.apache.struts2.interceptor;
 
 import org.apache.struts2.ActionInvocation;
 import org.apache.struts2.ModelDriven;
+import org.apache.struts2.inject.Inject;
+import org.apache.struts2.ognl.ThreadAllowlist;
 import org.apache.struts2.util.CompoundRoot;
 import org.apache.struts2.util.ValueStack;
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/struts2/interceptor/ModelDrivenInterceptor.java
```diff
@@ -79,20 +81,27 @@
...
 
-        if (action instanceof ModelDriven modelDriven) {
+        if (action instanceof ModelDriven<?> modelDriven) {
             ValueStack stack = invocation.getStack();
             Object model = modelDriven.getModel();
-            if (model !=  null) {
-            	stack.push(model);
+            if (model != null) {
+                stack.push(model);
+                threadAllowlist.allowClassHierarchy(model.getClass());
```
kusalk, line 104: We OGNL allowlist the model class immediately in the `ModelDrivenInterceptor` as we expect this class to be safe and accessible from templates

core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java
```diff
@@ -358,9 +357,8 @@ protected boolean isParameterAnnotatedAndAllowlist(String name, Object action) {
         long paramDepth = name.codePoints().mapToObj(c -> (char) c).filter(NESTING_CHARS::contains).count();
 
         if (action instanceof ModelDriven<?> && !ActionContext.getContext().getValueStack().peek().equals(action)) {
-            LOG.debug("Model driven Action detected, exempting from @StrutsParameter annotation requirement and OGNL allowlisting model type");
-            // (Exempted by annotation on org.apache.struts2.ModelDriven#getModel)
-            return hasValidAnnotatedMember("model", action, paramDepth + 1);
+            LOG.debug("Model driven Action detected, exempting from @StrutsParameter annotation requirement");
+            return true;
```
kusalk, line 361: Just return `true` - I previously had this go through the normal flow but it's unnecessary now that allowlisting is already handled by `ModelDrivenInterceptor`

core/src/main/java/org/apache/struts2/ognl/ThreadAllowlist.java
```diff
@@ -34,6 +36,15 @@ public class ThreadAllowlist {
 
     private final ThreadLocal<Set<Class<?>>> allowlist = new ThreadLocal<>();
 
+    /**
+     * @since 7.1.0
+     */
+    public void allowClassHierarchy(Class<?> clazz) {
```
kusalk, line 42: New allowlisting helper method - generally we always want to allowlist the whole hierarchy. Otherwise, methods on the superclasses may not be callable!

lukaszlenart (approved):
👏

PR #1255, opened 2025-04-27
WW-5544 Marks ReflectionContextFactory as deprecated and uses ActionContext instead

lukaszlenart:
WW-5544
--

Cherry-picks changes from PR #1254


kusalk (approved):

lukaszlenart:
@kusalk should I also remove this deprecated interface based on https://issues.apache.org/jira/browse/WW-5545?

kusalk:
@lukaszlenart Unless we are remediating a vulnerability, we should avoiding deleting APIs that can break existing Struts applications, in minor version releases (should be reserved for major release 8.0). In this case, there may still be Struts applications that rely on injecting the `OgnlReflectionContextFactory` bean and its removal would break them

PR #1262, opened 2025-05-05
WW-5546 Fixes NPE when uploaded file is empty

lukaszlenart:
WW-5546
--

### DIFF
--- a/core/src/test/java/org/apache/struts2/interceptor/ActionFileUploadInterceptorTest.java
+++ b/core/src/test/java/org/apache/struts2/interceptor/ActionFileUploadInterceptorTest.java
@@ -98,6 +98,53 @@ public class ActionFileUploadInterceptorTest extends StrutsInternalTestCase {
         }
     };
 
+    public static final UploadedFile NULL_CONTENT = new UploadedFile() {
+        @Override
+        public Long length() {
+            return 0L;
+        }
+
+        @Override
+        public String getName() {
+            return "";
+        }
+
+        @Override
+        public boolean isFile() {
+            return false;
+        }
+
+        @Override
+        public boolean delete() {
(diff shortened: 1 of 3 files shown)


kusalk (approved):

core/src/test/java/org/apache/struts2/interceptor/ActionFileUploadInterceptorTest.java
```diff
@@ -205,6 +252,22 @@ public void testAcceptFileWithNoFile() {
...
+    public void testAcceptFileWithNoContent() {
+        interceptor.setAllowedTypes("text/plain");
+
+        ValidationAwareSupport validation = new ValidationAwareSupport();
+        boolean notOk = interceptor.acceptFile(validation, NULL_CONTENT, "filename.html", "text/plain", "inputName");
+
+        assertFalse(notOk);
+        assertFalse(validation.getFieldErrors().isEmpty());
+        assertTrue(validation.hasErrors());
+        List<String> errors = validation.getFieldErrors().get("inputName");
```
sepe81, line 264: just a small suggestion to use AssertJ:

        assertThat(validation.getFieldErrors().get("inputName"))
            .hasSize(1).first(STRING)
            .startsWith("Error uploading:")
            .contains("inputName");

sepe81 (approved):

core/src/test/java/org/apache/struts2/interceptor/ActionFileUploadInterceptorTest.java
lukaszlenart: I want to make this change as small as possible, I will do a large refactoring once cherry-picking changes into Struts 7

PR #1265, opened 2025-05-08
WW-5548 Defines proper request attributes when forwarding or including final path

lukaszlenart:
WW-5548
--

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/ServletDispatcherResult.java
+++ b/core/src/main/java/org/apache/struts2/result/ServletDispatcherResult.java
@@ -179,7 +179,7 @@ public class ServletDispatcherResult extends StrutsResultSupport {
             } else {
                 LOG.debug("Including location: {}", finalLocation);
 
-                // These attributes are required when forwarding to another servlet, see specification:
+                // These attributes are required when including another resource, see specification:
                 // https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#included-request-parameters
                 request.setAttribute(RequestDispatcher.INCLUDE_MAPPING, request.getHttpServletMapping());
                 request.setAttribute(RequestDispatcher.INCLUDE_REQUEST_URI, request.getRequestURI());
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/struts2/result/ServletDispatcherResult.java
```diff
@@ -173,9 +166,28 @@ public void doExecute(String finalLocation, ActionInvocation invocation) throws
                 request.setAttribute("struts.view_uri", finalLocation);
                 request.setAttribute("struts.request_uri", request.getRequestURI());
 
+                // These attributes are required when forwarding to another servlet, see specification:
```
MFAshby, line 169: I think setting these is the responsibility of the servlet container e.g. tomcat, inside the implementation of the `forward` method. We should not have to do it here at all.

core/src/main/java/org/apache/struts2/result/ServletDispatcherResult.java
```diff
@@ -173,9 +166,28 @@ public void doExecute(String finalLocation, ActionInvocation invocation) throws
...
+                // https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#forwarded-request-parameters
+                request.setAttribute(RequestDispatcher.FORWARD_MAPPING, request.getHttpServletMapping());
+                request.setAttribute(RequestDispatcher.FORWARD_REQUEST_URI, request.getRequestURI());
+                request.setAttribute(RequestDispatcher.FORWARD_CONTEXT_PATH, request.getContextPath());
+                request.setAttribute(RequestDispatcher.FORWARD_SERVLET_PATH, request.getServletPath());
```
MFAshby, line 174: I think this is missing a guard to set the attribute only if it is not already set. There can be multiple forwards, and this attribute should contain the _original_ request as far as I understand. 

Per the other comment I think this is redundant, the servlet container should be setting these properties and we should not touch them.

MFAshby (changes requested):
Thanks very much for your swift response to this issue! It's very much appreciated.

Please see inline comments regarding some issues I still see.

MFAshby (approved):
Fabulous, thank you very much! 😀

lukaszlenart:
@MFAshby thanks a lot for review, you were right and these attributes should be controlled by the container.

lukaszlenart:
This change broke integration with Sitemesh if you want to decorate the final JSP, see the example
https://github.com/apache/struts-examples/pull/461

PR #1294, opened 2025-07-16
Report required depth of rejected parameters

yasuoz:
The standard ParametersInterceptor does not report the depth that should be set for StrutsParameter, so we added it.


lukaszlenart:
Please create a JIRA ticket to cover this change, this helps when releasing a new verison.
https://issues.apache.org/jira/browse/WW

PR #1309, opened 2025-08-07
WW-5524 Fixup StrutsConverterFactory

MFAshby:
It should delegate back to ObjectFactory#buildBean instead of directly calling Container#inject, otherwise overrides of buildBean in subclasses of ObjectFactory e.g. SpringObjectFactory are skipped; meaning that TypeConverters cannot make use of Spring dependency injection

Fixes: https://issues.apache.org/jira/projects/WW/issues/WW-5524


lukaszlenart:
Cherry picked #1316

PR #1377, opened 2025-10-01
WW-5578 Don't catch and swallow ConfigurationException in InterceptorBuilder.

MFAshby:
WW-5578
--

This can mask configuration errors and result in a non-working application.


kusalk:
Good call, interceptors can be used to enforce auth/security and so failing fast is indeed desirable. If users have a genuine need for an optionally loaded interceptor, then we should implement some mechanism to explicitly declare that.

kusalk (approved):

PR #1651, opened 2026-04-06
fix(json-plugin): WW-5624 enforce @StrutsParameter in JSONPopulator.populateObject()

tranquac:
## Summary

`JSONPopulator.populateObject()` in the struts2-json-plugin sets action properties via direct Java reflection (`Method.invoke()`) without checking the `@StrutsParameter` annotation. This bypasses the parameter allowlisting that `ParametersInterceptor` enforces for URL parameters, enabling mass assignment of unannotated properties via JSON request body.

## Changes

- **JSONPopulator.java**: Add `@StrutsParameter` annotation check before `Method.invoke()`. When `struts.parameters.requireAnnotations` is enabled and the setter lacks the annotation, the property is skipped with a debug log message.
- **JSONInterceptor.java**: Wire the `struts.parameters.requireAnnotations` setting via `@Inject` into `JSONPopulator.setRequireAnnotations()`.

## Impact

Without this fix:
- URL params to unannotated setters: **BLOCKED** (ParametersInterceptor enforces)
- JSON body to same unannotated setters: **BYPASSES** (JSONPopulator ignores annotation)

With this fix, both pathways consistently enforce `@StrutsParameter`.

## Test

A PoC application with 8 test cases demonstrates the bypass and fix. An action with `@StrutsParameter`-annotated setter (`setUsername`) and unannotated setters (`setRole`, `setAdmin`) shows that JSON body sets unannotated fields before the fix, and blocks them after.

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONInterceptor.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONInterceptor.java
@@ -378,6 +378,12 @@ public class JSONInterceptor extends AbstractInterceptor {
         this.defaultEncoding = val;
     }
 
+    @Inject(value = StrutsConstants.STRUTS_PARAMETERS_REQUIRE_ANNOTATIONS, required = false)
+    public void setRequireAnnotations(String requireAnnotations) {
+        boolean required = BooleanUtils.toBoolean(requireAnnotations);
+        this.populator.setRequireAnnotations(required);
+    }
+
     /**
      * @param ignoreHierarchy Ignore properties defined on base classes of the root object.
      */
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONPopulator.java
@@ -20,6 +20,7 @@ package org.apache.struts2.json;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.struts2.interceptor.parameter.StrutsParameter;
 import org.apache.struts2.json.annotations.JSON;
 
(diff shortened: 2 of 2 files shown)


lukaszlenart:
Could you create a JIRA ticket first? I can include also changes in #1652

tranquac:
hi @lukaszlenart

How can I create a JIRA ticket?

have a good day!
Quac Tran

tranquac:
I requested a Jira account!

And i will create create a JIRA ticket soon!

tranquac:
Hi @lukaszlenart 

I created JIRA ticket for this issue: https://issues.apache.org/jira/projects/WW/issues/WW-5624?filter=allissues

lukaszlenart:
Thanks for reporting this. The issue in the JSON plugin looks valid: `JSONPopulator.populateObject()` currently bypasses the `ParametersInterceptor` checks for `@StrutsParameter`.

That said, I don't think this fix is sufficient as-is. In Struts, `@StrutsParameter` support is not only a check for an annotated setter. The core `ParametersInterceptor` logic also enforces nesting depth and related authorization semantics. This change currently adds only a setter-level annotation check, so it does not fully match existing Struts behavior.

My main concern is that `JSONPopulator` recursively populates nested beans, collections, arrays, and maps. Requiring `@StrutsParameter` on every nested setter would be stricter than current Struts semantics, where authorization is based on the exposed root member and the permitted depth. That could break valid nested binding scenarios while still not aligning fully with the core parameter-binding contract.

I think the right direction is to align the JSON plugin with the shared parameter-binding authorization rules used by `ParametersInterceptor`, instead of implementing a local setter-only check.

tranquac:

lukaszlenart:
Superseded by #1657

PR #1653, opened 2026-04-06
fix(core): WW-5623 HTML-encode form action in PostbackResult to prevent XSS

tranquac:
## Summary

PostbackResult.doExecute() at line 107 embeds finalLocation into a form action attribute via raw string concatenation without HTML encoding. The response Content-Type is text/html (line 103). A double-quote character in the location breaks out of the attribute, enabling reflected XSS.

This is an encoding inconsistency: form field names and values at lines 218-219 ARE properly URL-encoded via URLEncoder.encode(), but the form action attribute was not encoded at all.

## Changes

- **PostbackResult.java**: Add encodeHtml() method to escape &, ", <, > in finalLocation before embedding in the HTML form tag. Consistent with the existing encoding approach for form field values.

## Impact

When a developer uses PostbackResult with an OGNL expression referencing a user-controllable property (a documented framework feature for dynamic routing), an attacker can inject arbitrary HTML attributes and elements via the form action attribute.

## Test

A PoC application with 5 test scenarios verifies the vulnerability and fix. Browser-based testing with Playwright confirms the XSS alert fires before the fix.

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/PostbackResult.java
+++ b/core/src/main/java/org/apache/struts2/result/PostbackResult.java
@@ -104,7 +104,8 @@ public class PostbackResult extends StrutsResultSupport {
 
         // Render
         PrintWriter pw = new PrintWriter(response.getOutputStream());
-        pw.write("<!DOCTYPE html><html><body><form action=\"" + finalLocation + "\" method=\"POST\">");
+        String safeLocation = encodeHtml(finalLocation);
+        pw.write("<!DOCTYPE html><html><body><form action=\"" + safeLocation + "\" method=\"POST\">");
         writeFormElements(request, pw);
         writePrologueScript(pw);
         pw.write("</html>");
@@ -211,6 +212,19 @@ public class PostbackResult extends StrutsResultSupport {
 
     public final void setPrependServletContext(boolean prependServletContext) {
         this.prependServletContext = prependServletContext;
+    }
+
+    /**
+     * Encodes special HTML characters to prevent injection in HTML attribute context.
+     */
+    private static String encodeHtml(String value) {
+        if (value == null) {
+            return "";
+        }
+        return value.replace("&", "&amp;")
+                     .replace("\"", "&quot;")
+                     .replace("<", "&lt;")
+                     .replace(">", "&gt;");
     }
 
     protected void writeFormElement(PrintWriter pw, String name, String[] values) throws UnsupportedEncodingException {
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Could you create a JIRA ticket first?

tranquac:
I requested a Jira account!

And i will create create a JIRA ticket soon!

tranquac:
hello @lukaszlenart 

I created JIRA ticket for this issue: https://issues.apache.org/jira/projects/WW/issues/WW-5623?filter=allissues

lukaszlenart:
Please add a focused unit test for this change in `PostbackResultTest`, covering a malicious `finalLocation` with characters like `"`, `&`, `<`, and `>` and asserting they are properly escaped in the rendered `action` attribute. The fix looks correct for WW-5623, but it should include regression coverage. Also, for consistency with the rest of Struts core, you may want to use `StringEscapeUtils.escapeHtml4(...)` instead of a custom `encodeHtml()` helper.

tranquac:

tranquac:
Hi @lukaszlenart,

Thank you for the review feedback! I've pushed a follow-up commit addressing both points:

### Changes in this update

**1. Replaced custom `encodeHtml()` with `StringEscapeUtils.escapeHtml4()`**

Removed the custom helper method and switched to `org.apache.commons.text.StringEscapeUtils.escapeHtml4()`, consistent with how the rest of Struts core handles HTML encoding (e.g., `DefaultActionProxy`, `Property`, `TextProviderHelper`).

**2. Added 3 focused unit tests in `PostbackResultTest`**

| Test | What it covers |
|---|---|
| `testFormActionHtmlEscaping` | XSS payload with double-quote attribute breakout (`"onmouseover="alert(1)"`). Asserts escaped entities in rendered action attribute and verifies raw `"` cannot break out. |
| `testFormActionEscapesAllHtmlSpecialChars` | Covers all 4 critical HTML characters (`"`, `&`, `<`, `>`) individually, asserting each is properly escaped to its HTML entity (`&quot;`, `&amp;`, `&lt;`, `&gt;`). |
| `testFormActionCleanLocationUnchanged` | Regression test — verifies that a clean URL without special characters renders unchanged in the action attribute. |

All 7 tests pass (4 existing + 3 new):
```
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Please let me know if there's anything else you'd like me to adjust.

tranquac:
If there are any further issues that need correction or addition, let me know so I can improve it. Thanks @lukaszlenart and reviewers!

rgielen (approved):
LGTM 👍

tranquac:
Hi @jogep, @aleksandr-m, @yasserzamani, @sdutry, @cnenning, @kusalk

Could you pls help me check this PR?

lukaszlenart (approved):
LGTM 👍

lukaszlenart:
## Just one important remark

This PR doesn't address a vulnerability, it's pure framework hardening. The potential attack vector assumes a developer uses a raw user input in OGNL expression evaluation which is against our guideline:
https://struts.apache.org/security/#do-not-use-incoming-untrusted-user-input-in-forced-expression-evaluation

PR #1657, opened 2026-04-09
WW-5624: Enforce @StrutsParameter on JSON/REST body deserialization

tranquac:
## Summary

Fixes mass assignment bypass where JSON body (`JSONInterceptor`) and REST body (`ContentTypeInterceptor`) deserialization bypassed `@StrutsParameter` annotation checks, even when `struts.parameters.requireAnnotations=true`.

This patch:
- Extracts shared `ParameterAuthorizer` service from `ParametersInterceptor` (no code duplication)
- Wires it into json-plugin and rest-plugin
- **JSON plugin**: Filters unauthorized Map keys before `populateObject()`
- **REST plugin**: Two-phase deserialization when `requireAnnotations=true`; direct deserialization for backward compat when disabled
- OGNL `ThreadAllowlist` side effects remain in `ParametersInterceptor` only
- Full 3-level DI wiring

## Design Decisions

1. **Shared service**: Authorization logic extracted once into `DefaultParameterAuthorizer` and shared by all interceptors.
2. **No OGNL side effects**: `ThreadAllowlist` operations stay in `ParametersInterceptor`. JSON/REST plugins use reflection/Jackson, not OGNL.
3. **REST gated on requireAnnotations**: When disabled (default), direct `handler.toObject()` for backward compat. When enabled, two-phase deserialization.
4. **JSON top-level filter**: Matches `@StrutsParameter(depth=N)` model since `JSONPopulator` uses reflection for nesting.
5. **RPC/SMD not filtered**: Different trust model (method invocation, not property injection).

## Files Changed (13 files, +810/-10)

- `ParameterAuthorizer.java` (NEW interface)
- `DefaultParameterAuthorizer.java` (NEW impl)
- `ParametersInterceptor.java` (delegates to authorizer)
- `StrutsConstants.java`, `StrutsBeanSelectionProvider.java`, `DefaultConfiguration.java`, `struts-beans.xml` (DI wiring)
- `JSONInterceptor.java` (inject + filter)
- `ContentTypeInterceptor.java` (inject + two-phase)
- 4 test files (19 new tests)

## Test Plan

- [x] ParameterAuthorizerTest: 15 tests
- [x] StrutsParameterAnnotationTest: 32 regression tests
- [x] ParametersInterceptorTest: 35 tests
- [x] JSONInterceptorTest: 24 tests (2 new)
- [x] ContentTypeInterceptorTest: 5 tests (2 new)
- [x] Full suites: json-plugin 122, rest-plugin 76 - zero regressions

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/parameter/DefaultParameterAuthorizer.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/parameter/DefaultParameterAuthorizer.java
@@ -1,0 +1,229 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.struts2.interceptor.parameter;
+
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.struts2.StrutsConstants;
+import org.apache.struts2.inject.Inject;
+import org.apache.struts2.ognl.OgnlUtil;
+import org.apache.struts2.util.ProxyService;
+
+import java.beans.BeanInfo;
+import java.beans.IntrospectionException;
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static java.lang.String.format;
+import static org.apache.commons.lang3.StringUtils.indexOfAny;
+import static org.apache.struts2.security.DefaultAcceptedPatternsChecker.NESTING_CHARS;
+import static org.apache.struts2.security.DefaultAcceptedPatternsChecker.NESTING_CHARS_STR;
+import static org.apache.struts2.util.DebugUtils.notifyDeveloperOfError;
+
+/**
+ * Default implementation of {@link ParameterAuthorizer} that checks {@link StrutsParameter} annotations on the target
+ * object's members to determine whether a parameter is authorized for injection.
+ *
+ * <p>This implementation extracts the authorization logic from {@link ParametersInterceptor} so that it can be shared
+ * with other input channels (JSON plugin, REST plugin) without duplicating code.</p>
+ *
+ * <p>Unlike {@link ParametersInterceptor}, this implementation does NOT perform OGNL ThreadAllowlist side effects.
+ * Those remain specific to the OGNL-based parameter injection path.</p>
+ *
+ * @since 7.2.0
+ */
+public class DefaultParameterAuthorizer implements ParameterAuthorizer {
+
+    private static final Logger LOG = LogManager.getLogger(DefaultParameterAuthorizer.class);
+
+    private boolean requireAnnotations = false;
+    private boolean requireAnnotationsTransitionMode = false;
+    private boolean devMode = false;
+
+    private OgnlUtil ognlUtil;
+    private ProxyService proxyService;
+
+    @Inject
+    public void setOgnlUtil(OgnlUtil ognlUtil) {
+        this.ognlUtil = ognlUtil;
+    }
+
+    @Inject
+    public void setProxyService(ProxyService proxyService) {
+        this.proxyService = proxyService;
+    }
+
+    @Inject(StrutsConstants.STRUTS_DEVMODE)
+    public void setDevMode(String mode) {
+        this.devMode = BooleanUtils.toBoolean(mode);
+    }
+
+    @Inject(value = StrutsConstants.STRUTS_PARAMETERS_REQUIRE_ANNOTATIONS, required = false)
+    public void setRequireAnnotations(String requireAnnotations) {
+        this.requireAnnotations = BooleanUtils.toBoolean(requireAnnotations);
+    }
+
+    @Inject(value = StrutsConstants.STRUTS_PARAMETERS_REQUIRE_ANNOTATIONS_TRANSITION, required = false)
+    public void setRequireAnnotationsTransitionMode(String transitionMode) {
+        this.requireAnnotationsTransitionMode = BooleanUtils.toBoolean(transitionMode);
+    }
+
+    @Override
+    public boolean isAuthorized(String parameterName, Object target, Object action) {
+        if (parameterName == null || parameterName.isEmpty()) {
+            return false;
+        }
+
+        if (!requireAnnotations) {
+            return true;
+        }
+
+        long paramDepth = parameterName.codePoints().mapToObj(c -> (char) c).filter(NESTING_CHARS::contains).count();
+
+        // ModelDriven exemption: when target is the model (target != action), exempt from annotation requirements
+        if (target != action) {
+            LOG.debug("Model driven target detected, exempting from @StrutsParameter annotation requirement");
+            return true;
+        }
+
+        // Transition mode: depth-0 (non-nested) parameters are exempt
+        if (requireAnnotationsTransitionMode && paramDepth == 0) {
+            LOG.debug("Annotation transition mode enabled, exempting non-nested parameter [{}] from @StrutsParameter annotation requirement",
+                    parameterName);
+            return true;
+        }
+
+        int nestingIndex = indexOfAny(parameterName, NESTING_CHARS_STR);
+        String rootProperty = nestingIndex == -1 ? parameterName : parameterName.substring(0, nestingIndex);
+        String normalisedRootProperty = Character.toLowerCase(rootProperty.charAt(0)) + rootProperty.substring(1);
+
+        return hasValidAnnotatedMember(normalisedRootProperty, target, paramDepth);
+    }
+
+    protected boolean hasValidAnnotatedMember(String rootProperty, Object target, long paramDepth) {
+        LOG.debug("Checking target [{}] for a matching, correctly annotated member for property [{}]",
+                target.getClass().getSimpleName(), rootProperty);
+        BeanInfo beanInfo = getBeanInfo(target);
+        if (beanInfo == null) {
+            return hasValidAnnotatedField(target, rootProperty, paramDepth);
+        }
+
+        Optional<PropertyDescriptor> propDescOpt = Arrays.stream(beanInfo.getPropertyDescriptors())
+                .filter(desc -> desc.getName().equals(rootProperty)).findFirst();
+        if (propDescOpt.isEmpty()) {
+            return hasValidAnnotatedField(target, rootProperty, paramDepth);
+        }
+
+        if (hasValidAnnotatedPropertyDescriptor(target, propDescOpt.get(), paramDepth)) {
+            return true;
+        }
+
+        return hasValidAnnotatedField(target, rootProperty, paramDepth);
+    }
+
+    protected boolean hasValidAnnotatedPropertyDescriptor(Object target, PropertyDescriptor propDesc, long paramDepth) {
+        Class<?> targetClass = ultimateClass(target);
+        Method relevantMethod = paramDepth == 0 ? propDesc.getWriteMethod() : propDesc.getReadMethod();
+        if (relevantMethod == null) {
+            return false;
+        }
+        if (getPermittedInjectionDepth(relevantMethod) < paramDepth) {
+            String logMessage = format(
+                    "Parameter injection for method [%s] on target [%s] rejected. Ensure it is annotated with @StrutsParameter with an appropriate 'depth'.",
+                    relevantMethod.getName(),
+                    relevantMethod.getDeclaringClass().getName());
+            if (devMode) {
+                notifyDeveloperOfError(LOG, target, logMessage);
+            } else {
+                LOG.debug(logMessage);
+            }
+            return false;
+        }
+        LOG.debug("Success: Matching annotated method [{}] found for property [{}] of depth [{}] on target [{}]",
+                relevantMethod.getName(), propDesc.getName(), paramDepth, targetClass.getSimpleName());
+        return true;
+    }
+
+    protected boolean hasValidAnnotatedField(Object target, String fieldName, long paramDepth) {
+        Class<?> targetClass = ultimateClass(target);
+        LOG.debug("No matching annotated method found for property [{}] of depth [{}] on target [{}], now also checking for public field",
+                fieldName, paramDepth, targetClass.getSimpleName());
+        Field field;
+        try {
+            field = targetClass.getDeclaredField(fieldName);
+        } catch (NoSuchFieldException e) {
+            LOG.debug("Matching field for property [{}] not found on target [{}]", fieldName, targetClass.getSimpleName());
+            return false;
+        }
+        if (!Modifier.isPublic(field.getModifiers())) {
+            LOG.debug("Matching field [{}] is not public on target [{}]", field.getName(), targetClass.getSimpleName());
+            return false;
+        }
+        if (getPermittedInjectionDepth(field) < paramDepth) {
+            String logMessage = format(
+                    "Parameter injection for field [%s] on target [%s] rejected. Ensure it is annotated with @StrutsParameter with an appropriate 'depth'.",
+                    field.getName(),
+                    targetClass.getName());
+            if (devMode) {
+                notifyDeveloperOfError(LOG, target, logMessage);
+            } else {
+                LOG.debug(logMessage);
+            }
+            return false;
+        }
+        LOG.debug("Success: Matching annotated public field [{}] found for property of depth [{}] on target [{}]",
+                field.getName(), paramDepth, targetClass.getSimpleName());
+        return true;
+    }
+
+    protected int getPermittedInjectionDepth(AnnotatedElement element) {
+        StrutsParameter annotation = getParameterAnnotation(element);
+        if (annotation == null) {
+            return -1;
+        }
+        return annotation.depth();
+    }
+
+    protected StrutsParameter getParameterAnnotation(AnnotatedElement element) {
+        return element.getAnnotation(StrutsParameter.class);
+    }
+
+    protected Class<?> ultimateClass(Object target) {
+        if (proxyService.isProxy(target)) {
+            return proxyService.ultimateTargetClass(target);
+        }
+        return target.getClass();
+    }
+
+    protected BeanInfo getBeanInfo(Object target) {
+        Class<?> targetClass = ultimateClass(target);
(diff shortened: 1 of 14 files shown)


tranquac:
This PR supersedes #1651 and #1652 with a consolidated approach that addresses the review feedback:

1. **Shared service, no duplication**: `ParameterAuthorizer` is extracted once from `ParametersInterceptor` and shared by JSON/REST plugins (as requested by @lukaszlenart).
2. **OGNL isolation**: `ThreadAllowlist` side effects stay in `ParametersInterceptor` only — the shared service is OGNL-free.
3. **Backward compatible**: REST two-phase deserialization only activates when `struts.parameters.requireAnnotations=true`. Direct deserialization remains the default.
4. **Complete DI wiring**: All 3 levels (struts-beans.xml, StrutsBeanSelectionProvider alias, DefaultConfiguration factory) are registered.
5. **Comprehensive tests**: 19 new tests covering authorization, rejection, ModelDriven exemption, transition mode, depth enforcement, and null/empty parameter guards. All 280 existing tests across core/json/rest pass.

core/src/main/java/org/apache/struts2/interceptor/parameter/DefaultParameterAuthorizer.java
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * <p>This implementation extracts the authorization logic from {@link ParametersInterceptor} so that it can be shared
+ * with other input channels (JSON plugin, REST plugin) without duplicating code.</p>
+ *
+ * <p>Unlike {@link ParametersInterceptor}, this implementation does NOT perform OGNL ThreadAllowlist side effects.
+ * Those remain specific to the OGNL-based parameter injection path.</p>
+ *
+ * @since 7.2.0
+ */
+public class DefaultParameterAuthorizer implements ParameterAuthorizer {
```
lukaszlenart: I prefer to use `Struts` prefix as a clear indicator that this a framework implementation of the interface.
```suggestion
public class StrutsParameterAuthorizer implements ParameterAuthorizer {
```

lukaszlenart:
The new PR is much closer to the right direction than `#1651` / `#1652` because it extracts shared authorization logic and avoids duplicating ad hoc checks in each plugin. I still see a few remaining gaps that should be addressed before this is fully aligned with `ParametersInterceptor` semantics:

issue (blocking): The `ModelDriven` exemption in `DefaultParameterAuthorizer` looks too broad. Using `target != action` as the signal means any non-action JSON root object can be treated like a `ModelDriven` model and bypass `@StrutsParameter` checks. I think this should mirror the existing `ParametersInterceptor` semantics more closely, instead of inferring `ModelDriven` purely from object identity.

issue (blocking): `JSONInterceptor.filterUnauthorizedKeys()` only filters top-level JSON keys. That improves the previous behavior, but it still does not enforce nested `@StrutsParameter(depth=N)` semantics. Once an allowed top-level key remains in the map, `JSONPopulator` can still recursively populate the nested object graph underneath it. I think this needs path-aware authorization, not only root-key filtering. Either walk nested maps/lists and authorize synthesized paths recursively, or move the authorization check into the recursive population flow so each nested segment is checked against the same depth rules as `ParametersInterceptor`.

issue (blocking): The two-phase REST approach in `ContentTypeInterceptor.copyAuthorizedProperties()` is better than the previous direct merge, but it still authorizes only top-level bean properties. If an allowed property is a complex object, the full nested graph deserialized into the staging instance is copied over in one step without nested authorization. I think the copy phase needs to become a deep authorized merge, not a shallow bean-property copy.

issue: `copyAuthorizedProperties()` currently skips `null` values, which changes semantics when `struts.parameters.requireAnnotations=true`. That means an explicitly provided `null` in the request body can no longer clear an authorized property. Please confirm whether this behavior change is intentional. If not, explicit nulls for authorized properties should probably still be copied. If it is intentional, it should be documented and covered by tests.

issue: The two-phase deserialization path now requires a public no-arg constructor on the target type. That is a new constraint which may break existing actions/models when `requireAnnotations=true`. Please verify whether this constructor requirement is acceptable for Struts REST usage. If it is, I would at least add explicit tests and mention it in the PR/JIRA notes as a compatibility constraint.

suggestion: Please also add regression tests for transition mode with nested JSON bodies, and for `JSONInterceptor.root` resolving to a non-action object. Those are the cases most likely to expose the remaining semantic gaps.

lukaszlenart (changes requested):

lukaszlenart:
I checked the `alias()` wiring against `BeanSelectionProvider` / `AbstractBeanSelectionProvider`, and this part looks correct. The new `ParameterAuthorizer` follows the existing bean-selection pattern: config key in `StrutsConstants`, alias registration in `StrutsBeanSelectionProvider`, and default implementation registration via factory / named bean.

tranquac:
Thanks for the thorough review @lukaszlenart! I've pushed a new commit ([aa78085](https://github.com/tranquac/struts/commit/aa78085)) addressing all 5 blocking issues plus the inline suggestion. Here's a point-by-point summary:

---

### 1. Rename `DefaultParameterAuthorizer` → `StrutsParameterAuthorizer` ✅

Done — class renamed, all references updated (`struts-beans.xml`, `DefaultConfiguration.java`, both test files). Git detected it as a rename (93% similarity).

---

### 2. ModelDriven exemption too broad ✅

Changed the guard in `StrutsParameterAuthorizer.isAuthorized()` from:
```java
if (target != action) { return true; }
```
to:
```java
if (target != action && action instanceof ModelDriven) { return true; }
```
Now only actions that explicitly implement `ModelDriven` will have their model exempted. If a `JSONInterceptor` `root` configuration resolves to a non-action object on a non-ModelDriven action, annotation checks are still enforced. Added a regression test `nonModelDrivenAction_differentTarget_notExempt()` to lock this in.

---

### 3. Nested enforcement missing in `JSONInterceptor` ✅

`filterUnauthorizedKeys()` now delegates to a recursive `filterUnauthorizedKeysRecursive()` that builds dot-notation paths (`"address.city"`) as it descends into nested Maps and Lists:

```java
String fullPath = prefix.isEmpty() ? key : prefix + "." + key;
if (!parameterAuthorizer.isAuthorized(fullPath, target, action)) { it.remove(); }
// recurse into nested Map / List
```

This reuses the existing `@StrutsParameter(depth=N)` depth-counting in `StrutsParameterAuthorizer` — a nested key `"address.city"` has depth=1, which requires `getAddress()` to carry `@StrutsParameter(depth≥1)`. Added regression tests for nested key filtering and for the non-action root object case.

---

### 4. REST shallow copy problem ✅

`copyAuthorizedProperties()` is now recursive. For each property, if the value is a *complex bean type* (not primitive, String, Number, Date, Temporal, Collection, Map, array, enum), it creates a fresh target instance and recurses with the extended path prefix:

```java
if (isNestedBeanType(sourceValue.getClass())) {
    copyAuthorizedProperties(sourceValue, targetValue, action, fullPath); // depth-aware recurse
} else {
    writeMethod.invoke(target, sourceValue);
}
```

Collections and Maps are copied as-is — their contents were already deserialized and the depth check on the parent property is sufficient.

---

### 5. Null-skipping semantics — kept intentionally, now documented ✅

After careful consideration: in two-phase deserialization, `null` in `freshInstance` is **ambiguous** — it could mean "field was explicitly set to null in the request body" or "field was not present at all". Copying null would silently wipe pre-initialized fields on the target for every property not sent by the client.

The safer default is to skip null, and I've added an inline comment explaining the reasoning:
```java
// Intentionally skip null values: in two-phase deserialization, properties NOT present in the
// request body will be null in the fresh instance. Copying null would clear pre-initialized
// fields on the target. This is the safer default.
```

If you feel strongly that explicit null should be propagated, I can add a `clearMissingFields` flag (default `false`) — but I wanted your input before adding API surface.

---

### 6. No-arg constructor requirement ✅

Added a `createFreshInstance()` helper that catches `ReflectiveOperationException` and returns `null` on failure. The interceptor now has two paths:

- **Two-phase (preferred)**: deserialize into `freshInstance`, then `copyAuthorizedProperties()` — requires no-arg constructor
- **Single-phase fallback**: deserialize directly into `target`, then `scrubUnauthorizedProperties()` nulls out any unauthorized property — no constructor requirement

The scrub fallback is logged at `WARN` so operators know the less-safe path is active. Backward compat fully preserved.

---

### Test results

```
core:         2920 tests — 0 failures, 0 errors
plugins/json:  124 tests — 0 failures, 0 errors  
plugins/rest:   76 tests — 0 failures, 0 errors
```

New regression tests added:
- `ParameterAuthorizerTest#nonModelDrivenAction_differentTarget_notExempt`
- `JSONInterceptorTest#testNestedJsonKeysRecursivelyFiltered`
- `JSONInterceptorTest#testNonActionRootObjectStillChecked`

Please let me know if anything needs further adjustment!

lukaszlenart:
Thanks, this is clearly closer to the intended direction and it addresses several of the earlier concerns. I still see one blocking parity gap with `ParametersInterceptor`: collection/indexed paths.

In core, authorization depth is based on the full parameter path, e.g. `publicPojoListDepthOne[0].key` must be rejected while `publicPojoListDepthTwo[0].key` is accepted. The updated JSON list recursion appears to reuse the same prefix for list items instead of synthesizing an indexed path, which undercounts nesting and can make list element binding more permissive than `ParametersInterceptor`.

I also still have the same concern for the REST side where `Collection` / `Map` / arrays are copied as-is, and for the fallback path that does direct deserialization followed by top-level scrubbing only. Those paths still do not look fully equivalent to the core nested authorization model.

tranquac:
Thanks for the detailed review. v3 patch (commit `9490dfa`) addresses all four gaps you identified:

---

### Gap 1 — JSON: indexed-path depth for list elements

`filterUnauthorizedList` was passing the bare `prefix` unchanged into `filterUnauthorizedKeysRecursive`, so a path like `publicPojoListDepthOne.key` (depth=1) was being checked instead of `publicPojoListDepthOne[0].key` (depth=2). That let list element properties pass under a lower depth annotation than `ParametersInterceptor` would allow.

Fix: pass `prefix + "[0]"` as `elementPrefix` for all items in the list. `NESTING_CHARS` includes `[`, so the extra bracket increments `paramDepth` by 1, exactly matching the indexed-path depth semantics:

```java
String elementPrefix = prefix + "[0]";
for (Object item : list) {
    if (item instanceof Map) {
        filterUnauthorizedKeysRecursive((Map) item, elementPrefix, target, action);
    } else if (item instanceof java.util.List) {
        filterUnauthorizedList((java.util.List) item, elementPrefix, target, action);
    }
}
```

The nested-list `else if` branch handles `List<List<Map>>` shapes that the previous version left unfiltered.

---

### Gap 2 — REST: `authTarget` always = root action/model

`isAuthorized(fullPath, target, action)` uses `target` to look up the root property annotation. On recursive calls, `target` was being replaced with the nested object (e.g., `Address`), so `isAuthorized("address.city", addressObject, action)` looked for a property named `"address"` on `Address` — not found, silently rejecting all nested properties.

Fix: add a new `authTarget` parameter to `copyAuthorizedProperties`, always set to the root action/model at the initial call site, and passed unchanged through every level of recursion. All `isAuthorized` calls now use `authTarget`, never the traversal-level `target`:

```java
// Initial call:
copyAuthorizedProperties(freshInstance, target, invocation.getAction(), target, "");

// Signature:
private void copyAuthorizedProperties(
        Object source, Object target, Object action, Object authTarget, String prefix)

// All recursive calls:
copyAuthorizedProperties(sourceValue, targetValue, action, authTarget, fullPath);
```

---

### Gap 3 — REST: Collection / Map / array deep authorization

Previously `Collection`, `Map`, and `Pojo[]` properties were copied as-is after the parent-level path check. That skipped element-level authorization entirely.

Three new helpers — `deepCopyAuthorizedCollection`, `deepCopyAuthorizedMap`, `deepCopyAuthorizedArray` — iterate every element/value and:
1. Check `path + "[0]"` against `authTarget` (adds one `[` → depth +1, matching `ParametersInterceptor`)
2. Create a fresh instance and recurse via `copyAuthorizedProperties` for complex types
3. **Skip** the element (rather than copy it) when no no-arg constructor is available — avoids leaking an unfiltered object graph

---

### Gap 4 — REST: `scrubUnauthorizedProperties` fallback — full recursion

`scrubUnauthorizedProperties` previously only iterated top-level property descriptors, leaving nested beans and collection/map elements unscrubbed.

Fix: refactored into `scrubUnauthorizedPropertiesRecursive` which:
- Recurses into nested beans with `fullPath` prefix
- Iterates `Collection` elements with `fullPath + "[0]"` prefix
- Iterates `Map` values with `fullPath + "[0]"` prefix (new, missing from the earlier plan)
- Uses an identity-based `visited` set (`System.identityHashCode`) to break cycles
- Passes `authTarget` (root object) unchanged at every level

---

### Test results

```
core:         2920 tests, 0 failures
plugins/json:  124 tests, 0 failures
plugins/rest:   76 tests, 0 failures
Total:         3120 tests, 0 failures
```

Let me know if there is anything further you would like adjusted.

tranquac:
Follow-up commit `df533b3` fixes three additional correctness issues found during an independent review of the v3 patch:

---

### Fix 1 — Collection/Map type preservation in deep copy helpers

`deepCopyAuthorizedCollection` was always returning `new ArrayList()`. If the action field is typed as `Set<Pojo>`, the setter has signature `void setTags(Set<Pojo> s)` — invoking it with an `ArrayList` throws `IllegalArgumentException` at runtime.

Same class of issue in `deepCopyAuthorizedMap` (always returned `LinkedHashMap`).

Fix: inspect the source collection/map type and instantiate the appropriate concrete type:

```java
// Collection:
if (source instanceof SortedSet) {
    result = new TreeSet(((SortedSet) source).comparator());
} else if (source instanceof Set) {
    result = new LinkedHashSet();
} else {
    result = new ArrayList();
}

// Map:
if (source instanceof SortedMap) {
    result = new TreeMap(((SortedMap) source).comparator());
} else {
    result = new LinkedHashMap();
}
```

---

### Fix 2 — Identity-based visited set in `scrubUnauthorizedPropertiesRecursive`

The circular reference guard used `Set<Integer>` + `System.identityHashCode(target)`. `System.identityHashCode` is NOT guaranteed unique — two distinct live objects can have the same identity hash code. A collision would cause a valid nested object to be skipped during scrubbing, leaving unauthorized nested properties un-nulled.

Fix: replaced with `Collections.newSetFromMap(new IdentityHashMap<>())`, which uses reference equality (`==`) and has no collision risk:

```java
Collections.newSetFromMap(new IdentityHashMap<>())  // identity set
// ...
if (!visited.add(target)) { return; }  // target object itself, not its hash
```

---

### Fix 3 — `isNestedBeanType` now excludes all standard-library leaf packages

Previously only `java.lang.*` and `java.math.*` were excluded. Types like `UUID`, `Currency`, `Locale` (in `java.util`), `URI`/`URL` (in `java.net`), and `Path`/`File` (in `java.nio`/`java.io`) returned `true` from `isNestedBeanType`, causing the code to attempt to recurse into their internal fields. This would silently drop fields of those types (no `@StrutsParameter` annotation on internal JDK fields → rejected → value not copied).

Fix: broadened exclusions:

```java
// java.util.* non-Collection/Map leaf types (UUID, Currency, Locale, Date, etc.)
if (clazz.getName().startsWith("java.util.") && !Collection.class.isAssignableFrom(clazz)
        && !Map.class.isAssignableFrom(clazz)) {
    return false;
}
// All java.time.* temporal types
if (clazz.getName().startsWith("java.time.")) return false;
// I/O and network value types
if (clazz.getName().startsWith("java.net.") || clazz.getName().startsWith("java.io.")
        || clazz.getName().startsWith("java.nio.")) {
    return false;
}
```

---

### Test results

```
plugins/json:  124 tests, 0 failures
plugins/rest:   76 tests, 0 failures
```

lukaszlenart:
Thanks, this is much closer now and it addresses most of the earlier concerns.

I still see one remaining blocker in the REST path: `copyAuthorizedProperties(...)` still has a bulk-copy fallback when a nested target bean is null and a fresh instance cannot be created. In that case it falls back to `writeMethod.invoke(target, sourceValue)`, which copies the whole nested object graph in one step and bypasses per-path authorization for everything under that property. That still does not look equivalent to `ParametersInterceptor` semantics.

One additional point, which I do not currently consider blocking, is scalar collection / map / array parity. The new helpers correctly add `"[0]"` for nested bean elements, but scalar elements are still copied directly. In core, indexed paths like `tags[0]` still contribute nesting depth even when the leaf value is scalar, so this may still be slightly more permissive than the form/query parameter path for collection-like inputs. I do not see that as a security blocker in the same class as unrestricted nested bean copying, but it may still be worth tightening later for consistency.

lukaszlenart:
Suggestion: Consider rejecting deserialization entirely when `requireAnnotations=true` and no no-arg constructor exists, rather than applying a best-effort scrub. Log a warning and skip body processing.

tranquac:
Thanks for the continued review. v4 patch (commit `72cc282`) addresses both points:

---

### Blocking fix — bulk-copy fallback removed from `copyAuthorizedProperties`

The previous code fell back to `writeMethod.invoke(target, sourceValue)` when a nested target bean was null and `createFreshInstance` failed:

```java
// BEFORE (v3 — bug):
} else {
    writeMethod.invoke(target, sourceValue); // entire object graph copied unfiltered
    continue;
}
```

This copied the entire nested object graph in one step, bypassing per-path authorization for every property underneath that node.

Fix: log a warning and skip, consistent with the policy already applied to collection/array elements:

```java
// AFTER (v4):
} else {
    LOG.warn("REST nested bean [{}] skipped — no no-arg constructor for [{}],"
            + " cannot authorize its nested properties",
            fullPath, sourceValue.getClass().getName());
    continue;
}
```

---

### Suggestion implemented — reject body when no no-arg constructor at top level

Previously the fallback was to deserialize directly into the target and then apply a best-effort scrub. As you noted, the scrub path cannot guarantee that all nested unauthorized properties are nulled out.

v4 rejects body deserialization entirely when `requireAnnotations=true` and no no-arg constructor is available:

```java
Object freshInstance = createFreshInstance(target.getClass());
if (freshInstance != null) {
    handler.toObject(invocation, reader, freshInstance);
    copyAuthorizedProperties(freshInstance, target, invocation.getAction(), target, "");
} else {
    LOG.warn("REST body rejected: requireAnnotations=true but [{}] has no no-arg constructor; "
            + "body deserialization skipped to preserve @StrutsParameter authorization integrity",
            target.getClass().getName());
    // handler.toObject() never called
}
```

`scrubUnauthorizedProperties` and its recursive helper have been removed as dead code.

---

### Test results

```
plugins/rest: 76 tests, 0 failures
```

lukaszlenart:
Thanks, the latest update addresses the remaining blocking concern on the REST path. Skipping body processing when no suitable top-level instance can be created, and skipping nested beans instead of bulk-copying them, preserves the `@StrutsParameter` authorization model much better.

I still see some minor parity / follow-up items around scalar collection handling, but I do not consider those blocking. From my side, this looks good to go. I will let others to chime in before merging the PR - thanks a lot!

tranquac:
If there are any further issues that need correction or addition, let me know so I can improve it. Thanks @lukaszlenart  and reviewers!

tranquac:
Hi @lukaszlenart 

Could you provide an update on the review progress from you and the team? I think we need to make a quick decision since we are making this potential security risk public to the world.

rgielen (approved):
LGTM 👍

lukaszlenart:
> Could you provide an update on the review progress from you and the team? I think we need to make a quick decision since we are making this potential security risk public to the world.

A bit too late to worry about that ... if you have any additional info about attack vector or some in-depth analysis please send them to security@struts.apache.org - right now I treat this PR a security hardening change.

tranquac:
yep @lukaszlenart 

Do we need to wait for all reviewers to approve? I noticed that a few of them haven't been active recently

tranquac:
Hi @jogep, @aleksandr-m, @yasserzamani, @sdutry, @cnenning, @kusalk

Could you pls help me check this PR?

lukaszlenart (approved):
LGTM 👍 

Closes #1651
Closes #1652

PR #1666, opened 2026-04-19
introduce private method to remove clones

aaaZayne:
Simplify error handling by introducing _ErrorIfAbsent_


core/src/main/java/org/apache/struts2/dispatcher/multipart/AbstractMultiPartRequest.java
```diff
@@ -278,15 +276,17 @@ public void parse(HttpServletRequest request, String saveDir) throws IOException
...
             LOG.warn("Unable to parse request", e);
             LocalizedMessage errorMessage = buildErrorMessage(e.getClass(), e.getMessage(), new Object[]{});
-            if (!errors.contains(errorMessage)) {
-                errors.add(errorMessage);
-            }
+            ErrorIfAbsent(errorMessage);
+        }
+    }
+
+    private void ErrorIfAbsent(LocalizedMessage errorMessage) {
```
lukaszlenart: Naming is wrong, it must use `camelCase` notation and it would be good to better express what it does, eg.: `addErrorIfAbsent` or `registerErrorIfAbsent`

lukaszlenart (changes requested):

lukaszlenart (approved):

PR #1689, opened 2026-05-17
fix(core): enforce class-level HTTP method annotations for wildcard-resolved unannotated methods

g0w6y:
## Summary

Fixes a regression introduced by the WW-5535 fix (commit 4d2eb93) where class-level
HTTP method annotations are silently ignored for wildcard-resolved action methods that
carry no method-level annotation.

## Problem

`HttpMethodInterceptor.intercept()` uses an `if/else-if` structure that creates a dead
zone. After the WW-5535 fix made wildcard-resolved methods report `isMethodSpecified()=true`,
the class-level annotation branch became structurally unreachable for unannotated methods:

```java
if (invocation.getProxy().isMethodSpecified()) {
    Method method = action.getClass().getMethod(invocation.getProxy().getMethod());
    if (AnnotationUtils.isAnnotatedBy(method, HTTP_METHOD_ANNOTATIONS)) {
        return doIntercept(invocation, method);
    }
    // unannotated method falls through silently
} else if (AnnotationUtils.isAnnotatedBy(action.getClass(), HTTP_METHOD_ANNOTATIONS)) {
    return doIntercept(invocation, action.getClass()); // never reached
}
return invocation.invoke(); // no enforcement
```

**Affected scenario:**

```java
@HttpPost // intends to restrict all requests to POST
public class OrderAction extends ActionSupport {
    public String create() { ... } // no method-level annotation
}
```

```xml
<action name="order-*" class="com.example.OrderAction" method="{1}">
```

`GET /order-create` — resolves `create()` via wildcard, `isMethodSpecified()=true`,
method has no annotation, else-if never evaluated, `@HttpPost` on the class is ignored,
request proceeds.

## Fix

Convert `else if` to a standalone `if` so the class-level annotation check is always
evaluated as a fallback when the method carries no annotation. Method-level annotations
still take precedence (checked first). One-line change.

```java
if (invocation.getProxy().isMethodSpecified()) {
    Method method = action.getClass().getMethod(invocation.getProxy().getMethod());
    if (AnnotationUtils.isAnnotatedBy(method, HTTP_METHOD_ANNOTATIONS)) {
        return doIntercept(invocation, method);
    }
}
if (AnnotationUtils.isAnnotatedBy(action.getClass(), HTTP_METHOD_ANNOTATIONS)) {
    return doIntercept(invocation, action.getClass());
}
return invocation.invoke();
```

## Tests Added

Two regression tests in `HttpMethodInterceptorTest`:

- `testWildcardResolvedUnannotatedMethodRespectsClassLevelAnnotation` — GET rejected on
  a class annotated with `@AllowedHttpMethod(POST)` when the resolved method is unannotated
- `testWildcardResolvedUnannotatedMethodAllowsPostWithClassLevelAnnotation` — POST allowed
  on the same configuration

## Related

- Fixes regression introduced by: #1592
- Original issue: [WW-5535](https://issues.apache.org/jira/browse/WW-5535)

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/httpmethod/HttpMethodInterceptor.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/httpmethod/HttpMethodInterceptor.java
@@ -90,7 +90,8 @@ public class HttpMethodInterceptor extends AbstractInterceptor {
                     invocation.getProxy().getMethod(), AllowedHttpMethod.class.getSimpleName(), request.getMethod());
                 return doIntercept(invocation, method);
             }
-        } else if (AnnotationUtils.isAnnotatedBy(action.getClass(), HTTP_METHOD_ANNOTATIONS)) {
+        }
+        if (AnnotationUtils.isAnnotatedBy(action.getClass(), HTTP_METHOD_ANNOTATIONS)) {
             LOG.debug("Action: {} annotated with: {}, checking if request: {} meets allowed methods!",
                 action, AllowedHttpMethod.class.getSimpleName(), request.getMethod());
             return doIntercept(invocation, action.getClass());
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Why did you close this PR?

g0w6y:
please check https://github.com/apache/struts/pull/1690 @lukaszlenart

PR #1701, opened 2026-05-20
WW-5623 fix(core): HTML-encode form action in PostbackResult to prevent XSS

lukaszlenart:
Backports #1653 to the 6.x line.

`PostbackResult.doExecute()` embeds `finalLocation` into a `<form action="">` attribute via raw string concatenation. Wrap it with `StringEscapeUtils.escapeHtml4()`, consistent with the encoding approach used in `DefaultActionProxy`, `Property`, and `TextProviderHelper`.

See [#1653](https://github.com/apache/struts/pull/1653) for the full analysis.

Adds three regression tests covering the XSS payload, all special characters, and the clean-URL regression case.

Fixes [WW-5623](https://issues.apache.org/jira/browse/WW-5623).

### DIFF
--- a/core/src/main/java/org/apache/struts2/result/PostbackResult.java
+++ b/core/src/main/java/org/apache/struts2/result/PostbackResult.java
@@ -21,6 +21,7 @@ package org.apache.struts2.result;
 import com.opensymphony.xwork2.ActionContext;
 import com.opensymphony.xwork2.ActionInvocation;
 import com.opensymphony.xwork2.inject.Inject;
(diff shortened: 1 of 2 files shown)


lukaszlenart:
## Just one important remark

This PR doesn't address a vulnerability, it's pure framework hardening. The potential attack vector assumes a developer uses a raw user input in OGNL expression evaluation which is against our guideline:
https://struts.apache.org/security/#do-not-use-incoming-untrusted-user-input-in-forced-expression-evaluation

PR #1721, opened 2026-05-29
WW-5630 - Performance Issue SecurityMemberAccess

brianandle:
Closes [WW-5630](https://issues.apache.org/jira/browse/WW-5630)


core/src/main/java/org/apache/struts2/util/ConfigParseUtil.java
```diff
@@ -33,6 +36,13 @@
 import static org.apache.commons.lang3.StringUtils.strip;
 
 public class ConfigParseUtil {
+    // Size the cache to prevent excessive memory usage in environments with many classloaders and/or large numbers of classes being validated.
+    // While still providing a reasonable caching benefit for common cases (e.g. multiple Struts instances in the same container, or multiple calls to validate the same class across different containers).
+    private static final int MAX_CLASS_CACHE_SIZE = 50;
+
+    private static final Cache<ClassLookupKey, Class<?>> VALIDATED_CLASS_CACHE = Caffeine.newBuilder()
```
lukaszlenart: **Classloader-pinning risk (WW-5537 territory).** This `static` cache holds strong `Class<?>` *values*, each of which keeps its defining `ClassLoader` (and every class that loader defined) alive until size-50 LRU eviction. In the production hot path the loader is always `OgnlUtil.class.getClassLoader()`, but `validateClasses(Set, ClassLoader)` is `public` — any caller passing a webapp loader creates a leak vector in whatever loader holds `ConfigParseUtil`.

Recommend `weakValues()` (Class no longer pins its loader) plus `weakKeys()` on a real `ClassLoader` key — see the comment on `classLoaderName`.

core/src/main/java/org/apache/struts2/util/ConfigParseUtil.java
```diff
@@ -73,14 +83,67 @@ public static Set<Class<?>> validateClasses(Set<String> classNames, ClassLoader
...
+                    throw new ClassLookupException(e);
+                }
+            });
+        } catch (ClassLookupException e) {
+            throw (ClassNotFoundException) e.getCause();
+        }
+    }
+
+    private static String classLoaderName(ClassLoader classLoader) {
+        return String.valueOf(classLoader);
```
lukaszlenart: **Most important: this string is not a reliable classloader identity.** `ClassLoader` doesn't override `toString()`, so this is normally `Class@hexHash` — but a custom loader overriding `toString()` to a constant (your own `CountingClassLoader` does exactly this), or a plain `hashCode()` collision, makes two *distinct* loaders produce the same key. The cache then returns a `Class<?>` loaded by a different loader.

Since those `Class` objects drive `SecurityMemberAccess` exclusion/allowlist checks (`==` / `isAssignableFrom`), a wrong-loader `Class` could subtly weaken OGNL access control. Key on the actual `ClassLoader` reference instead — e.g. a `Cache<ClassLoader, Cache<String, Class<?>>>` with `weakKeys()`, which uses identity equality and is GC-safe, fixing this and the pinning issue in one go. The `classLoaderName` helper name also implies a stable identity it doesn't provide.

core/src/main/java/org/apache/struts2/util/ConfigParseUtil.java
```diff
@@ -73,14 +83,67 @@ public static Set<Class<?>> validateClasses(Set<String> classNames, ClassLoader
...
+        try {
+            return VALIDATED_CLASS_CACHE.get(lookupKey, key -> {
+                try {
+                    return validatingClassLoader.loadClass(key.className);
+                } catch (ClassNotFoundException e) {
+                    throw new ClassLookupException(e);
+                }
+            });
+        } catch (ClassLookupException e) {
+            throw (ClassNotFoundException) e.getCause();
```
lukaszlenart, line 111: Minor: the unchecked cast `(ClassNotFoundException) e.getCause()` is safe today because `ClassLookupException`'s only constructor takes a `ClassNotFoundException`, but a one-line comment documenting that invariant would prevent a future NPE/CCE if another constructor is ever added.

core/src/test/java/org/apache/struts2/util/ConfigParseUtilTest.java
```diff
@@ -0,0 +1,161 @@
...
+        for (int i = 0; i < 10; i++) {
+            ConfigParseUtil.validateClasses(classNames, loader);
+        }
+
+        assertEquals(1, loader.getStringClassLoads());
+    }
+
+    public void testValidateClassesSeparatesEntriesAcrossDifferentClassLoaders() {
+        CountingClassLoader firstLoader = new CountingClassLoader(getClass().getClassLoader(), "loader-one");
+        CountingClassLoader secondLoader = new CountingClassLoader(getClass().getClassLoader(), "loader-two");
```
lukaszlenart, line 66: Good coverage overall. One gap that matters: every test uses `toString()` as the loader identity (`"loader-one"`, `"loader-two"`, `"loader-" + i`), which is exactly what masks the key-collision bug. Please add a test with **two distinct loaders that return the same `toString()`** and assert each loads independently — against the current `String.valueOf(classLoader)` keying that test would fail (cross-loader cache hit), demonstrating the issue and guarding the fix.

lukaszlenart (changes requested):
Thanks for tackling WW-5630 — the motivation is solid. `SecurityMemberAccess` is `Scope.PROTOTYPE` (`StrutsBeanSelectionProvider.java:454`), so it's reconstructed on every value-stack creation (~per request), and each construction re-runs the `@Inject use*` setters → `validateClasses` → `ClassLoader.loadClass(...)`. Memoizing that is worthwhile, and Caffeine is already a `core` dependency.

Two blocking issues before merge, both rooted in the cache key:

1. **Key on classloader identity, not `String.valueOf(classLoader)`.** This can return a `Class<?>` from the wrong loader (collisions / constant `toString()`), and those `Class` objects feed OGNL exclusion/allowlist checks that rely on `==` — security-adjacent.
2. **Static cache holding strong `Class<?>` values pins classloaders** — relevant given the WW-5537 leak work on adjacent branches.

Both are solved together with `Caffeine.newBuilder().maximumSize(...).weakKeys().weakValues().build()` keyed on the real `ClassLoader` (e.g. `Cache<ClassLoader, Cache<String, Class<?>>>`). `weakKeys()` gives identity equality (fixes #1) and GC-safety (fixes #2). Details inline.

brianandle:
@lukaszlenart I believe I've resolved all the concerns you raised and added additional UTs as well. 

-------------------------------------------------------------------------------
Test set: org.apache.struts2.util.ConfigParseUtilTest
-------------------------------------------------------------------------------
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 s -- in org.apache.struts2.util.ConfigParseUtilTest

core/src/main/java/org/apache/struts2/util/ConfigParseUtil.java
```diff
@@ -33,6 +35,15 @@
 import static org.apache.commons.lang3.StringUtils.strip;
 
 public class ConfigParseUtil {
+    // Size the cache to prevent excessive memory usage in environments with many classloaders and/or large numbers of classes being validated.
+    // While still providing a reasonable caching benefit for common cases (e.g. multiple Struts instances in the same container, or multiple calls to validate the same class across different containers).
+    private static final int MAX_CLASS_CACHE_SIZE = 50;
+
+    private static final Cache<ClassLoader, Cache<String, Class<?>>> VALIDATED_CLASS_CACHE = Caffeine.newBuilder()
+            .weakKeys()
+            .weakValues()
```
lukaszlenart: Non-blocking: `weakValues()` on the **outer** cache makes the per-loader inner `Cache` (the *value*) weakly reachable. Nothing else holds a strong reference to an inner cache, so GC can reclaim it while its `ClassLoader` key is still alive and in active use — which intermittently empties the cache and partially defeats the optimization under memory pressure.

`weakKeys()` already guarantees the entry is dropped as soon as the loader becomes unreachable, so the outer `weakValues()` isn't needed for leak-safety. Suggest removing it from the **outer** builder (keep `weakKeys()` here and the `weakValues()` on the inner cache at line 97, which correctly lets unreferenced `Class` values be reclaimed). Correctness is fine either way; this is purely about cache effectiveness.

Also minor: with the redesign `maximumSize(50)` now bounds the number of *classloaders*, not classes — the `MAX_CLASS_CACHE_SIZE` comment above still describes bounding classes.

lukaszlenart (approved):
Thanks for the quick turnaround — this addresses all four points cleanly:

- ✅ Keyed on the actual `ClassLoader` via a nested `Cache<ClassLoader, Cache<String, Class<?>>>` with `weakKeys()` → identity equality (no more `toString()` collisions) and GC-safe.
- ✅ `weakValues()` so cached `Class` objects no longer pin their loader.
- ✅ `ClassLookupKey`/`classLoaderName` removed; cast invariant now documented on `ClassLookupException`.
- ✅ Added the same-`toString()` collision test plus several others.

I ran `mvn test -DskipAssembly -pl core -Dtest=ConfigParseUtilTest` locally — **9/9 green**. Approving.

One non-blocking nuance left on the outer cache (inline) — worth a look but not a blocker.

lukaszlenart:
@brianandle thanks, all looks good 💪

PR #1774, opened 2026-07-09
WW-5642 fix(rest): authorize @StrutsParameter on record/creator-bound REST body properties

g0w6y:
Fixes [WW-5642](https://issues.apache.org/jira/browse/WW-5642)

## Summary

`ParameterAuthorizingModule` enforces `@StrutsParameter` on REST/JSON body deserialization by wrapping each Jackson property's `deserializeAndSet` / `deserializeSetAndReturn` (`AuthorizingSettableBeanProperty`). Jackson never calls either method for **creator-bound properties** — Java `record`s, `@JsonCreator` constructors, `@ConstructorProperties` — it instead calls `SettableBeanProperty#deserialize` directly, which Jackson declares `final`, so the existing wrapper cannot intercept it.

Practical effect: with `struts.parameters.requireAnnotations` enabled, any `record`-typed field anywhere in a REST action's request body (top-level or nested) is populated with **no `@StrutsParameter` check at all**, silently defeating the protection for that entire subtree. This is easy to hit unintentionally — `record` is the idiomatic way to model immutable REST DTOs on the Java versions Struts now targets.

Verified with a live reproduction before writing the fix: a `record` field with an unauthorized component was populated anyway; after the fix, the same component is correctly rejected while an authorized sibling component still passes through.

## Fix

Add `AuthorizingValueDeserializer`, which wraps the property's *value deserializer* rather than the `SettableBeanProperty` — the value deserializer is what the final `deserialize()` method delegates to, so wrapping it is the only available interception point for the creator-bound path. It is installed from `AuthorizingSettableBeanProperty#withValueDeserializer`, scoped to `CreatorProperty` specifically, so ordinary setter/field/builder-pattern properties — already correctly authorized via the existing `deserializeAndSet`/`deserializeSetAndReturn` overrides — are not checked a second time (which would also double-push the path-authorization stack and produce incorrect nested paths).

## Test plan

- [x] New regression test `testRecordComponentAuthorizedByPath` in `ParameterAuthorizingModuleTest`: a `record`-typed nested property with one authorized and one unauthorized component — confirms the unauthorized component is rejected and the authorized one still deserializes.
- [x] Full existing `ParameterAuthorizingModuleTest` suite (setter, builder, nested, collection, map, array authorization paths) passes unchanged — confirms no regression from scoping the new check to creator-bound properties only.
- [x] Full `struts2-rest-plugin` module test suite passes (106/106).

### DIFF
--- a/core/src/main/java/org/apache/struts2/interceptor/parameter/ParameterAuthorizationContext.java
+++ b/core/src/main/java/org/apache/struts2/interceptor/parameter/ParameterAuthorizationContext.java
@@ -39,6 +39,7 @@ public final class ParameterAuthorizationContext {
 
     private static final ThreadLocal<State> STATE = new ThreadLocal<>();
     private static final ThreadLocal<Deque<String>> PATH_STACK = ThreadLocal.withInitial(ArrayDeque::new);
+    private static final ThreadLocal<Deque<RedactionScope>> REDACTION_STACK = ThreadLocal.withInitial(ArrayDeque::new);
 
     private ParameterAuthorizationContext() {
         // utility
@@ -67,6 +68,7 @@ public final class ParameterAuthorizationContext {
     public static void unbind() {
         STATE.remove();
         PATH_STACK.remove();
+        REDACTION_STACK.remove();
     }
 
     /**
@@ -130,6 +132,48 @@ public final class ParameterAuthorizationContext {
         return prefix.isEmpty() ? propertyName : prefix + "." + propertyName;
     }
 
+    /**
+     * Pushes a fresh redaction scope, tracking whether any creator-bound property directly inside
+     * the object about to be constructed was rejected by {@link #isAuthorized}. Callers that
+     * construct a bean via a creator/builder (e.g. a bean-level deserializer wrapper) push a scope
+     * before construction and pop it in a {@code finally} block afterwards.
+     */
+    public static void pushRedactionScope() {
+        REDACTION_STACK.get().push(new RedactionScope());
+    }
+
+    /**
+     * Pops the current redaction scope. Has no effect if the stack is empty.
+     */
+    public static void popRedactionScope() {
+        Deque<RedactionScope> stack = REDACTION_STACK.get();
+        if (!stack.isEmpty()) {
+            stack.pop();
+        }
+    }
+
+    /**
+     * Marks the current redaction scope as having dropped at least one property. Used to
+     * distinguish "this construction failed because we substituted a redacted value" (safe to
+     * treat the whole object as unauthorized) from "this construction failed for an unrelated
+     * reason" (a genuine client error, which must still propagate).
+     */
+    public static void markRedacted() {
+        RedactionScope top = REDACTION_STACK.get().peek();
+        if (top != null) {
+            top.redacted = true;
+        }
+    }
+
+    /**
+     * @return {@code true} if {@link #markRedacted()} was called since the current redaction scope
+     * was pushed.
+     */
+    public static boolean wasRedactedInCurrentScope() {
+        RedactionScope top = REDACTION_STACK.get().peek();
+        return top != null && top.redacted;
+    }
+
     private static final class State {
         final ParameterAuthorizer authorizer;
         final Object target;
@@ -140,5 +184,9 @@ public final class ParameterAuthorizationContext {
             this.target = target;
             this.action = action;
         }
+    }
+
+    private static final class RedactionScope {
+        boolean redacted;
     }
 }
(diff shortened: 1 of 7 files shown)


lukaszlenart:
Thanks for the PR but please create a JIRA ticket to cover this work 🙏

g0w6y:
Created WW-5642 to track this: https://issues.apache.org/jira/browse/WW-5642. Ready for review/merge whenever you get a chance  thanks!

g0w6y:
Given this fixes an authorization bypass where unauthorized REST body fields could be set, should we handle this as a security advisory instead of a normal fix? Happy to follow whichever process you prefer, just let me know.

lukaszlenart:
Thanks — the creator-bound gap is real and worth closing. As it stands though the coverage isn't complete: it handles records/`@JsonCreator` via `withValueDeserializer`, but primitive/required creator components (rejection returns `null`), collection-path handling, and the broader creator styles (`@ConstructorProperties`, `@JsonCreator` constructors, top-level records, collection/map creator params) aren't covered or tested. I'd want those addressed before it lands.

Are you planning to extend it, or would you prefer I take it forward under a linked JIRA?

g0w6y:
Pushed a fix for the gaps above (redacted creator-bound construction failures now fail closed instead of throwing, with regression tests for the other creator styles) and since this is an authz bypass, should it go through a security advisory like #1775/#1776?

lukaszlenart:
I recommend reading [SECURITY](https://github.com/apache/struts/blob/main/SECURITY.md) guideline first to understand how responsible disclosure looks like

lukaszlenart:
Thanks for the thorough follow-up — `0c93ea2fa` closes the completeness gaps I raised. I re-ran both suites locally against the PR head:

- `ParameterAuthorizationContextTest` (core) — green
- `ParameterAuthorizingModuleTest` (rest) — 23/23 green

The `RedactionAwareDeserializer` + redaction-scope stack is a clean way to handle the construction-failure case: substituting a stand-in for a dropped creator-bound component and then treating a resulting construction failure as "drop the whole object" is consistent with how a rejected non-creator nested property already behaves. Good that `testValidatingRecord_genuineClientErrorStillPropagates` pins down the other side of it — a construction failure with nothing dropped still propagates, so genuine client/data errors aren't masked. Coverage across records, static-factory `@JsonCreator`, `@ConstructorProperties`, top-level records, 3-level nesting, and `List`/`Map` creator params is exactly the matrix I was after.

Three small, non-blocking notes:

1. **Array-typed creator param.** `prefixForNested` handles `type.isArray()`, but only `List`/`Map` element paths are exercised in tests — an `Item[]` creator-param case would round out the collection matrix.
2. **`FAIL_ON_NULL_FOR_PRIMITIVES` off (the default).** With it disabled, a dropped primitive creator component silently becomes the type default (`0`/`false`) rather than dropping the object. That's fine — the point is the client value never lands — but worth a one-line comment so it reads as a deliberate choice rather than an oversight.
3. **Co-located failures.** If the same object both had a property dropped *and* hit an unrelated `JsonMappingException`, the current scope is marked, so the unrelated error is folded into "object dropped." Harmless in outcome, just slightly less informative to the caller — fine to leave, worth being aware of.

None of these block. Nice work tightening it up.

g0w6y:
Addressed all three notes in 3b184bf.

Added the array creator param test (testArrayOfRecordsAsCreatorParam) with a WithArray fixture, so the collection matrix now covers List, Map and array. Also added the two clarifying comments you suggested: one in AuthorizingValueDeserializer noting that a redacted primitive component falls back to the type default when FAIL_ON_NULL_FOR_PRIMITIVES is off, and one in RedactionAwareDeserializer noting that a redaction co located with an unrelated mapping error is folded into object dropped as a deliberate fail closed choice.

Full ParameterAuthorizingModuleTest suite passes locally (24 tests). This change lives in the REST Jackson plugin and is independent of the JSON plugin per request refactor in #1782, so it can land on its own. Ready whenever you are.

lukaszlenart:
Thanks for this — the analysis is correct and the fix is well-targeted. To frame it for anyone reading: the per-property `@StrutsParameter` enforcement introduced in 7.2.0 covered Jackson's setter/field path but not the creator-bound path (records, `@JsonCreator`, `@ConstructorProperties`), which Jackson populates through the value deserializer. This PR **completes that enforcement** rather than fixing a regression — before 7.2.0 REST bodies had no per-property control at all, so nothing that was previously protected was exposed.

The approach (wrapping the creator-property value deserializer, plus fail-closing construction that fails after a redaction) is sound, and the record / static-factory / `@ConstructorProperties` / nested / collection test matrix is exactly what I'd want to see. I'll take it through review and merge once CI is green. Appreciate you also checking on the disclosure process up front — for this one, completing a brand-new control in the open is fine.

PR #1775, opened 2026-07-09
WW-5643 fix(json): confine StrutsJSONReader parse state to the parsing thread

g0w6y:
Fixes [WW-5643](https://issues.apache.org/jira/browse/WW-5643)

## Summary

`JSONInterceptor` obtains its `JSONReader` once via `@Inject` (through `JSONUtil`) and reuses that same instance across every concurrent request handled by that interceptor -- this is how every Struts interceptor's injected dependencies work, not something specific to this plugin. `StrutsJSONReader` (introduced to enforce the `maxElements`/`maxDepth`/`maxStringLength`/`maxKeyLength` DoS limits) kept its parse cursor, current character, token, string/number buffer and nesting-depth counter as **plain instance fields**. Two concurrent `read()` calls on the same reader instance therefore tear each other's parsing state.

Concretely, under ordinary concurrent traffic to the same JSON action (no special configuration required -- these limits are on by default):

- The shared `depth` counter can be decremented by an unrelated concurrent request finishing its own parse while another request's deeply-nested payload is still mid-parse, letting a payload nested deeper than the configured `maxDepth` through undetected. This defeats the exact DoS protection this limit exists to provide.
- The shared character cursor and token buffer let fragments of **one request's JSON body leak into a different, concurrently-parsed request's result** -- i.e. one user's request data can end up inside an unrelated concurrent user's deserialized object.

Both were verified with live, quantified reproductions before writing the fix (not just reasoned about): a payload nested one level deeper than `maxDepth` (which must always be rejected) was incorrectly accepted under two-thread contention on a shared reader instance, and a "victim" payload containing a unique marker string bled into a concurrently-parsing "attacker" payload's own parsed result on the same shared reader.

## Fix

Move the cursor, current character, token, buffer and depth into a `ParseState` object confined to a `ThreadLocal`, created fresh in `read(String)` and cleared in a `finally` block. All method signatures and control flow are otherwise unchanged, so existing `StrutsJSONReader` subclasses (the `next()`/`skipWhiteSpace()`/`object()`/`array()`/`number()`/`string()`/`add()`/`addDigits()`/`unicode()` extension points are all `protected`) continue to work unmodified -- the parse-state fields were already `private`, so no external code could have been touching them directly. The limit fields (`maxElements`, `maxDepth`, `maxStringLength`, `maxKeyLength`) stay as ordinary instance fields, since they're set to the same value on every call for a given interceptor configuration and are safe to share across threads.

## Test plan

- [x] Two new regression tests in `StrutsJSONReaderTest`:
  - `testConcurrentReuseDoesNotBypassMaxDepth`: two threads share one reader instance; one repeatedly submits a payload one level deeper than `maxDepth`, the other hammers the same instance with unrelated shallow payloads. Asserts the over-depth payload is *never* accepted.
  - `testConcurrentReuseDoesNotLeakDataAcrossParses`: two threads share one reader instance; a "victim" thread submits JSON containing a unique marker string while an "attacker" thread concurrently parses unrelated JSON and inspects its own result. Asserts the attacker's parsed result never contains a fragment of the victim's data.
- [x] Full existing `StrutsJSONReaderTest`/`JSONReaderTest`/`JSONInterceptorTest` suites pass unchanged -- confirms no behavioral regression from the refactor.
- [x] Full `struts2-json-plugin` module test suite passes (127/127).

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/StrutsJSONReader.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/StrutsJSONReader.java
@@ -30,6 +30,13 @@ import java.util.Map;
  * Deserializes an object from a JSON string with configurable limits
  * to prevent denial-of-service attacks via malicious payloads.
  * </p>
+ *
+ * <p>
+ * A single StrutsJSONReader instance is shared across all concurrent requests handled by a given
+ * JSONInterceptor (it is injected once, not created per request), so the cursor, token buffer and
+ * nesting depth of an in-progress parse are kept in a {@link ThreadLocal}, not instance fields --
+ * otherwise two concurrent {@link #read(String)} calls would corrupt each other's parse state.
+ * </p>
  */
 public class StrutsJSONReader implements JSONReader {
     private static final Object OBJECT_END = new Object();
@@ -47,16 +54,20 @@ public class StrutsJSONReader implements JSONReader {
             't', '\t'
     );
 
-    private CharacterIterator it;
-    private char c;
-    private Object token;
-    private final StringBuilder buf = new StringBuilder();
+    private static final class ParseState {
+        private CharacterIterator it;
+        private char c;
+        private Object token;
+        private final StringBuilder buf = new StringBuilder();
+        private int depth;
+    }
+
+    private static final ThreadLocal<ParseState> PARSE_STATE = new ThreadLocal<>();
 
     private int maxElements = DEFAULT_MAX_ELEMENTS;
     private int maxDepth = DEFAULT_MAX_DEPTH;
     private int maxStringLength = DEFAULT_MAX_STRING_LENGTH;
     private int maxKeyLength = DEFAULT_MAX_KEY_LENGTH;
-    private int depth;
 
     @Override
     public void setMaxElements(int maxElements) {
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Thank you for the PR but please create a JIRA ticket to cover this work 🙏 Also bear in mind that I work on changes in the JSON plugin as well #1773

g0w6y:
Created WW-5643 to track this: https://issues.apache.org/jira/browse/WW-5643. Ready for review/merge whenever you get a chance thanks!

g0w6y:
Given this fixes a cross request data leak between concurrent users along with a maxDepth limit bypass, should we handle this as a security advisory instead of a normal fix? Happy to follow whichever process you prefer, just let me know.

lukaszlenart:
Thanks for the patch, and especially the concurrency regression tests. The thread-safety problem is real — `StrutsJSONReader` keeps per-parse state on a shared instance. I'm merging this as the immediate fix.

Longer term I'd rather address it at the lifecycle level in `JSONUtil` (a fresh instance per operation) rather than per-field `ThreadLocal`s, to avoid the cleanup contract. Tracked in [WW-5650](https://issues.apache.org/jira/browse/WW-5650), which will preserve your regression tests. Appreciate the contribution.

g0w6y:
hey @lukaszlenart lets go with security advisory too whats your opinion

lukaszlenart:
I opened a followup PR #1782 to address the mentioned improvement

PR #1776, opened 2026-07-09
WW-5644 fix(json): confine StrutsJSONWriter write state to the writing thread

g0w6y:
Fixes [WW-5644](https://issues.apache.org/jira/browse/WW-5644)

## Summary

`JSONUtil` obtains its `JSONWriter` once via `@Inject` and reuses that same instance across every concurrent response handled by that `JSONResult`/`JSONInterceptor` configuration -- this is how every Struts interceptor's injected dependencies work, not something specific to this plugin. `StrutsJSONWriter` kept its output buffer, cyclic-reference-detection stack, root object, and expression-path state (`buf`, `stack`, `root`, `buildExpr`, `exprStack`, `excludeProperties`, `includeProperties`, `excludeNullProperties`) as **plain instance fields**, all reset in place at the start of `write()`:

```java
this.buf.setLength(0);
this.stack.clear();
this.root = object;
...
this.value(object, null);
return this.buf.toString();
```

## Root cause

Two concurrent `write()` calls on the same writer instance race on that reset. `buf.setLength(0)` at the top of one call can wipe out another, unrelated, concurrently-in-progress call's already-appended output; both calls then keep appending their own serialized fragments into the same, shared `StringBuilder`. Because each call finishes by reading back whatever the buffer currently contains (`this.buf.toString()`), **one request's `write()` call can return a different, concurrently-served request's fully serialized response body** -- not just a corrupted mix, but the other request's complete JSON, verbatim. This is the same class of defect as #1775 (`StrutsJSONReader` sharing per-parse state across concurrent request bodies), but on the response side, which is exercised by the default `json` result type on every request with no opt-in configuration required.

Verified with a live, quantified reproduction before writing the fix: two threads sharing one `StrutsJSONWriter` instance, one repeatedly serializing a "victim" object containing a unique secret marker, the other repeatedly serializing an unrelated "attacker" object and inspecting its own returned string. Out of 50,000 attempts, the attacker's own `write()` call returned the victim's complete secret value 180 times, plus a partial fragment of it 612 more times.

## Fix

Move `buf`/`stack`/`root`/`buildExpr`/`exprStack`/`excludeProperties`/`includeProperties`/`excludeNullProperties` into a `WriteState` object confined to a `ThreadLocal`, created fresh in `write(...)` and cleared in a `finally` block. All method signatures and control flow are otherwise unchanged, so existing `StrutsJSONWriter` subclasses (`bean()`/`map()`/`array()`/`string()`/`add()`/etc. are all `protected` extension points) continue to work unmodified -- the per-write-state fields were already `private`, so no external code could have been touching them directly. `ignoreHierarchy`/`dateFormat`/`enumAsBean`/`excludeProxyProperties` stay as ordinary instance fields, since they're set to the same value on every call for a given writer configuration and are safe to share across threads.

## Test plan

- [x] New regression test `testConcurrentReuseDoesNotSwapResponsesAcrossWrites` in `StrutsJSONWriterTest`: two threads share one writer instance, a "victim" thread repeatedly serializes an object containing a unique secret marker while an "attacker" thread concurrently serializes an unrelated object and inspects its own returned string. Asserts the attacker's response never contains the victim's secret.
- [x] Full existing `StrutsJSONWriterTest` suite passes unchanged -- confirms no behavioral regression from the refactor.
- [x] Full `struts2-json-plugin` module test suite passes (126/126).

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/StrutsJSONWriter.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/StrutsJSONWriter.java
@@ -66,6 +66,14 @@ import java.util.regex.Pattern;
  * Serializes an object into JavaScript Object Notation (JSON). If cyclic
  * references are detected they will be nulled out.
  * </p>
+ *
+ * <p>
+ * A single StrutsJSONWriter instance is shared across all concurrent responses handled by a given
+ * JSONResult/JSONInterceptor (it is injected once, not created per request), so the output buffer,
+ * cycle-detection stack and other per-write state are kept in a {@link ThreadLocal}, not instance
+ * fields -- otherwise two concurrent {@link #write(Object)} calls would corrupt each other's output,
+ * including returning one request's serialized data as another, unrelated request's response.
+ * </p>
  */
 public class StrutsJSONWriter implements JSONWriter {
 
@@ -84,17 +92,22 @@ public class StrutsJSONWriter implements JSONWriter {
         BEAN_INFO_CACHE.clear();
     }
 
-    private final StringBuilder buf = new StringBuilder();
-    private final Deque<Object> stack = new ArrayDeque<>();
+    private static final class WriteState {
+        private final StringBuilder buf = new StringBuilder();
+        private final Deque<Object> stack = new ArrayDeque<>();
+        private Object root;
+        private boolean buildExpr = true;
+        private String exprStack = "";
+        private Collection<Pattern> excludeProperties;
+        private Collection<Pattern> includeProperties;
+        private boolean excludeNullProperties;
+    }
+
+    private static final ThreadLocal<WriteState> WRITE_STATE = new ThreadLocal<>();
+
     private boolean ignoreHierarchy = true;
-    private Object root;
-    private boolean buildExpr = true;
-    private String exprStack = "";
-    private Collection<Pattern> excludeProperties;
-    private Collection<Pattern> includeProperties;
     private DateFormat dateFormat;
     private boolean enumAsBean = ENUM_AS_BEAN_DEFAULT;
-    private boolean excludeNullProperties;
     private boolean excludeProxyProperties;
     private ProxyService proxyService;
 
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Thank you for the PR but please create a JIRA ticket to cover this work 🙏 Also bear in mind that I work on changes in the JSON plugin as well #1773

g0w6y:
Created WW-5644 to track this: https://issues.apache.org/jira/browse/WW-5644. Ready for review/merge whenever you get a chance thanks!

g0w6y:
Given this fixes a cross request response leak between concurrent users, should we handle this as a security advisory instead of a normal fix? Happy to follow whichever process you prefer, just let me know.

lukaszlenart:
Thanks for the patch, and especially the concurrency regression tests. The thread-safety problem is real — `StrutsJSONWriter` keeps per-write state on a shared instance. I'm merging this as the immediate fix.

Longer term I'd rather address it at the lifecycle level in `JSONUtil` (a fresh instance per operation) rather than per-field `ThreadLocal`s, to avoid the cleanup contract. Tracked in [WW-5650](https://issues.apache.org/jira/browse/WW-5650), which will preserve your regression tests. Appreciate the contribution.

g0w6y:
hey @lukaszlenart lets go with security advisory too whats your opinion

lukaszlenart:
I opened a followup PR #1782 to address the mentioned improvement

g0w6y:
Quick follow-up for my records: are WW-5644 (#1776) and WW-5643 (#1775) getting a CVE / security advisory, or are they being treated as normal hardening fixes? Either is fine, just want to know how they'll be classified.

lukaszlenart:
Please follow the responsible disclosure process and ask such questions via [security@struts.apache.org](mailto:security@struts.apache.org) as documented in the [SECURITY.md](https://github.com/apache/struts/blob/main/SECURITY.md#do-not-disclose-through-a-pull-request-commit-or-issue)

PR #1780, opened 2026-07-11
WW-5646 Modernize path normalization in Include component

arunmanni-ai:
The path normalization logic in Include.getContextRelativePath() used the 
legacy java.util.Stack class and did not handle the edge case where 
dot-dot segments exceed the available path depth.

This patch:
- Replaces java.util.Stack with ArrayDeque (recommended since JDK 6)
- Adds an isEmpty() guard before pop() to handle excess dot-dot gracefully
- Uses enhanced for-loop instead of indexed elementAt() iteration
- Adds test coverage for edge case dot-dot paths

### DIFF
--- a/core/src/main/java/org/apache/struts2/components/Include.java
+++ b/core/src/main/java/org/apache/struts2/components/Include.java
@@ -50,6 +50,7 @@ import java.util.Map;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.StringTokenizer;
+import java.util.Iterator;
 
 /**
  * <!-- START SNIPPET: javadoc -->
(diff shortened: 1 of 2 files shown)


arunmanni-ai:
Fixed a regression introduced by the Stack → ArrayDeque swap: the rebuild loop iterated the deque head-to-tail (most recently pushed first), which is the reverse of Stack's insertion-order iteration. This caused rebuilt paths like `car/view.jsp` to come out as `view.jsp/car`, which is what the CI failures were catching.

Fixed by using `descendingIterator()` to restore the original oldest-first ordering when rebuilding the flat path string. The traversal-prevention logic (tokenizing, `..` → pop, the `isEmpty()` guard clamping excess `..` at root) is unchanged.

Verified locally — all passed.

lukaszlenart (approved):

lukaszlenart:
Great, thanks a lot for this patch!

PR #1784, opened 2026-07-14
WW-4858 Evaluate JSON name allowlist at leaf keys only

lukaszlenart:
Fixes [WW-4858](https://issues.apache.org/jira/browse/WW-4858)

## What & why

Follow-up to #1773. That change added `ParametersInterceptor`-style name/value
filtering to `JSONInterceptor`'s reflection population path, but applied **every**
name check at **every** node of the JSON tree before recursing. Two of those
checks — **accepted name patterns** (`AcceptedPatternsChecker`) and the
**`ParameterNameAware`** callback — target the *full dotted binding path*, so
evaluating them on an intermediate node was wrong:

- A leaf-targeting accepted pattern such as `bean\.stringField` fails to match the
  intermediate node `bean`, so the whole `bean` subtree was dropped before the
  leaf was ever visited. The flat `ParametersInterceptor` path only evaluates the
  complete name `bean.stringField`, so it accepts it. Divergence.
- Same shape for `ParameterNameAware`: `acceptableParameterName("bean")` was
  invoked on the intermediate node, which the flat path never sees.
- For arrays, the accepted allowlist judged the **container** name (`list`)
  instead of the element path (`list[0]`), so a container-matching pattern let
  elements through that the flat path would reject at `list[0]`.

These only ever *tightened* acceptance (no security bypass), but they made
restrictive accepted/`ParameterNameAware` config silently drop nested JSON.

## Change

Split the per-key gate in the recursive walk:

- **Per-node (every node):** key length, excluded name patterns, `@StrutsParameter`
  authorization, and (opt-in) property filters. Exclusion is prefix-safe and
  authorization is intentionally hierarchical, so these are correct at intermediate
  nodes — unchanged.
- **Leaf keys only:** accepted name patterns and `ParameterNameAware`, now also
  applied to scalar array elements at their indexed path (`items[0]`).

This reproduces the flat-path semantics exactly. `excludeProperties`/`includeProperties`
behavior is unchanged (include patterns already hierarchy-expand across levels).

## Tests (`JSONInterceptorTest`)

- `testAcceptedNamePatternRejectsNestedKey` — leaf-targeting accepted pattern now
  populates the nested leaf and drops the non-matching sibling.
- `testAcceptedNamePatternAppliesToListElementPath` — accepted patterns evaluated
  at the array element path, not the container.
- `testParameterNameAwareDoesNotRejectIntermediateNode` — an action rejecting the
  intermediate node name no longer drops the subtree.
- `testIncludePropertiesAppliedToNestedInputWhenEnabled` — regression guard;
  nested include-property filtering still works.

All three fix-driving tests were confirmed to fail before the change.

## Follow-up: scalar array element parity

A later commit closes a gap raised in review: scalar list elements previously ran only
the leaf name-allowlist and value checks, skipping the per-node checks (length, excluded
patterns, `@StrutsParameter` authorization, property filters). They now also run
`isAcceptableNode` at the element path, so a scalar element is gated exactly as
`ParametersInterceptor` gates the flat name `items[0]`. Added tests:
`testExcludedNamePatternAppliesToListElementPath` and `testAuthorizationAppliesToListElementPath`.

## Behavior note (7.3.0)

Under `requireAnnotations`, JSON population of scalar collection elements (e.g.
`{"list":["x"]}` → `list[0]`) now enforces `@StrutsParameter` on the property **getter** at
`depth ≥ 1`, matching the documented depth model and the form-parameter
(`ParametersInterceptor`) path. Previously the JSON path checked only the container property
(setter, depth 0), which under-enforced authorization. Actions relying on the old behavior
must annotate the getter — e.g. `@StrutsParameter(depth = 1) List<String> getList()`.
List-of-objects (`items[0].name`, depth 2) is unaffected.

## Testing

`mvn test -pl plugins/json -am -DskipAssembly` — 148 tests, all pass.

## Note

#1783 (which added tests documenting the accepted-name limitation) is now merged. This PR builds on it: it upgrades that PR's `testAcceptedNamePatternRejectsNestedKey` to the plain `bean\.stringField` pattern (which the fix now makes work), reuses its `testIncludePropertiesAppliedToNestedInputWhenEnabled` as a regression guard, and adds the production fix plus net-new tests. Rebased onto current `main`.

### DIFF
--- a/plugins/json/src/main/java/org/apache/struts2/json/JSONInterceptor.java
+++ b/plugins/json/src/main/java/org/apache/struts2/json/JSONInterceptor.java
@@ -235,12 +235,19 @@ public class JSONInterceptor extends AbstractInterceptor {
             }
             String fullPath = prefix.isEmpty() ? key : prefix + "." + key;
 
-            if (!isAcceptableKey(fullPath, target, action)) {
+            Object value = entry.getValue();
+            boolean leaf = !(value instanceof Map) && !(value instanceof java.util.List);
+
+            // Per-node checks (length, excluded patterns, @StrutsParameter authorization, property
+            // filters) apply to every node. Name-allowlist checks (accepted patterns,
+            // ParameterNameAware) apply only at leaf keys, so an intermediate node is never gated by
+            // a pattern written for a leaf path — mirroring ParametersInterceptor, which only ever
+            // evaluates complete leaf names.
+            if (!isAcceptableNode(fullPath, target, action) || (leaf && !isAcceptableLeafName(fullPath, action))) {
                 it.remove();
                 continue;
             }
 
-            Object value = entry.getValue();
             if (value instanceof Map) {
                 filterUnacceptableKeysRecursive((Map) value, fullPath, target, action);
             } else if (value instanceof java.util.List) {
@@ -263,15 +270,23 @@ public class JSONInterceptor extends AbstractInterceptor {
                 filterUnacceptableKeysRecursive((Map) item, elementPrefix, target, action);
             } else if (item instanceof java.util.List) {
                 filterUnacceptableList((java.util.List) item, elementPrefix, target, action);
-            // Scalar list elements are value-checked only; their parent key already passed name/authorization checks.
-            } else if (!isAcceptableValue(elementPrefix, item, action)) {
+            // Scalar list elements are leaf binding targets: apply the leaf name-allowlist checks at
+            // the element path (e.g. "items[0]") plus value checks. The container key already passed
+            // the per-node checks in the caller.
+            } else if (!isAcceptableLeafName(elementPrefix, action) || !isAcceptableValue(elementPrefix, item, action)) {
                 it.remove();
             }
         }
     }
 
-    @SuppressWarnings("rawtypes")
-    private boolean isAcceptableKey(String fullPath, Object target, Object action) {
+    /**
+     * Checks applied to every node of the JSON tree (both intermediate objects/arrays and leaves):
+     * key length, excluded name patterns, {@code @StrutsParameter} authorization, and — when
+     * enabled — the interceptor's own property filters. Excluded patterns are prefix-safe and
+     * authorization is intentionally hierarchical (a parent must be authorized for a child to be
+     * reachable), so these are correct to evaluate at intermediate nodes.
+     */
+    private boolean isAcceptableNode(String fullPath, Object target, Object action) {
         if (fullPath.length() > paramNameMaxLength) {
             LOG.warn("JSON body parameter [{}] is too long, allowed length is [{}]; rejected", fullPath, paramNameMaxLength);
             return false;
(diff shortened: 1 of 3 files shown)


lukaszlenart (commented):
## Code Review

### Overview
Follow-up to #1773. Splits the per-key gate so the two *full-path-targeting* checks — accepted name patterns and `ParameterNameAware` — run **only at leaf keys** (scalar values, incl. scalar array elements), while length/exclusion/`@StrutsParameter`/property-filter checks stay per-node. This reproduces the flat `ParametersInterceptor` semantics, which only ever judge complete leaf names. Rename `isAcceptableKey` → `isAcceptableNode` + new `isAcceptableLeafName`.

### Correctness — ✅
Traced against flat-path semantics:
- Nested object `{"bean":{"stringField":...}}`: `bean` is a Map → `leaf=false` → only `isAcceptableNode("bean")`; recursion reaches leaf `bean.stringField`, gated by both `isAcceptableNode` (length/excluded/auth/filters) and `isAcceptableLeafName` (accepted/`ParameterNameAware`). Combined == flat check on `bean.stringField`.
- Scalar array `{"list":["x","y"]}`: container `list` passes `isAcceptableNode` (no accepted check), then each element gated by `isAcceptableLeafName("list[0]")` (the new addition) + pre-existing value check. Correct leaf path.
- Boolean `!isAcceptableNode(...) || (leaf && !isAcceptableLeafName(...))` and the `leaf` predicate (`!Map && !List`) are correct.

**Security:** change only *relaxes* the previously over-tightened intermediate-node gating; no boundary weakened. Excluded patterns and `@StrutsParameter` authorization stay per-node (correct — exclusion is prefix-safe, authorization is intentionally hierarchical), and every leaf still gets its full-path allowlist check. A leaf the flat path would reject is still rejected. No bypass.

### Code quality & conventions — ✅
- Clear, accurate Javadoc on both extracted methods explaining *why* each group belongs at its level — guards against regressing to the #1773 behavior.
- Dropping `@SuppressWarnings("rawtypes")` from `isAcceptableNode` is correct (body no longer touches raw types).
- No dangling callers of the renamed method.

Minor nits (non-blocking):
- `java.util.List` fully qualified inline rather than imported (consistent with surrounding code).
- Map/nested-List *elements* inside a list still skip `isAcceptableNode` at the container-element path — **pre-existing**, out of scope, and the leaf underneath is still fully checked.

### Test coverage — ✅
- `testAcceptedNamePatternRejectsNestedKey` upgraded to the plain `bean\.stringField` pattern.
- `testAcceptedNamePatternAppliesToListElementPath` — allowlist judged at element path, not container (`getList() == null || isEmpty()` assertion is correct).
- `testParameterNameAwareDoesNotRejectIntermediateNode` — exercises the intermediate-node case.
- The `ParameterAwareTestAction.acceptableParameterName` change (also rejecting `"bean"`) does **not** affect the other two consumers, which only use `foo`/`bar`/`baz`. Verified.

### Verdict
**LGTM.** Correct, well-documented, properly test-driven. The two nits are optional tidy-ups for a later pass; neither blocks merge.

PR #1809, opened 2026-07-23
WW-5462 Move org.apache.struts2.mock package out of the published struts2-core jar

lukaszlenart:
Fixes [WW-5462](https://issues.apache.org/jira/browse/WW-5462)

Removes the test-only `org.apache.struts2.mock` package from the published `struts2-core` jar.

## Changes

- **`core/pom.xml`**: attaches a filtered `test-jar` containing only `org/apache/struts2/mock/**` (with `<archive combine.self="override"/>` so it gets a default manifest instead of the main jar's OSGi manifest)
- **Six plugin poms** (json, rest, spring, xslt, jasperreports, jasperreports7): consume the test-jar as a `<type>test-jar</type>` test-scope dependency
- **`MockContainer`**: deleted — zero references anywhere in the codebase
- **5 mock classes** (`MockActionInvocation`, `MockActionProxy`, `MockInterceptor`, `MockObjectTypeDeterminer`, `MockResult`): moved as pure renames from `core/src/main/java` to `core/src/test/java` — package unchanged, so no import edits anywhere
- **spring plugin**: dropped its byte-identical local copy of `DummyTextProvider`

## Notes

- **Breaking change (7.3.0)**: downstream projects using these mocks in their own tests must switch to `struts2-junit-plugin`, Mockito, or copy the needed class. Release notes need a migration note. The attached `-tests.jar` is internal build plumbing, not supported API (deliberately not in the bom).
- Dev-build caveat: in unpackaged reactor runs (`mvn test`), Maven resolves the test-jar dep to core's `target/test-classes` directory (filter applies only at packaging); verified green in both modes. CI's `mvn verify` exercises the packaged/filtered path.
- Design/plan docs included under `docs/superpowers/`.

## Verification

- Full reactor `mvn test -DskipAssembly`: core 2997 + plugins (spring 61, junit 12, jasperreports 9, jasperreports7 12, json 163, rest 118, xslt 17) — all green
- `mvn clean install -Pjakartaee11 -DskipAssembly`: green
- `struts2-core-*-tests.jar` contains only the 8 mock classes, default manifest; main jar no longer contains the package


lukaszlenart:
Superseded: WW-5462 deferred to Struts 8.0.0, to be realised as a supported struts2-mocks module built on the extracted struts2-api (WW-4759) rather than this interim test-jar approach.

PR #1890, opened 2026-09-01
WW-5713 Fail closed for legacy Tiles OGNL evaluation

kuranikaran:
## Summary

The Tiles plugin registers separate `S2:` and legacy `OGNL:` attribute-expression evaluators. The legacy evaluator operates directly against the Tiles `Request` and does not use the Struts OGNL controls used by `S2:`.

This change keeps `OGNL:` registered but makes it fail closed by default. Evaluation throws Tiles `EvaluationException` with migration guidance and does not parse or evaluate the expression.

Applications that temporarily require the existing raw behavior can explicitly set:

```xml
<constant name="struts.tiles.ognl.legacy.enabled" value="true"/>
```

Normal Tiles startup no longer uses `DispatcherListener`. The no-argument factory registers a lazy configured `OGNL:` evaluator. On its first evaluation, the evaluator obtains the `Dispatcher` from the current request's `ServletContext`, reads `struts.tiles.ognl.legacy.enabled`, and caches the resulting delegate for that evaluator instance.

A false or missing flag, missing `Dispatcher`, or non-servlet `Request` fails closed with `EvaluationException`. Only an explicit `true` constructs the raw evaluator and installs the global `OgnlRuntime` Tiles `Request` property accessor. Legacy mode preserves the existing behavior and emits one migration warning when first used. Public boolean constructors support custom initializers that need to select disabled or legacy behavior explicitly.

The compatibility constant and legacy evaluator are deprecated in Struts 7.4.0 and targeted for removal in Struts 8.0.0.

`S2:`, `I18N:`, and `EL:` remain unchanged.

## Testing

- compiled the Tiles plugin
- ran the focused evaluator, factory, listener-lifecycle, integration, and legacy compatibility tests in both class orders
- ran the complete Tiles module test suite (561 tests, all passing)
- ran the complete Tiles module `verify` lifecycle (561 tests, all passing), including Apache RAT and JAR packaging
- tested per-webapp isolation, one-time cached configuration resolution, non-servlet and missing-Dispatcher handling, construction gating, concurrent first evaluation, explicit public constructors, and real servlet-backed `S2:` evaluation
- inspected compiled bytecode to confirm the normal listener contains no `DispatcherListener` integration and raw evaluator construction remains unreachable until explicit `true` is established
- ran the default and Jakarta EE 11 reactors; all modules through Tiles passed, after which Showcase encountered two external `gitbox.apache.org` download timeouts

## Classification

This is defense-in-depth hardening. No attacker-controlled expression source or concrete vulnerability impact has been demonstrated, and no security advisory is proposed.

Issue: [WW-5713](https://issues.apache.org/jira/browse/WW-5713)

### DIFF
--- a/plugins/tiles/README.md
+++ b/plugins/tiles/README.md
@@ -4,3 +4,17 @@ You will find more details in [documentation](https://struts.apache.org/plugins/
 
 ## Installation
 Just drop this plugin JAR into `WEB-INF/lib` folder or add it as a Maven dependency.
+
+## Legacy Tiles OGNL expressions
+
+The legacy Tiles `OGNL:` attribute-expression evaluator is deprecated in Struts 7.4.0 and disabled by default. Use
+`S2:` for expressions that should be evaluated against the Struts ValueStack, or use an ordinary Tiles mechanism.
+
+Applications that temporarily require the legacy raw evaluator can set the following Struts constant:
+
+```xml
+<constant name="struts.tiles.ognl.legacy.enabled" value="true"/>
+```
+
+Enabling the constant produces a startup warning. The compatibility constant is deprecated in Struts 7.4.0; both it
+and the legacy evaluator are targeted for removal in Struts 8.0.0.
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/DisabledOgnlAttributeEvaluator.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/DisabledOgnlAttributeEvaluator.java
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *  http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
@@ -16,35 +16,23 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.tiles.ognl;
+package org.apache.struts2.tiles;
 
-import ognl.Ognl;
-import ognl.OgnlException;
 import org.apache.tiles.core.evaluator.AbstractAttributeEvaluator;
 import org.apache.tiles.core.evaluator.EvaluationException;
 import org.apache.tiles.request.Request;
 
 /**
- * Evaluates attribute expressions and expressions with OGNL language.
- *
- * @since 2.2.0
- * @deprecated This legacy evaluator does not use the Struts OGNL controls used by {@code S2:} and is disabled by
- * default. Temporary use requires {@code struts.tiles.ognl.legacy.enabled=true}. Migrate to {@code S2:} or ordinary
- * Tiles mechanisms. This evaluator is targeted for removal in Struts 8.0.0.
+ * Fails closed when the deprecated Tiles OGNL evaluator has not been explicitly enabled.
  */
-@Deprecated(since = "7.4.0", forRemoval = true)
-public class OGNLAttributeEvaluator extends AbstractAttributeEvaluator {
+final class DisabledOgnlAttributeEvaluator extends AbstractAttributeEvaluator {
 
-    /** {@inheritDoc} */
+    static final String DISABLED_MESSAGE = "The Tiles OGNL evaluator is disabled. Migrate the expression to S2:, "
+        + "or temporarily enable struts.tiles.ognl.legacy.enabled. Legacy Tiles OGNL support will be removed in "
+        + "Struts 8.0.0.";
+
     @Override
     public Object evaluate(String expression, Request request) {
-        if (expression == null) {
-            throw new IllegalArgumentException("The expression parameter cannot be null");
-        }
-        try {
-            return Ognl.getValue(expression, request);
-        } catch (OgnlException e) {
-            throw new EvaluationException("Cannot evaluate OGNL expression '" + expression + "'", e);
-        }
+        throw new EvaluationException(DISABLED_MESSAGE, null);
     }
 }
--- a/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesContainerFactory.java
+++ b/plugins/tiles/src/main/java/org/apache/struts2/tiles/StrutsTilesContainerFactory.java
@@ -39,6 +39,7 @@ import org.apache.tiles.core.definition.pattern.PatternDefinitionResolver;
 import org.apache.tiles.core.definition.pattern.PrefixedPatternDefinitionResolver;
 import org.apache.tiles.core.definition.pattern.regexp.RegexpDefinitionPatternMatcherFactory;
 import org.apache.tiles.core.definition.pattern.wildcard.WildcardDefinitionPatternMatcherFactory;
+import org.apache.tiles.core.evaluator.AttributeEvaluator;
 import org.apache.tiles.core.evaluator.AttributeEvaluatorFactory;
 import org.apache.tiles.core.evaluator.BasicAttributeEvaluatorFactory;
 import org.apache.tiles.core.evaluator.impl.DirectAttributeEvaluator;
@@ -73,6 +74,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Dedicated Struts factory to build Tiles container with support for:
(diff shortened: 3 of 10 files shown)


lukaszlenart:
Thanks for filing the ticket and the PR, and sorry for the slow turnaround.

The evaluator part is right, including the two things that are easy to get wrong and that ruled out
both of the options we discussed earlier:

- it gates *construction* (`createConfiguredOgnlEvaluator`) rather than `evaluate()`, so on the
  default path we no longer perform the global
  `OgnlRuntime.setPropertyAccessor(Request.class, ...)` mutation at all — that was the main thing I
  wanted and it's easy to miss;
- disabled mode throws `EvaluationException` instead of returning an empty string, and the language
  stays registered rather than being unregistered. Both alternatives fail silently — an empty string
  is indistinguishable from an empty attribute, and an unregistered language falls through to
  `DirectAttributeEvaluator` and renders `OGNL:...` as literal page text.

So the hardening itself I'm happy with. What I can't take as it stands is the bootstrap change.

**The `StrutsTilesListener` / `DispatcherListener` bootstrap**

`Dispatcher.dispatcherListeners` is `private static final` (`Dispatcher.java:126`) — one list per
classloader, not per web application, and `Dispatcher` clears it wholesale at `:515`. Hooking the
Tiles lifecycle to it has three consequences in a shared-classloader container:

- `StrutsTilesListener.java:73` — `dispatcherDestroyed` fires for *any* Dispatcher, so undeploying
  webapp A destroys webapp B's live Tiles container, and because the list is cleared B can never
  rebuild it.
- `StrutsTilesListener.java:92` — the `struts.tiles.ognl.legacy.enabled` value is read from
  whichever Dispatcher initialises first, which may belong to a different application. A webapp that
  never opted in can have the raw `OGNL:` evaluator switched back on by a neighbour. That inverts the
  control this PR adds, which is why it's the blocking one for me.
- `StrutsTilesListener.java:46` — if no Struts Dispatcher ever initialises, Tiles is now never
  initialised at all, where previously it was built at context startup. That's a behaviour change for
  setups that don't route through our Dispatcher.

Also `StrutsTilesInitializer.java:41`: both flag-carrying constructors are package-private and only
the listener reads the constant, so an application using its own initializer or factory — which the
factory's own javadoc suggests — can't reach the documented escape hatch and gets no diagnostic.

None of this is you going off-piste. Getting the constant to `createAttributeEvaluatorFactory` is
genuinely awkward: it's a Tiles `AbstractTilesContainerFactory` override with no container access,
and container creation happens at application init when there may be no `ActionContext`. I flagged
that as the only non-trivial part of the change when we discussed the design, and this is where it
bites.

What I'd like instead is to decouple the two: keep the container bootstrap exactly where it is today
and resolve the flag without adopting the Dispatcher lifecycle — reading the constant lazily at
factory time or on first evaluation, or, if a listener really is needed, keying it to its own
`ServletContext` identity so it ignores lifecycle events from other applications. Happy to talk
through the options if you'd rather settle the approach before rewriting.

Two loose ends:

- The escape hatch needs a visible end date or it becomes permanent, so I've filed
  [WW-5714](https://issues.apache.org/jira/browse/WW-5714) for the removal in 8.0.0 alongside your
  WW-5713 deprecation in 7.4.0. Nothing for you to do there — it just means the flag can't quietly
  become permanent.
- Do you want `support/struts-6-x-x` as well? The same evaluator is there. It would be plain
  `@Deprecated` without `forRemoval`, since that branch is still on Java 8.

Credit is yours in the release notes either way.

kuranikaran:
Thanks for the detailed review. Your shared-classloader concerns make sense, especially the possibility of one application’s Dispatcher lifecycle or legacy setting affecting another application. I agree that the current DispatcherListener wiring should be removed.

Before rewriting it, I propose this approach:

- Restore StrutsTilesListener and the normal Tiles bootstrap lifecycle to their existing main-branch behavior, with no DispatcherListener registration.
- Keep OGNL: registered with a lazy gating evaluator.
- At evaluation time, derive the ServletContext from the current Tiles Request/ApplicationContext and resolve only that context’s Dispatcher.
- If its struts.tiles.ognl.legacy.enabled value is absent or false, throw EvaluationException before constructing the raw evaluator or changing the global Request property accessor.
- If it is explicitly true, construct/cache the legacy evaluator and emit the migration warning once.
- Continue checking the flag against the current request’s ServletContext so another webapp cannot opt this one into legacy behavior.
- If no Dispatcher exists, fail closed by default.
- Make the explicit boolean constructors public for applications using their own StrutsTilesInitializer or StrutsTilesContainerFactory, including direct Tiles setups without a Dispatcher.

This should preserve the original bootstrap behavior, remove the static cross-application lifecycle dependency, and ensure raw OGNL construction remains reachable only through an explicit application-local opt-in.

Would this direction address your concerns before I revise the patch?

I’m also happy to prepare the support/struts-6-x-x backport as a separate PR after we settle and merge the main-line design, using plain @Deprecated as noted.

lukaszlenart:
Sorry for the delay — yes, that direction addresses all of it. Please go ahead.

Resolving the Dispatcher from the current request's `ServletContext` is the right key: `Dispatcher.getInstance(ServletContext)` (`Dispatcher.java:280`) reads the `SERVLET_DISPATCHER` attribute, which is per-webapp and cleared on cleanup, so no static state is involved and a neighbouring application can't opt this one in. Constructing the raw evaluator lazily and only after that check keeps the `OgnlRuntime.setPropertyAccessor(Request.class, ...)` mutation off the default path, which was the main thing. Public constructors for the explicit opt-in cover the custom-initializer case.

Two small things while you're in there:

- Resolve the flag once per evaluator instance and hold the result, rather than looking it up from the container on every evaluation. The constant can't change at runtime, and there is one evaluator per Tiles container, so caching the decision is safe and avoids a container lookup per rendered attribute.
- `ServletUtil.getServletRequest(request)` throws `NotAServletEnvironmentException` for a non-servlet `Request`. Treat that the same as "no Dispatcher": fail closed with the migration message, not a stack trace from the lookup.

A separate PR for `support/struts-6-x-x` afterwards is fine — plain `@Deprecated` there, as discussed.

kuranikaran:
Implemented the approved redesign in `917ce740`.

The normal Tiles bootstrap no longer uses `DispatcherListener`. On the first `OGNL:` evaluation, the evaluator resolves the Dispatcher from that request’s `ServletContext`, reads `struts.tiles.ognl.legacy.enabled` once, and caches the resulting delegate for its lifetime.

A missing Dispatcher, false or missing flag, or non-servlet request now fails closed with the migration `EvaluationException`. The raw evaluator and global `Request` accessor are constructed only after an application-local explicit opt-in.

The public boolean constructors remain available for custom initializers. Tiles validation passes with 561/561 tests, including per-webapp isolation, cached resolution, non-servlet handling, construction gating, concurrency, and unchanged `S2:` behavior.

The new GitHub Actions runs are awaiting repository approval.

lukaszlenart (approved):
Thanks Karan — `917ce740` does what we agreed, both small things from the go-ahead included. Checked against the earlier points:

- Dispatcher is resolved from the request's `ServletContext` on first use and the decision is held per evaluator instance; no static state is left anywhere, and `StrutsTilesListenerTest` pins that the listener stays out of the Dispatcher lifecycle.
- `createOGNLEvaluator()` and the `OgnlRuntime.setPropertyAccessor(Request.class, …)` mutation are only reached after the flag reads `true`; the integration test asserting zero raw-evaluator constructions at container build is exactly the guard that matters here.
- Disabled path throws `EvaluationException` with a fixed message — not an empty string, not literal text — and the expression is neither evaluated nor echoed.
- Missing Dispatcher and non-servlet `Request` both fail closed.

Verified locally: `plugins/tiles` 561/0, clean merge onto `main`; CI is green on the head.

Two non-blocking notes, your call whether to fold them in before merge:

**nitpick (non-blocking):** the `@deprecated` Javadoc on `TilesConstants.STRUTS_TILES_OGNL_LEGACY_ENABLED` and `OGNLAttributeEvaluator` names the removal version. In this repo the Javadoc stays at "deprecated since X, use Y"; `forRemoval = true` plus WW-5714 carry the rest. The runtime messages and the README are fine as they are — that is the migration prompt users actually see.

**thought (non-blocking):** `isLegacyOgnlEnabled` walks the request wrapper chain to reach the `ServletContext`. `createAttributeEvaluatorFactory` already has the container's own `ApplicationContext`, so `ServletUtil.getServletContext(applicationContext)` would tie the decision to the webapp that owns the evaluator however the `Request` was built, and the `NotAServletEnvironmentException` branch goes away. Same behaviour in every normal deployment; just a little less to reason about.

On the docs side, the Tiles plugin page on struts.apache.org needs the new constant — I can take that. Ignore the earlier remarks about a `support/struts-6-x-x` port: WW-5713 targets 7.4.0 only, nothing is needed on the 6.x line.

